PR #12, opened 2017-03-21
Use layer id (not object) in relations; YAML2QGS

gacarrillor:
+ In generated YAML files, avoid using layer objects and use instead layer inner ids in 'relations' section. After all, layer objects are already defined in 'legend' section.
+ Support YAML2QGS operation from the command line. You need to comment/uncomment lines in the bottom of metaproject.py to toggle modes.

Using layer ids should ease both YAML merging and layer grouping.

### DIFF
--- a/dataobjects/layers.py
+++ b/dataobjects/layers.py
@@ -22,21 +22,24 @@ from qgis.core import QgsVectorLayer, QgsDataSourceUri
 
 
 class Layer:
-    def __init__(self, provider, uri):
+    def __init__(self, provider, uri, id=None):
         self.provider = provider
         self.uri = uri
+        self.inner_id = self.inner_id() if id is None else id
         self.__layer = None
 
     def dump(self):
         definition = dict()
         definition['provider'] = self.provider
         definition['uri'] = self.uri
+        definition['innerId'] = self.inner_id
 
         return definition
 
     def load(self, definition):
         self.provider = definition['provider']
         self.uri = definition['uri']
+        self.inner_id = definition['innerId']
 
     def create(self):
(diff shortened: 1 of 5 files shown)


dataobjects/layers.py
```diff
@@ -46,4 +49,13 @@ def create(self):
         return self.__layer
 
     def table_name(self):
-        return QgsDataSourceUri(self.uri).table()
\ No newline at end of file
+        return QgsDataSourceUri(self.uri).table()
+
+    def inner_id(self):
```
m-kuhn: Can you add some small documentation that explains what this is used for (i.e. only within code here and yaml -> or also qgis project file).
I know I haven't been very good at documenting code but if more people start working on this, we really will need to improve on this kind of quality (myself included).

metaproject.py
```diff
@@ -66,6 +66,39 @@ def debug_log_message(message, tag, level):
 
     QgsApplication.exitQgis()
 
+def yaml2qgs(argv):
+    parser = argparse.ArgumentParser('Generate QGIS project from a YAML file.')
```
m-kuhn: Would be nice to have one single entry point with one parser that calls other methods based on arguments.

m-kuhn:
Thanks @gacarrillor 

Both are very good improvements. I left some comments and there might be more (didn't have the time to fully review).

Can you maybe split this into two different pull requests, so they can be reviewed and merged individually?

No worries for the travis errors, tests have not been working. Just fixed it.

gacarrillor:
Closing to open separated PRs as the author suggested.

PR #25, opened 2017-04-04
Add dialog to select ili2pg options

### LINKED ISSUE #23: Option for EPSG-Code
--defaultSrsCode
Without option, EPSG 21781 is set

gacarrillor:
Initially, options presented in the new dialog correspond to what we need to change from the way Project Generator executes ili2pg, which remains as default. 

Since options might not be self-explanatory, I've added tooltips to them. Selected options are stored in QSettings.

Fix #23


m-kuhn:
Thanks for the proposal. 

Can you shortly elaborate on the reason for the individual options to be available for the user?

This has been discussed (in general) at a meeting yesterday, and the conclusion was that we should only

 * add options that can be passed to ili2pg when they are properly supported by the project generator
 * when there is a real need that these options need to be specified differently for real projects

The reason for this is

 * one can always import the model manually (outside the plugin) and run the plugin on the generated db
 * what should be avoided is people messing around with the options and then opening issues "combination of A, B and C not supported for data model D, please fix!" or plainly forgetting that they have been messing with the options "because they are there".

m-kuhn:
Or to put it simple:

let's first discuss if the things you propose to add as new options shouldn't just be changed by default.

gacarrillor:
It makes sense. Perhaps Stefan (@edigonzales) can give me a hand here. For reference, this is the proposed dialog for Project Generator:

Project Generator ili2pg options

- **defaultSrsCode**: This one is easy :), we don't use the default EPSG code.
- **createEnumColAsItfCode**: Our model extends domains from LADM. Therefore, this option does not allow ili2pg to finish the job. See [docs](https://github.com/claeis/ili2db/blob/master/docs/ili2db.rst).

- **coalesceMultiSurface**: I think this one wouldn't affect our use case, but we simply don't need to enable it.
- **nameByTopic**: This shouldn't affect anyone if it's present in the dialog. We prefer to get table names with no prefix.

- We use **smart2Inheritance** instead of **smart1Inheritance**.
- **importTid**:  I don't have a strong position about this one. 

Keep in mind that our model is still evolving. The options we use have been chosen by @aranguren and validated by Stefan recently.

edigonzales:
Before going too much into details: Is the projectgenerator plugin used for importing INTERLIS transfer files **and** generating the qgs file?

I think different options of ili2pg would be needed for a proper import but are not needed for the project generator, e.g:

- EPSG-Code would be needed to make an proper import, but would not be needed for a pure project generator since it can be read from the database.
- smartXYInheritance / nameByTopic: has no impact on the project generator. 

Did I get it right?

m-kuhn:
What the generator does is

 1. read the postgres metadata catalog
 2. read the ili2pg metadata catalog (to be done)
 3. add additional information from configuration files (yaml, to be done)
 4. read other sources of information (noting planned, but possible)
 5. generate a qgis project file

To make it easier for the user, importing ili (and in the near future xtf files) is supported. The options set here are not directly used by the generator and passed on to the ili2pg (which in turn writes the catalog information which will then be used by the project generator). What this means is: doing the ili2pg task(s) outside the project generator and running the generator on the created database has the exact same result as running it inside. Therefore the need for allowing options inside the generator is not very high, professionals can run ili2pg manually.

The goals for the integrated ili2pg import are:

 * make it easy for the user to import interlis models/data (one-click solution... ok; maybe 2 clicks ;) )
 * make it hard for a user to shoot himself in the foot (by choosing good options for him)
 * therefore:
  * hide options that are always safe to use
  * hide options that don't play well with qgis
  * prefer stability and predictability over flexibility and personal preference

* **CRS** is important (with or without import), no questions asked. That should even be on the main dialog since it's not (necessarily?) present in interlis files and therefore often *needs* to be specified manually. 
* **nameByTopic**: let's change the default (until someone has a strong reason to change it back). Having to support different naming schemes *might* make the code harder to maintain, so let's only allow for options if this is strictly required.
* **createEnumColAsItfCode** this was filed as an issue, right? Is the upstream fix in ili2pg on your todo list? I think we this is just a workaround for a different issue.
* **coalesceMultiSurface** sounds like it's safe to set as default. Let's ditch the option and just set it (until there's an issue with it).
* **Inheritance** "proper" handling of inheritance is still on the todo list. It's very likely that the generator will focus on supporting one or two inheritance strategies well and doesn't care about the others. Until we are there I don't mind this option.
* **importTid** no idea

gacarrillor:
> "To make it easier for the user, importing ili (and in the near future xtf files) is supported. The options set here are not directly used by the generator and passed on to the ili2pg (which in turn writes the catalog information which will then be used by the project generator). **What this means is: doing the ili2pg task(s) outside the project generator and running the generator on the created database has the exact same result as running it inside.** Therefore the need for allowing options inside the generator is not very high, professionals can run ili2pg manually."

This is exactly what this PR aims for :)

Notes:

 - **nameByTopic**: I don't understand why naming schemes could affect the plugin. 

 - **createEnumColAsItfCode**: Please let me know which issue you are referring to.

m-kuhn:
> This is exactly what this PR aims for :)

As far as I can see, this PR aims to make more options available inside the plugin, which is quite the opposite ;)

* **nameByTopic**: It just potentially might because of the mapping between ili names and db names. Let's make the point on this: is it a problem for you if we take your approach as standard?

* **createEnumColAsItfCode**: https://github.com/claeis/ili2db/issues/19

gacarrillor:
> As far as I can see, this PR aims to make more options available inside the plugin, which is quite the opposite ;)

The PR aims to make it possible to import an INTERLIS model into PostgreSQL using the plugin, just as we do from the command line. 

My intention is not to expose in the dialog a large number of options. It's fine that we can reduce them to the minimum. 

 - **nameByTopic**: 

> Let's make the point on this: is it a problem for you if we take your approach as standard?

It's fine for us, let's hope it's also fine for people there in Switzerland.

  - **createEnumColAsItfCode**: 

> claeis/ili2db#19`

Even if this issue is solved, we'll still extend LADM domains and _createEnumColAsItfCode_ will still be incompatible with it.

m-kuhn:
* Expose CRS prominently (on the main interface in the ili part, sensible heuristics defaults still to be evaluated)
* Change default value for **nameByTopic**, remove the option
* Change default value for **importTid**, remove the option
* Remove the option for **coalesceMultiSurface**
* Leave the options for **createEnumColAsItfCode** (I am under the impression that the [discussion here](https://github.com/claeis/ili2db/issues/19#issuecomment-290111061) actually is related to the kind of extensions you mention. Let's re-evaluate this if it turns out to be true.)
* Leave the **inheritance strategy** for now.

Sorry to be a bit complicated here, I just made the experience that adding options and cluttering the interface always happens very easily and evaluating and removing options at a later stage is always harder.

m-kuhn:
Update from todays discussion: `--createEnumColAsItfCode` is a leftover from interlis 1 and it should be safe for the projectgenerator to never specify the parameter. One less options :+1:

gacarrillor:
Good, so the only option that would remain into the dialog is smart1Inheritance / smart2Inheritance. Do you think we can add it also on the main interface (horizontal radio buttons) or should we preserve the dialog?

gacarrillor:
I just updated the PR to reflect what we have agreed.

projectgenerator/gui/generate_project.py
```diff
@@ -181,3 +192,18 @@ def link_activated(self, link):
 
         else:
             QDesktopServices.openUrl(link)
+
+    def select_crs(self):
+        """ Open a dialog to select CRS """
+        crsSelector = QgsProjectionSelectionDialog(self)
+        crsSelector.setCrs(QgsCoordinateReferenceSystem(self.epsg, type=QgsCoordinateReferenceSystem.PostgisCrsId))
+        res = crsSelector.exec()
```
m-kuhn: Hmm, I thought that's called `exec_()`for some name collision or reserved keyword, but you probably tested it...

projectgenerator/libili2pg/iliimporter.py
```diff
@@ -92,15 +94,18 @@ def run(self):
...
-        args += ["--createEnumColAsItfCode"]
         args += ["--createNumChecks"]
-        args += ["--smart1Inheritance"]
         args += ["--coalesceMultiSurface"]
         args += ["--createGeomIdx"]
         args += ["--createFk"]
+        if self.configuration.inheritance == 'smart1':
+            args += ["--smart1Inheritance"]
+        else:
+            args += ["--smart2Inheritance"]
```
m-kuhn: That changes the default to smart2... No opinion about it yet from my side, just asking if it was done consciously

gacarrillor: It shouldn't change the default. When reading from `QSettings` while initializing the dialog, the default value is `'smart1'`.

projectgenerator/gui/generate_project.py
gacarrillor: Unlike with PyQt4, now both `exec()` and `exec_()` do the job. See [this answer](http://stackoverflow.com/questions/22610720/should-i-use-app-exec-or-app-exec-in-my-pyqt-application/22614643#22614643).

m-kuhn: Nice!

m-kuhn:
Thanks @gacarrillor 
I took the liberty and adjusted it a bit. Is this ok for you?

screenshot from 2017-04-12 09-56-36

m-kuhn:
Manually merged

PR #26, opened 2017-04-04
Small optimization loading layers and relations

gacarrillor:
Currently, our model has more than 100 tables(!) and the plugin takes a while loading them to QGIS.

From [the docs](http://qgis.org/api/classQgsProject.html#a9e9cf6a97f67d17b26faf95a6124cdf4), it's better to load a set of layers at once instead of one by one. I'm not sure if the same is true for relations, though. I couldn't find an `addRelations` method, but there is a `setRelations` one, which I'm using in this PR when appropriate.  

Please let me know if `setRelations` doesn't represent any improvement in terms of performance.


projectgenerator/libqgsprojectgen/dataobjects/project.py
```diff
@@ -69,23 +69,33 @@ def load(self, definition):
...
             qgis_project.setCrs(QgsCoordinateReferenceSystem.fromEpsgId(self.crs))
 
+        qgis_relations = list()
         for relation in self.relations:
             rel = relation.create()
             assert rel.isValid()
-            qgis_project.relationManager().addRelation(rel)
+            qgis_relations.append(rel)
+
+        if qgis_project.relationManager().relations():
```
m-kuhn: I think it doesn't make a performance difference, but it's still nice code.

Did you try the following code (which I think is a bit simpler):

```
qgis_relations = qgis_project.relationManager().relations()
for relation in self.relations:
    ...

qgis_project.relationManager().setRelations(qgis_relations)
```

m-kuhn:
Much faster indeed, manually merged

projectgenerator/libqgsprojectgen/dataobjects/project.py
gacarrillor: Certainly better, thanks.

PR #27, opened 2017-04-04
Filter tables by schema

### LINKED ISSUE #20: Filtering by schema
When loading directly from Postgis from a schema I noticed that also the table pulic.chenyx06_triangles was loaded (needed by ST_Fineltra())

This table is usually not wanted in the project and should be suppressed.

gacarrillor:
When a schema is given only load tables from that schema. 

When in PostGIS mode, one could leave the schema field blank. In that case, load tables from all schemas, except from those listed in IGNORED_SCHEMAS.

Fix #20


m-kuhn:
Good stuff 👍

m-kuhn:
Manually merged, thanks

PR #36, opened 2017-04-19
Optimize layer tree grouping (avoid moving layers)

gacarrillor:
Moving layers takes a while, perhaps due to layer cloning. 
This PR improves our use case (+200 tables) in almost 40 seconds (from 2m23s to 1m42s).

### DIFF
--- a/projectgenerator/libqgsprojectgen/dataobjects/project.py
+++ b/projectgenerator/libqgsprojectgen/dataobjects/project.py
@@ -78,7 +78,7 @@ class Project(QObject):
 
             qgis_layers.append(qgis_layer)
 
-        qgis_project.addMapLayers(qgis_layers)
+        qgis_project.addMapLayers(qgis_layers, False)
 
         if isinstance(self.crs, QgsCoordinateReferenceSystem):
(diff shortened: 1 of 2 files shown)


projectgenerator/libqgsprojectgen/dataobjects/project.py
```diff
@@ -95,8 +95,7 @@ def create(self, path: str, qgis_project: QgsProject):
 
         qgis_project.setCrs(self.crs)
 
-        if self.legend:
-            self.legend.create(qgis_project)
+        self.legend.create(qgis_project)
```
m-kuhn: Removing this might break some cases where a legend is not manually specified or generated.

Remember that the idea is that this should also be usable without ili2pg or pg involved

gacarrillor: I thought we could make sure that `self.legend` has at least a default value, say a `LegendGroup` corresponding to layer tree root. But since it is better to take the `None` case into account, I've reverted that `if` removal and added another clause to make sure all layers are added directly to layer tree if `self.legend` is `None`.

PR #37, opened 2017-05-02
Filter relations by schema

gacarrillor:
Follow-up of https://github.com/opengisch/projectgenerator/pull/27

### DIFF
--- a/projectgenerator/gui/generate_project.py
+++ b/projectgenerator/gui/generate_project.py
@@ -85,6 +85,8 @@ class GenerateProjectDialog(QDialog, DIALOG_UI):
                 self.txtStdout.setText(self.tr('Java could not be found. Please <a href="https://java.com/en/download/">install Java</a> and or <a href="#configure">configure a custom java path</a>. We also support the JAVA_HOME environment variable in case you prefer this.'))
                 return
 
+            configuration.schema = configuration.schema or configuration.database
+
         generator = Generator(configuration.uri, configuration.schema)
         available_layers = generator.layers()
         relations = generator.relations(available_layers)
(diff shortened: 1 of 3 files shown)


m-kuhn:
Will this also work if the schema is unspecified?

gacarrillor:
I think it's time to discuss something with you. When in 'Interlis' mode, I think the current handling of 'schema' parameter is not very intuitive; i.e., choosing the schema name to be the database name if no schema is provided. To improve that I think we could:

- always ask for a schema name (when in 'Interlis' mode), so users are always aware of which schema was used for the import; and,
- validate whether the given schema exists in the database, and if so, ask for another name (unless it's 'public') to avoid conflicts with existing database objects.

By the way, do you think warning/error messages should be printed in the main dialog's output area? I think so, just want to know if you agree.

m-kuhn:
I don't have the behavior in mind right now, but the way you describe it sounds like there is really room for improvement.

Some thoughts (which will only apply in interlis mode according to your specs):

 * by default fill the name from the model into the schema field (as placeholderText or text)
 * if there is a collision with an existing object make it yellow, add some nice tooltip and add a combobox to the right
   * Please choose (default)
   * Clean before restore! (in red)
   * Merge with existing schema
 * only enable the ok button if there is no collision or an option is chosen

Which error names / warnings are you referring to? I don't like the output window very much, it was just a quick and dirty way to proxy ili2pg output to the user. Better ideas welcome.

-----------

Meanwhile: I don't think this PR depends on this, it will probably need to be adjusted for multischema execution anyway. Could you adjust it and we move this discussion over to a new issue?

gacarrillor:
Right, I thought it would be easier to deal with, but things got a bit complicated now, so I agree that schema handling needs to be addressed in an issue.

We had a workshop last week showing the plugin off and we found several errors that need to be handled. For instance, DB connection issues. I will send those errors to the output area, if you don't mind, although, now that you mentioned it, why not having a log tab like QGIS Processing?

Finally, with the second commit, I make sure all relations in the database are fetched if the schema parameter is not given, which is in line with the approach we took for loading tables.

PR #42, opened 2017-05-22
Handle some potential errors

gacarrillor:
-  On Windows, I've come across JAVA_HOME paths that come with quotes and need to be cleaned before we append 'java' to build the executable path.

 -  If we load a model with no geometry tables, CRS would not be specified.

 - Use psycopg2 to catch errors with PostgreSQL database connections. Empty host/database/user are handled via `QRegExpValidator` and communicated to the user, first through an orange color in the corresponding line edit, and, after clicking on `Create`, through a message and control focus.

 - Subclass `QValidator` (`FileValidator`) to validate input Interlis files.
 
 - Make sure we set an encoding for `decode()` functions in iliimporter.py Otherwise, we get `UnicodeDecodeError`'s.

### DIFF
--- a/projectgenerator/gui/generate_project.py
+++ b/projectgenerator/gui/generate_project.py
@@ -25,8 +25,8 @@ from psycopg2 import OperationalError
 from projectgenerator.gui.config import ConfigDialog
 from projectgenerator.gui.ili2pg_options import Ili2pgOptionsDialog
 from projectgenerator.libili2pg.iliimporter import JavaNotFoundError
-from projectgenerator.utils.qt_utils import make_file_selector, Validators
-from qgis.PyQt.QtGui import QColor, QDesktopServices, QFont, QRegExpValidator
+from projectgenerator.utils.qt_utils import make_file_selector, Validators, FileValidator
+from qgis.PyQt.QtGui import QColor, QDesktopServices, QFont, QRegExpValidator, QValidator
 from qgis.PyQt.QtWidgets import QDialog, QDialogButtonBox
 from qgis.PyQt.QtCore import QCoreApplication, QSettings, QRegExp
 from qgis.core import QgsProject, QgsCoordinateReferenceSystem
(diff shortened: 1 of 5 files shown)


projectgenerator/libili2pg/iliimporter.py
```diff
@@ -121,7 +121,9 @@ def run(self):
             # By default try JAVA_HOME and PATH
             java_paths = []
             if 'JAVA_HOME' in os.environ:
-                java_paths += [os.path.join(os.environ['JAVA_HOME'], 'java')]
+                paths = os.environ['JAVA_HOME'].split(";")
```
m-kuhn: Should that be [`os.pathsep`](https://docs.python.org/3/library/os.html#os.pathsep)?

projectgenerator/gui/generate_project.py
```diff
@@ -64,6 +64,13 @@ def __init__(self, parent=None):
     def accepted(self):
         configuration = self.updated_configuration()
 
+        if not configuration.host:
+            self.txtStdout.setText(self.tr('Please set a host before creating the project.'))
+            return
+        if not configuration.database:
+            self.txtStdout.setText(self.tr('Please set a database before creating the project.'))
+            return
```
m-kuhn: Would be great to handle that with some warnings/colors on the dialog itself so the user doesn't need to click a button first to get the message but instead will have hints on the dialog already for recommended/required fields.

projectgenerator/libili2pg/iliimporter.py
gacarrillor: Genau.

projectgenerator/gui/generate_project.py
gacarrillor: I agree, I'll have a look at it.

projectgenerator/qgs_project_generator.py
```diff
@@ -28,6 +30,8 @@ def __init__(self, iface):
         self.iface = iface
         self.__generate_action = None
         self.__configure_action = None
+        if locale.getlocale() == (None, None):
+            locale.setlocale(locale.LC_ALL, '')
```
m-kuhn: What are these lines for?

gacarrillor: Those lines set the locale to the user's preferred locale (see https://docs.python.org/3.0/library/locale.html#locale.LC_ALL). It turns out that on our Windows machines `getlocale()` gives us `(None, None)`. Only after we call `locale.setlocale(locale.LC_ALL, '')`, we are able to get a proper encoding to use as `decode()` parameter.

projectgenerator/gui/generate_project.py
gacarrillor: Done in https://github.com/opengisch/projectgenerator/pull/42/commits/ac4ca7932f8fb11267b3f97cda83a86fe4616c7f

PR #58, opened 2017-06-27
Gui improvements

seralra96:
- Lock GUI controls when importing from PostGIS.

- Show Close button when finishing import from PostGIS.

- Add waiting cursor when plugin is running.


projectgenerator/gui/generate_project.py
```diff
@@ -91,11 +91,11 @@ def __init__(self, base_config, parent=None):
     def accepted(self):
         configuration = self.updated_configuration()
 
-        if not self.ili_file_line_edit.validator().validate(configuration.ilifile, 0)[0] == QValidator.Acceptable \
-                and not self.ili_models_line_edit.text():
-            self.txtStdout.setText(self.tr('Please set a valid INTERLIS file or model before creating the project.'))
-            self.ili_file_line_edit.setFocus()
-            return
+        if self.type_combo_box.currentData() == 'ili':
+            if not self.ili_file_line_edit.validator().validate(configuration.ilifile, 0)[0] == QValidator.Acceptable:
```
m-kuhn: We can also run without an ili file by providing a model name from a (remote or local) repository

projectgenerator/gui/generate_project.py
```diff
@@ -117,6 +117,16 @@ def accepted(self):
 
         self.save_configuration(configuration)
 
+        cursor = QApplication.overrideCursor()
+        if cursor is None or cursor == 0:
+            QApplication.setOverrideCursor( Qt.WaitCursor )
+        elif cursor.shape() != Qt.WaitCursor:
+            QApplication.setOverrideCursor( Qt.WaitCursor )
```
m-kuhn: I think it's best to set the cursor without any check beforehand. What would be great but would need more refactoring is the following

```
try:
    QApplication.setOverrideCursor( ... )
    some code ...
finally:
    QApplication.restoreOverrideCursor()
```

That would be exception safe (I hate to see QGIS stuck with a wait cursor because some exception was raised that the dev didn't anticipate). But one needs to make sure that there is no return within the try-block.

m-kuhn:
Good stuff!

m-kuhn:
Thanks a lot

PR #60, opened 2017-07-05
Fix XML parsing issues

gacarrillor:
This PR fixes a couple of issues we've come across when attempting to access remote model repositories.

Please see commit messages for details and let me know if something is not clear.

### DIFF
--- a/projectgenerator/libili2pg/ilicache.py
+++ b/projectgenerator/libili2pg/ilicache.py
@@ -95,6 +95,8 @@ class IliCache(QObject):
             root = ET.parse(file).getroot()
         except ET.ParseError as e:
             QgsMessageLog.logMessage(self.tr('Could not parse ilisite file `{file}` ({exception})'.format(file=file, exception=str(e))), self.tr('Projectgenerator'))
+            return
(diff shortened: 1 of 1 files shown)


projectgenerator/libili2pg/ilicache.py
```diff
@@ -95,6 +95,8 @@ def _process_ilisite(self, file):
             root = ET.parse(file).getroot()
         except ET.ParseError as e:
             QgsMessageLog.logMessage(self.tr('Could not parse ilisite file `{file}` ({exception})'.format(file=file, exception=str(e))), self.tr('Projectgenerator'))
+            return
```
m-kuhn: Just a minor sidenote: I try to avoid early exits with `return` as much as possible. Basically to make it easier to follow the codeflow. In a complex method it's easier for me to read

    if not error:
        do_something()

than

    if error:
        return
    # ... many lines of code ...
    do_someting()     # me finally arriving here with reading and wondering why this code never gets executed

But the case here seems simple enough, let's hope it doesn't evolve too much ;)

PR #61, opened 2017-07-10
Export v2

gacarrillor:
This PR brings Export back to life. 

We are done adding validation and GUI improvements!

### DIFF
--- a/projectgenerator/gui/export.py
+++ b/projectgenerator/gui/export.py
@@ -22,8 +22,8 @@ import os
 
 from projectgenerator.libili2pg.iliexporter import JavaNotFoundError
 from projectgenerator.libili2pg.ilicache import IliCache
-from projectgenerator.utils.qt_utils import make_save_file_selector, Validators, make_folder_selector
-from qgis.PyQt.QtGui import QColor, QDesktopServices, QFont, QRegExpValidator
+from projectgenerator.utils.qt_utils import make_save_file_selector, Validators, FileValidator, make_folder_selector
+from qgis.PyQt.QtGui import QColor, QDesktopServices, QFont, QRegExpValidator, QValidator
 from qgis.PyQt.QtWidgets import QDialog, QDialogButtonBox, QApplication, QCompleter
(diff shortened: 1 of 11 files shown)


projectgenerator/gui/export.py
```diff
@@ -19,12 +19,16 @@
 """
 
 import os
+from PyQt5.QtCore import Qt
+
+from PyQt5.QtWidgets import QCompleter
```
m-kuhn: Can you rewrite these to the same style as below? They both do the same, just for consistency.

    from qgis.PyQt.QtWidgets import xxx

projectgenerator/gui/export.py
```diff
@@ -46,9 +50,32 @@ def __init__(self, base_config, parent=None):
         self.base_configuration = base_config
         self.restore_configuration()
 
+        self.validators = Validators()
+        regexp = QRegExp("[a-zA-Z_0-9]+") # Non empty string
```
m-kuhn: I think for hostnames there are other characters allowed (or even expected) like `.` (ip addresses), `-` etc. Is this properly handled here? Maybe it's better to really just check for anything else than empty string?

gacarrillor: Fertig!

m-kuhn:
Nice, this looks good.

One last question, is it required to provide it with a .ili file? I thought that an export could also get the model information from a model in a repository (or even in the database?)

gacarrillor:
Right, we don't need an ili file for exporting data. As far as I know, ili2pg reads model definition from the database itself. However, we need to pass the 'scope' we'd like to limit data export to. ili2pg supports these scopes for exporting data, models, topics, baskets, or datasets. One of them must be given (see https://github.com/claeis/ili2db/blob/master/src/ch/ehi/ili2db/base/Ili2db.java#L1276). We are currently just using models, but we could add support for other scopes as soon as we need them.

gacarrillor:
Right, a model would be enough. Could you please point me to what leads you to think about an ili file?

m-kuhn:
Can't find it any more, all good :)
Nice work

PR #63, opened 2017-07-10
Random fixes

gacarrillor:
1. Adjust `placeholderText` for schema according to current mode (Interlis/PostGIS).
2. Parse local ili models with `utf-8` encoding. If that fails (`UnicodeDecodeError`), try with 'latin-1'. If that still fails, log it and continue silently. 
3. Since ili2pg creates the schema in lower case, deal with lower-case schemas in Project Generator. Otherwise, the plugin won't be able to load tables from the database when using schemas with capital letters.
4. Send ili file parameter to the end of the command.

### DIFF
--- a/projectgenerator/libili2pg/ilicache.py
+++ b/projectgenerator/libili2pg/ilicache.py
@@ -130,28 +130,43 @@ class IliCache(QObject):
         '''
         Parses all .ili files in the given ``path`` (non-recursively)
         '''
-        model = dict()
+        models = list()
+        fileModels = list()
+        for ilifile in glob.iglob(os.path.join(path, '*.ili')):
+            try:
+                fileModels = self.parse_ili_file(ilifile, "utf-8")
+            except UnicodeDecodeError:
+                try:
+                    fileModels = self.parse_ili_file(ilifile, "latin1")
+                except UnicodeDecodeError:
+                    QgsMessageLog.logMessage(self.tr('Could not parse ili file `{ilifile}`. We suggest you to encode it in UTF-8. ({exception})'.format(ilifile=ilifile, exception=str(e))), self.tr('Projectgenerator'))
+
+            models.extend(fileModels)
+
+        self.repositories[path] = sorted(models, key=lambda m: m['version'], reverse=True)
+        self.models_changed.emit()
+
+    def parse_ili_file(self, ilipath, encoding):
+        '''
+        Parses an ili file returning models and version data
+        '''
         models = list()
         re_model = re.compile('\s*MODEL\s*([\w\d_-]+)\s.*')
         re_model_version = re.compile('VERSION "([ \w\d\._-]+)".*')
+        with open(ilipath, 'r', encoding=encoding) as file:
+            for line in file:
+                result = re_model.search(line)
+                if result:
+                    model = dict()
+                    model['name'] = result.group(1)
+                    model['version'] = ''
+                    models += [model]
 
-        for ilifile in glob.iglob(os.path.join(path, '*.ili')):
-            with open(ilifile, 'r') as file:
-                for line in file:
-                    result = re_model.search(line)
-                    if result:
-                        model = dict()
-                        model['name'] = result.group(1)
-                        model['version'] = ''
-                        models += [model]
+                result = re_model_version.search(line)
+                if result:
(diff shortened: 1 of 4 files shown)


projectgenerator/libili2pg/ilicache.py
```diff
@@ -130,29 +130,44 @@ def process_local_ili_folder(self, path):
...
-                        model['version'] = ''
-                        models += [model]
-
-                    result = re_model_version.search(line)
-                    if result:
-                        model['version'] = result.group(1)
-
-        print(repr(models[0]['name']))
+            try:
+                fileModels = self.parse_ili_file(ilifile, "utf-8")
```
m-kuhn: Nice.

I recently read that they are supposed to be utf-8 anyway somewhere in the standard, but not sure if that also applies to interlis 1, so keeping the latin1 as backup sounds ok.

Any chance to make the error more visible (maybe even if latin1 applies, for the sake of educating the user)?

m-kuhn:
Looks good to me once the error reporting is more visible

projectgenerator/libili2pg/ilicache.py
gacarrillor: Yes, we can use `QMessageBar` to show a warning if latin1 is required and an error if another encoding is required.  We'll keep logging the error messages if you don't mind.

m-kuhn: Yes, logging is good, it's just too easy to miss  info if it's only logging.

projectgenerator/libili2pg/ilicache.py
```diff
@@ -130,29 +132,48 @@ def process_local_ili_folder(self, path):
...
+            except UnicodeDecodeError:
+                try:
+                    fileModels = self.parse_ili_file(ilifile, "latin1")
+                    self.iface.messageBar().pushWarning(self.tr('Project Generator'),
```
m-kuhn: Hmm, I think it would be good to remove the dependency on `iface` within this module. The reason is that it should also be used from a standalone app / travis ci script without the whole QGIS application running behind and a different error reporting approach. The easiest is probably to switch this over to a signal that is emitted from this class and let the `GenerateProjectDialog` handle sending this to the message bar.

gacarrillor: That makes sense, right.

m-kuhn: And thinking about this again, it might make sense to put the information somewhere on the dialog and not the message bar in the main QGIS window :)

gacarrillor: Do you mean, like this? http://docs.qgis.org/testing/en/_images/dialog-with-bar.png

m-kuhn: Or in the log window?
What's your preference?

gacarrillor: Chances are users won't read the message in the log window, as there are too many lines to read already. However, we (Sergio and me) think that the orange color of a `QgsMessageBar` warning is a bit aggressive for the user and might convey the message that the plugin threw errors. 

I'd be in favor of putting it in the log window. If it's not too visible, we may play with another font color.

m-kuhn: I agree on the aggressive color, especially given that the information is only interesting for model designers and if it's a regular user (not a model designer) it might be confusing.

Sidenote: do you know QtCreator? You can select "compile output" which gives you the full log or "Issues" where you have warnings in yellow and errors in red. I really like the way this is presented there. Just if someone feels like doing some fancy UI stuff ;)

screenshot from 2017-07-13 14-44-51
screenshot from 2017-07-13 14-44-41
screenshot from 2017-07-13 14-44-02

gacarrillor: That looks really nice indeed. QtCreator is something I'd like to start using. Sometimes I want to work on QGIS source code to fix small things and find a bit difficult to traverse the code from single cpp files. To be honest, when I told Lorenz about a meeting with you in die Schweiz, one of my objectives was to fix a small bug/add a small functionality to QGIS using QtCreator.

I'm sure there are other priorities that should come first, but who knows, perhaps one day we feel adventurous. :D

PR #64, opened 2017-07-18
Show messages inside log panel in Generate dialog

gacarrillor:
This part was missing from PR #63 :)

We are showing warning messages as INFOs (blue) and error messages as WARNINGs (orange). This is just to avoid scaring users with too strong colors. I hope it's ok for you.


m-kuhn:
Yep, that should look better than the main window bar.

PR #70, opened 2017-08-01
Fix 3 issues with domain-class relations

gacarrillor:
- Wrong assignation of domains if one name is contained in the other (ej: dom1 and dom10) (fixed REGEXP).
 - Wrong treatment of extended attrs (fixed smart2 function to bring parent's attr name and pair it with child domain).
 - Missing domain attributes inherited from parents and not extended in child class (fixed list of potential referencing classes by adding classes that extend other classes).

### DIFF
--- a/projectgenerator/libqgsprojectgen/generator/postgres/domain_relations.py
+++ b/projectgenerator/libqgsprojectgen/generator/postgres/domain_relations.py
@@ -66,8 +66,10 @@ class DomainRelation(Relation):
         if self.debug: print("Classes with domain attrs:", len(models_info))
 
         # Map class ili name with its correspondent pg name
+        # Take into account classes with domain attrs and those that extend other classes,
+        #   because parent classes might have domain attrs that will be transfered to children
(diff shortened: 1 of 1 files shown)


projectgenerator/libqgsprojectgen/generator/postgres/domain_relations.py
```diff
@@ -129,7 +131,7 @@ def find_domain_relations(self, layers, conn, schema, inheritance):
                             if self.debug: print("Relation:", relation.name)
                             relations.append(relation)
 
-        if self.debug: print("final_models_info:", models_info)
+        if self.debug: print("final_models_info:", models_info_with_ext)
```
m-kuhn: Just an unrelated comment, I think we should switch all these `print` statements to the `logging` module, so we can send it to stdout, files, message log ...

And who knows, a `QgsMessageLogHandler` or similar for `logging` might even be a candidate for shipping directly with QGIS 3.

PR #74, opened 2017-08-14
Fix in Export

gacarrillor:
`modeldir` param is not needed for Export, `ili2db` will get models from the database itself (table `t_ili2db_model`)

### DIFF
--- a/projectgenerator/libili2pg/ili2pg_config.py
+++ b/projectgenerator/libili2pg/ili2pg_config.py
@@ -38,10 +38,11 @@ class BaseConfiguration(object):
         self.custom_model_directories = settings.value('CustomModelDirectories', '', str)
         self.java_path = settings.value('JavaPath', '', str)
 
-    def to_ili2db_args(self):
+    def to_ili2db_args(self, export_modeldir=True):
         args = []
-        if self.custom_model_directories_enabled and self.custom_model_directories:
-            args += ['--modeldir', self.custom_model_directories]
+        if export_modeldir:
+            if self.custom_model_directories_enabled and self.custom_model_directories:
+                args += ['--modeldir', self.custom_model_directories]
         return args
 
     @property
--- a/projectgenerator/libili2pg/iliexporter.py
+++ b/projectgenerator/libili2pg/iliexporter.py
@@ -92,6 +92,8 @@ class Exporter(QObject):
         args += ["--dbdatabase", self.configuration.database]
         args += ["--dbschema", self.configuration.schema or self.configuration.database]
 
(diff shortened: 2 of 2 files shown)


m-kuhn:
Is it never required?

gacarrillor:
Not for Export operation because ili2db reads the models from the database itself.

m-kuhn:
Does it hurt to specify it nonetheless?
The reason I ask is because this patch proposes to remove the whole `BaseConfiguration` from the export. At the moment this really only is the modeldir, but tomorrow it might be something else (that is required for the export) in there as well.

gacarrillor:
Right now, the Export operation does not need `BaseConfiguration` (namely, `modeldir` parameter). If `BaseConfiguration` gets more parameters and they are needed/useful in `Export`, I'll make sure they're being passed to Export (without `modeldir` :) ).

m-kuhn:
To clarify, "does not need" or "must not receive"?

If it doesn't hurt right now, I would rather not apply this patch. Of course, doing something like the following will be mergable right now ;)

```py
def to_ili2db_args(self, export_modeldir=True)
```

gacarrillor:
Passing the parameter is meaningless. If we want to execute the Export operation in a clean way, we shouldn't pass it. 

Another way to look at it: `BaseConfiguration` is not common anymore for all operations. Perhaps something like this?

```
def to_ili2db_args(self, operation):
  if operation == 'export':
     ...
  else:
    ...
```

m-kuhn:
There's a tiny risk involved that we will end up with an API like this

```
base_config.to_ili2db_args(operation='import_with_smart_inheritance_type_1_without_some_other_option')
```

But it also works for me if it's `def to_ili2db_args(self, operation='default'):`, because I guess for 90% of the operations, it will just be the way it is.

gacarrillor:
ili2db operations are `schemaimport`, `import`, `export`, `update`, `replace`, `delete`. We should stick to them.

m-kuhn:
What's the proposal?

gacarrillor:
```
def to_ili2db_args(self, operation):
  if operation == 'export':
     ...
  else:
    ...
```
Operation is one of `schemaimport`, `import`, `export`, `update`, `replace`, `delete`. Right now we only need to deal with `export` differently (since you don't want to remove `args += self.configuration.base_configuration.to_ili2db_args()` from the export operation).

m-kuhn:
And this forces us to always specify a string (where typos can happen) in every call to this, just because export is special :/

gacarrillor:
Should be ready now.

PR #77, opened 2017-08-17
Make 'public' the default schema

gacarrillor:
The current behavior, i.e., using database name as default schema, causes confusion to users.

### DIFF
--- a/projectgenerator/gui/generate_project.py
+++ b/projectgenerator/gui/generate_project.py
@@ -113,6 +113,7 @@ class GenerateProjectDialog(QDialog, DIALOG_UI):
             self.pg_user_line_edit.setFocus()
             return
 
+        configuration.schema = configuration.schema or 'public'
         try:
             generator = Generator(configuration.uri, configuration.schema, configuration.inheritance)
             print('uri: {} schema: {} inheritance: {}'.format(configuration.uri, configuration.schema, configuration.inheritance))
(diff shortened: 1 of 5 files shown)


m-kuhn:
-1, public should not be used for real data.

gacarrillor:
Just out of curiosity, do you happen to have some documentation about it?
Just thought about making Project Generator work as ili2pg works when no schema is given.

m-kuhn:
Nope, this is just based on some conversations with others.
I think using schemas the way we are doing now is quite sensible since it nicely decouples different projects. If using public as default, the second import will result in a mixture of topics.

gacarrillor:
Actually, with the current master if you run Generate (schema import) twice without passing a value for schema, the result will be the same as the one you described, i.e., your models will be imported onto previously imported ones.

m-kuhn:
If you run it for the same model or different models?

gacarrillor:
You could run it twice with the same model and changing your schema import parameters (e.g., inheritance type). The point here is, this PR request won't change current master branch's behavior, it only makes sure the default schema is 'public' acting like ili2pg does.

m-kuhn:
> You could run it twice with the same model and changing your schema import parameters (e.g., inheritance type).

I agree this is a certain risk, if this is a problem we could add a warning like "...schema already exists... what should be done: continue anyway, clear before import or cancel?"

I still fail to see the big advantages of using public as default (I don't consider default ili2pg configuration to be a good argument for it. ili2pg does a good job with creating databases and keeping backward compatibility, but providing a usable default configuration is definitely not something i would consider one of it's strengths)

gacarrillor:
> I agree this is a certain risk, if this is a problem we could add a warning like "...schema already exists... what should be done: continue anyway, clear before import or cancel?"

Sounds very nice.

On the other hand, `public` is the default schema for PostgreSQL and PostgreSQL-related tools respect that. Why do we need to change that for our users? It's just the expected behavior.

m-kuhn:
I think in our case we have more context than other tools have and we can and should use this knowledge to generate better and collision safe structures (by default).
We even generate a project for the user, so he doesn't even need to know which schema is being used. The only one caring for this is the DB admin, a person who I expect to be able to deal with schemas.

PR #81, opened 2017-08-23
Add calendar popup for date/datetime fields

gacarrillor:


projectgenerator/libqgsprojectgen/generator/postgres/postgrescreator.py
```diff
@@ -187,6 +188,17 @@ def layers(self):
...
+                if 'time' in fielddef['data_type'] or 'date' in fielddef['data_type']:
+                    field.widget = 'DateTime'
+                    field.widget_config['calendar_popup'] = True
+                    # timestamps don't need further configuration
+                    if 'time' in fielddef['data_type'] and not 'timestamp' in fielddef['data_type']:
+                        field.widget_config['field_format'] = 'HH:mm:ss'
+                        field.widget_config['display_format'] = 'HH:mm:ss'
+                    elif 'date' in fielddef['data_type']:
+                        field.widget_config['field_format'] = 'yyyy-MM-dd'
+                        field.widget_config['display_format'] = 'yyyy-MM-dd'
```
m-kuhn: Is this format what's normally used at your place.
Here `dd.MM.yyyy` is more common. Could possibly determined from the QGIS locale, not sure.

m-kuhn: Any feedback on this?

gacarrillor: This definitely depends on the country. Perhaps we could format dates accordingly, but I'll need to have a look at how.

m-kuhn: What's it like for you?

gacarrillor: We can use 2017-08-24, but it's more common 24-08-2017. We never use 2017.08.24 nor 24.08.2017, although I like them :).

m-kuhn: I think that would work nicely for me:

```py
>>> QLocale(QgsApplication.instance().locale()).dateFormat(QLocale.ShortFormat)
'dd.MM.yy'
>>> QLocale(QgsApplication.instance().locale()).dateTimeFormat(QLocale.ShortFormat)
'dd.MM.yy HH:mm'
```

What's it like for you?

gacarrillor:
```
'd/MM/yy'
'd/MM/yy h:mm AP'
```
We can also use `/` :)
I think that's the way to go.

(I actually use English in my laptop, so for a user like me the date format would be a bit crazy 'M/d/yy', but I guess those are edge cases).

m-kuhn: Actually, I'm also one of those edge cases...

Apparently it's only the default `en(_US)`

```py
>>> QLocale('en').dateFormat(QLocale.ShortFormat)
'M/d/yy'
>>> QLocale('en_GB').dateFormat(QLocale.ShortFormat)
'dd/MM/yyyy'
```

Let's do it anyway and blame it on Qt :P

gacarrillor: Should be ready now.

PR #84, opened 2017-08-29
Radiobutton for model and Interlis file

gacarrillor:
Current status of models/ili-file selection in Generate GUI is a bit confusing. 

This PR aims to make it explicit what parameter you want to use as user: either use modeldir/model pair or an Interlis file. This allows us to validate input since those parameters are not optional anymore once users choose what parameter they prefer.


m-kuhn:
Isn't it possible (and meaningful) to specify interlis file and model?
Not that i like the current gui too much ;)

gacarrillor:
Right, it makes sense to pass both `--models` and an ili file. We'll give this a second try.

gacarrillor:
Closed in favor of https://github.com/opengisch/projectgenerator/pull/89

PR #90, opened 2017-09-11
Make variables for import and export xtf file settings independent

seralra96:


m-kuhn:
Hmmm... I would expect that people are mostly exporting from the same database after importing into it, but I'm not sure. Do you have different experiences there?

gacarrillor:
Definitely, we just want to make XTF file setting independent for Import and Export dialogs and don't intent to touch other settings. 

The rationale is that if a user imports an XTF, modifies data in QGIS and wants to export to XTF, the XTF file that appears in the Export dialog is the same she used when importing, which may lead her to made a mistake and overwrite the XTF she imported data from.

m-kuhn:
Is it just the filename?

m-kuhn:
Ah yes, sounds good 👍

PR #106, opened 2017-09-21
Ensure a best and not arbitrary time.

juusechec:
A way to wait for postgres container to become available.
This is because 10 seconds is not sufficient to up postgres in all computers.


juusechec:
The second commit update the README.md of tests, preventing unlimited container creation, one is created in each test execution.

screenshot from 2017-09-20 21-48-16

PR #113, opened 2017-09-27
Allows user view the help according to the QGIS system language

seralra96:


m-kuhn:
what's the reason to close this? i think that's desirable!

seralra96:
I put the changes of this PR in the "Geopackage" branch to avoid future conflicts, don't worry, we won't lose what was already done in this PR. :) 

I also did 2 actions in the project generator menu that are "help" and "about", this improves the accesibility to the plugin Documentation.

m-kuhn:
I noticed, cool 👍

I've just seen that after writing and am a bit afraid of the huge codechange there, but should be alright.

PR #117, opened 2017-11-25
Support selection of multiple models in Generate/Import/Export dialogs

gacarrillor:
multiple_dialogs

### DIFF
--- a/projectgenerator/gui/multiple_models.py
+++ b/projectgenerator/gui/multiple_models.py
@@ -1,0 +1,66 @@
+# -*- coding: utf-8 -*-
+"""
+/***************************************************************************
+                              -------------------
+        begin                : 25.11.2017
+        git sha              : :%H$
+        copyright            : (C) 2017 by Germán Carrillo (BSF-Swissphoto)
+        email                : gcarrillo@linuxmail.org
+ ***************************************************************************/
+
+/***************************************************************************
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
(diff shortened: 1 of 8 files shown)


projectgenerator/gui/multiple_models.py
```diff
@@ -0,0 +1,66 @@
...
+            self.models_line_edit.setText("")
+
+    def remove_model(self):
+        for item in self.model_list.selectedItems():
+            self.model_list.takeItem(self.model_list.row(item))
+
+    def accepted(self):
+        items = [self.model_list.item(x) for x in range(self.model_list.count())]
+        models = ";".join([i.text().strip() for i in items if i.text().strip()])
+        self.parent.ili_models_line_edit.setText(models)
```
m-kuhn: What do you think about changing this to the approach to connect the parent to the dialog's accepted signal instead of pushing the model from here. Then just have a very simple method in here to generate the model string

    def model_string(self):
        items = [self.model_list.item(x) for x in range(self.model_list.count())]
        models = ";".join([i.text().strip() for i in items if i.text().strip()])
        return models

This would reduce the assumption that this dialog has to make about the environment.
No big deal though ;)

m-kuhn:
Very nice, just one small recommendation

projectgenerator/gui/multiple_models.py
gacarrillor: I think it makes sense. I've just adjusted the PR.

m-kuhn:
Thanks

PR #136, opened 2018-01-18
Relations for multi-geometry tables

### LINKED ISSUE #116: Relations for tables with several geometry fields
Relations are not being propagated to layers with multiple geometry attributes when loaded to QGIS.

For instance, the `KbS_LV95_V1_3` model from INTERLIS repositories has the `Belasteter_Standort` class with point and polygon geometries. QGIS loads the PostgreSQL table `belasteter standort` as 2 layers (one per geometry type) called equally but only the one with the point geometry gets the relations.

@m-kuhn, I think this a QGIS issue.

m-kuhn:
When a table has multiple geometry columns it will be loaded several times.
In such a situation we also need to configure the relations accordingly.

Fix #116

### DIFF
--- a/projectgenerator/libqgsprojectgen/generator/generator.py
+++ b/projectgenerator/libqgsprojectgen/generator/generator.py
@@ -162,21 +162,24 @@ class Generator:
 
     def relations(self, layers):
         relations_info = self._get_relations_info()
-        mapped_layers = {layer.name: layer for layer in layers}
+        layer_map = dict()
+        for layer in layers:
+            if layer.name not in layer_map.keys():
+                layer_map[layer.name] = list()
+            layer_map[layer.name].append(layer)
         relations = list()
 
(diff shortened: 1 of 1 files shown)


m-kuhn:
I'd like to do a unit test but don't know yet when I'll find the time for this. A review will be welcome meanwhile.

gacarrillor:
@m-kuhn Just gave this PR a try. With [our model](https://github.com/AgenciaImplementacion/LADM_COL) (Catastro_Registro_Nucleo.ili), we still don' t get relation widgets for point layers, but only for polygon layers.

Terreno (Point)
terreno_geom_punto

Terreno (Polygon)
terreno_geom_poligono

m-kuhn:
Other example

`Belasteter_Standort.StatusAltlV` is not configured correctly in the polygon version of the layer

PR #144, opened 2018-03-01
Recursive search modeldir

juusechec:
Recursive search of ilimodels in modeldir

### DIFF
--- a/projectgenerator/libili2db/ilicache.py
+++ b/projectgenerator/libili2db/ilicache.py
@@ -56,7 +56,11 @@ class IliCache(QObject):
         if path[0] == '%':
             pass
         elif os.path.isdir(path):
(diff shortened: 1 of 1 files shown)


m-kuhn:
Nice.
Is there ever a situation where we want this to be disabled?
I'm wondering if we can skip the option and always search recursively?

juusechec:
With @gacarrillor we think is possible that the list of models in all sub-directories of a complex repository, collapse unnecessary the cache and the search.

m-kuhn:
Do you need to disable recursivity on your current repository?

For context: it works reliably with all swiss models (from dozens of different domains and government levels, which are aggregated from many cascading repositories).

juusechec:
Done the suggested change.

m-kuhn:
Thanks :tada:

PR #147, opened 2018-03-05
Order layers by geometry, even for subsequent calls to project.create()

gacarrillor:
New layers will be added to a proper place in the layer tree, depending on its geometry type. 
Geometry layers are always loaded at the top of the layer tree, i.e., above groups (like `tables` and `domains`).

@m-kuhn I think we could add these methods in QGIS as well (taking rasters into account), as it may be a commonly desired behaviour.

### DIFF
--- a/projectgenerator/utils/qgis_utils.py
+++ b/projectgenerator/utils/qgis_utils.py
@@ -1,0 +1,48 @@
+# -*- coding: utf-8 -*-
+"""
+/***************************************************************************
+    begin                :    05/03/18
+    git sha              :    :%H$
+    copyright            :    (C) 2018 by Germán Carrillo (BSF-Swissphoto)
+    email                :    gcarrillo@linuxmail.org
+ ***************************************************************************/
+
+/***************************************************************************
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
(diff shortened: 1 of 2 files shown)


m-kuhn:
Good point, would possibly be worth a QEP

projectgenerator/utils/qgis_utils.py
```diff
@@ -0,0 +1,48 @@
...
+"""
+
+from qgis.core import QgsLayerTreeNode, QgsProject, QgsWkbTypes
+
+layer_order = [QgsWkbTypes.PointGeometry,
+               QgsWkbTypes.LineGeometry,
+               QgsWkbTypes.PolygonGeometry,
+               QgsWkbTypes.UnknownGeometry]
+
+def get_first_index_for_geometry_type(geometry_type, group=QgsProject.instance().layerTreeRoot()):
```
m-kuhn: The layer tree should be passed through to here to not depend on QgsProject being a singleton

m-kuhn:
Looks almost good. Some minimal docstrings for the methods that explains what they do and we are good to merge.

projectgenerator/utils/qgis_utils.py
gacarrillor: Where could we get the layer tree from?

m-kuhn: Up there you have a reference to qgis_project https://github.com/opengisch/projectgenerator/pull/147/files#diff-5e65be2a57725b9de36675393dc84b8dR70

gacarrillor: Done!

PR #149, opened 2018-03-08
Fix domains for layers with multiple geometries

### LINKED ISSUE #116: Relations for tables with several geometry fields

m-kuhn:
Fix #116

### DIFF
--- a/projectgenerator/libqgsprojectgen/dataobjects/project.py
+++ b/projectgenerator/libqgsprojectgen/dataobjects/project.py
@@ -99,7 +99,7 @@ class Project(QObject):
         dict_domains = {
             layer.real_id: layer.is_domain for layer in self.layers}
         for relation in self.relations:
-            rel = relation.create()
+            rel = relation.create(qgis_project)
             assert rel.isValid()
             qgis_relations.append(rel)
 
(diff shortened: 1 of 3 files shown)


m-kuhn:
@gacarrillor what do you think about this PR, does this look good to you and fix the problems?

gacarrillor:
It looks goot to me, this is very close to what is expected! However, some relations are still missing in our Point table, as you can see comparing both forms:

Polygon table:
terreno_poligono

Point table:
terreno_punto

gacarrillor:
@m-kuhn Just tested the merged code... What I said in my latest comment is still happening.

m-kuhn:
Yes, I am fully aware, but thought that this fix alone is worth merging as well. I'll look into the other one soon.

PR #150, opened 2018-03-08
Add the model name for each table in tables_info (for both GPKG and PG)

gacarrillor:
We'd like to use the model from where each table comes in a custom "Load Layers" dialog of our Asistente LADM_COL plugin. 

Since we can use several models in an ilid2b schema import operation, having the model in the table info from Project Generator will help us to organize available layers in a hierarchical tree.

### DIFF
--- a/projectgenerator/utils/qgis_utils.py
+++ b/projectgenerator/utils/qgis_utils.py
@@ -1,0 +1,48 @@
+# -*- coding: utf-8 -*-
+"""
+/***************************************************************************
+    begin                :    05/03/18
+    git sha              :    :%H$
+    copyright            :    (C) 2018 by Germán Carrillo (BSF-Swissphoto)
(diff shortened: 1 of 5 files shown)


gacarrillor:
@m-kuhn I can give you a hand if you decide to merge this PR...

projectgenerator/utils/qgis_utils.py
```diff
@@ -17,7 +17,7 @@
  ***************************************************************************/
 """
 
-from qgis.core import QgsLayerTreeNode, QgsWkbTypes
+from qgis.core import QgsLayerTreeNode, QgsProject, QgsWkbTypes
```
m-kuhn: I guess this import is not actually required?

gacarrillor:
After 26 days this became a nightmare to keep up-to-date. 

I'll better create another PR.

PR #153, opened 2018-03-13
In PostgreSQL/PostGIS, avoid estimatedmetadata=true when setting layer URI

gacarrillor:
Just faced an issue with a `layer.featureCount()` returning 0 when the DB actually had features, which made one of our tools run wrongly.

To prevent that scenario, this PR avoids the `estimatedmetadata=true` parameter in layer URI.

### DIFF
--- a/projectgenerator/libqgsprojectgen/generator/generator.py
+++ b/projectgenerator/libqgsprojectgen/generator/generator.py
@@ -65,7 +65,7 @@ class Generator:
             if self.tool_name == 'ili2pg':
                 provider = 'postgres'
                 if record['geometry_column']:
-                    data_source_uri = '{uri} key={primary_key} estimatedmetadata=true srid={srid} type={type} table="{schema}"."{table}" ({geometry_column})'.format(
+                    data_source_uri = '{uri} key={primary_key} srid={srid} type={type} table="{schema}"."{table}" ({geometry_column})'.format(
                         uri=self.uri,
                         primary_key=record['primary_key'],
                         srid=record['srid'],


m-kuhn:
No strong opinion about the estimated metadata, but I don't think it influences featureCount(). See https://qgis.org/api/classQgsVectorLayer.html#ac6570f424ee7e55a28de9513cb3e5c74

gacarrillor:
I've got this info from pgAdmin3:

estimatedmetadata

With `estimatedmetadata=true` in layer's URI:
```
>>> iface.activeLayer().featureCount()
0
>>> iface.activeLayer().dataProvider().featureCount()
0
```
Without `estimatedmetadata=true` in layer's URI:

```
>>> iface.activeLayer().featureCount()
3
>>> iface.activeLayer().dataProvider().featureCount()
3
```

m-kuhn:
Oh, I see.
Doesn't this increase the loading time for large layers (because it will loop over all features)?
Does it change something if you use `countSymbolFeatures()` (and wait for the appropriate signal to be emitted before continuing with whatever your plugin does)

m-kuhn:
Are you able to fix this on your end as suggested? I'd really not want to change this in here.

gacarrillor:
I'll need to have a look at it  and then let you know if it works.

For me it's a bit weird because we do check that `getFeatures()` is greater than 0 a lot of times in the plugin to proceed with other stuff.

As a developer, such method should give me an actual number of features in the layer store when it's called. As a user I would rather let the plugin take time loading my layers than preventing me from using the tools based on a wrong or out-of-date feature count.

PR #154, opened 2018-03-15
Make transactions and evaluate_default_values optional parameters in Project dataobject

gacarrillor:
We will need to turn transactions off in our LADM plugin, so having this as a parameter for `Project()` would be handy.


m-kuhn:
Are you sure you prefer to work without transactions?
You won't be able to create a new object with children in a single form.

gacarrillor:
We could switch it on if we need it. We need to turn it off when generating some geometries on the fly and storing them in our tables. Namely, we polygonize `Boundaries` and store generated polygons in our `Plot` layer. Since `Plot` layer has `NOT NULL` constraints, adding features with the generated polygons fails when transactions are enabled, and we don't want that.

PR #157, opened 2018-03-18
Make Generate Project work on GPKG DBs and PostGIS schemas even if they don't come from INTERLIS

### LINKED ISSUE #155: project generator without interlis
From [this:](https://opengisch.github.io/projectgenerator/docs/en/user-guide.html) 
_"Note that you could eventually use the plugin for generating QGIS projects from your PostgreSQL/PostGIS schemas, even if they are not related to Interlis in any way, however, Project Generator exploits Interlis expresiveness..."_ 
I read, that a project could be generated from ANY schema - although it may not be as fancy as with interlis-extras - I thought it should load at least a list of layers from the PostGIS tables and views, but I got these messages and error and it seem that the plugin looks for interlis models in any way? 

```
psycopg2.ProgrammingError: FEHLER:  Relation „isy_in.t_ili2db_column_prop“ existiert nicht LINE 12:            LEFT JOIN isy_in.t_ili2db_column_prop unit ON c....                               ^ 
  File "C:/Users/TS/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\projectgenerator\gui\generate_project.py", line 251, in accepted
    available_layers = generator.layers()
  File "C:/Users/TS/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\projectgenerator\libqgsprojectgen\generator\generator.py", line 104, in layers
    fields_info = self._get_fields_info(record['tablename'])
  File "C:/Users/TS/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\projectgenerator\libqgsprojectgen\generator\generator.py", line 244, in _get_fields_info
    return self._db_connector.get_fields_info(table_name)
  File "C:/Users/TS/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\projectgenerator\libqgsprojectgen\dbconnector\pg_connector.py", line 161, in get_fields_info
    """.format(schema=self.schema, table=table_name))
  File "C:\OSGEO4~1\apps\Python36\lib\site-packages\psycopg2\extras.py", line 144, in execute
    return super(DictCursor, self).execute(query, vars)
psycopg2.ProgrammingError: FEHLER:  Relation „isy_in.t_ili2db_column_prop“ existiert nicht
LINE 12:             LEFT JOIN isy_in.t_ili2db_column_prop unit ON c....
                               ^
```		   

Do I have a misunderstanding of the project generator, is it not (yet) meant to be generic? - Which would be awesome by the way!! 
(I tried to combine it with my [experiments on importing ISYBAU XML](https://github.com/tschuettenberg/isybau2qgep) )

gacarrillor:
Also: 

- Allow table names with capital letters
- Don't disable controls after connection error or schema existence messages are shown, so the user can retry on the same dialog.
- Tests included.

Fix #155


projectgenerator/libqgsprojectgen/dataobjects/project.py
```diff
@@ -97,7 +97,7 @@ def create(self, path: str, qgis_project: QgsProject):
         qgis_relations = list(
             qgis_project.relationManager().relations().values())
         dict_domains = {
-            layer.real_id: layer.is_domain for layer in self.layers}
+            layer.layer.id(): layer.is_domain for layer in self.layers}
```
gacarrillor: I had to avoid `real_id` in this line because I was getting this nasty error:

```
SystemError: error return without exception set
  File "/home/germap/.local/share/QGIS/QGIS3/profiles/default/python/plugins/projectgenerator/gui/generate_project.py", line 290, in accepted
    project.create(None, qgis_project)
  File "/home/germap/.local/share/QGIS/QGIS3/profiles/default/python/plugins/projectgenerator/libqgsprojectgen/dataobjects/project.py", line 100, in create
    layer.real_id: layer.is_domain for layer in self.layers}
  File "/home/germap/.local/share/QGIS/QGIS3/profiles/default/python/plugins/projectgenerator/libqgsprojectgen/dataobjects/project.py", line 100, in 
    layer.real_id: layer.is_domain for layer in self.layers}
  File "/home/germap/.local/share/QGIS/QGIS3/profiles/default/python/plugins/projectgenerator/libqgsprojectgen/dataobjects/layers.py", line 111, in real_id
    if self.__layer:
SystemError: error return without exception set
```

Anyways, `layer.layer` at this point is already a `QgsVectorLayer`, so we don't need to do the check that the property `real_id` does internally.

m-kuhn:
Good move, can you add a unit test so we are sure we don't run into this again?

projectgenerator/gui/generate_project.py
```diff
@@ -166,40 +166,41 @@ def accepted(self):
...
                 self.txtStdout.setText(
-                    self.tr('Please set a database user before creating the project.'))
+                    self.tr('\n\nPlease set a database user before creating the project.'))
                 self.pg_user_line_edit.setFocus()
                 return
         elif self.type_combo_box.currentData() in ['ili2gpkg', 'gpkg']:
             if not configuration.dbfile or self.gpkg_file_line_edit.validator().validate(configuration.dbfile, 0)[0] != QValidator.Acceptable:
                 self.txtStdout.setText(
-                    self.tr('Please set a valid database file before creating the project.'))
+                    self.tr('\n\nPlease set a valid database file before creating the project.'))
```
m-kuhn: These changes really should go in a separate pull request. It's ok to leave them this time, but please keep it in mind for next time.

projectgenerator/libqgsprojectgen/dataobjects/project.py
m-kuhn: This looks very strange and it will be hard to know in the future, why this workaround was introduced.

Can we embed the call to `layer.real_id` in a `try: ...  except SystemError:` block and only revert to the other behavior when this actually happens? I think this way it will be more transparent what's going on and we can revert easier in the future.

gacarrillor: As I told you, at this point `layer.layer` is already a `QgsVectorLayer`, so we can access its methods without trouble.

This simple dict comprehension will become a bit ugly inside a try catch... but, ok.

BTW, no clue about the `SystemError`. Also, an Internet search didn't lead me to anything useful.

m-kuhn: What will happen if this is in the future - for whatever reason - called in a different context? Will we at least get a reasonable error in such a scenario?

m-kuhn:
Let me know when this is ready.

projectgenerator/libqgsprojectgen/dataobjects/project.py
gacarrillor: Since we above call:

```
for layer in self.layers:
    qgis_layer = layer.create()
```

at the point we use `layer.layer.id()`, `layer.layer` exists. We shouldn't get errors there. 

On the contrary, with the current situation a very strange error is happening. 

The change we're making in this PR avoids such strange error. I'll comment in that line why we changed from `layer.real_id` to `layer.layer.id()`.

gacarrillor:
@m-kuhn This PR is finally finished from our side!

projectgenerator/gui/generate_project.py
m-kuhn: Thinking about this again, these are currently breaking all the translations, (and the `\n\n` part is not something a translator should have to take care off). This needs to be implemented differently, with the `\n\n` part outside the translated text)

projectgenerator/libqgsprojectgen/dataobjects/project.py
m-kuhn: Alright, calling `layer.create()` is accepted as a reason to leave this as is.
(at the same time *strange error* is not accepted as a reason because *strange errors* tend to come back at one later on and bite one in the back.) ;)

projectgenerator/libqgsprojectgen/dbconnector/db_connector.py
```diff
@@ -31,8 +31,14 @@ def map_data_types(self, data_type):
         '''Map provider date/time types to QGIS date/time types'''
         return None
 
+    def db_or_schema_exists(self):
+        '''Whether the DB (for GPKG) or schema (for PG) exists or not.'''
+        return False
```
m-kuhn: Should this raise a `NotImplementedError`? Or is it expected that not every subclass implements this?
(minor issue)

projectgenerator/tests/test_projectgen_generic_databases.py
```diff
@@ -0,0 +1,185 @@
...
+        generator = Generator('ili2pg', uri, 'smart1', 'no_interlis_schema')
+        layers = generator.layers()
+
+        self.assertEqual(len(layers), 2)
+        self.assertEqual(len(generator.relations(layers)), 1)
+
+        for layer in layers:
+            self.assertIsNone(layer.geometry_column)
+
+        cur.execute("DROP SCHEMA no_interlis_schema CASCADE;")
```
m-kuhn: Better done in a `try: finally:` block.

I don't mind merging this as is for now, so this comment is an optional improvement proposal (and something to keep in mind for the future).

projectgenerator/libqgsprojectgen/dataobjects/project.py
gacarrillor: Yeah, since the first comment it was stated that `layer.layer` was already a `QgsVectorLayer`, so the change was safe anyways.

gacarrillor:
Ready for a review!

projectgenerator/tests/test_projectgen_generic_databases.py
juusechec: agree

m-kuhn: > Especially because the next tests are executed indifferently of unit test result!!

I think they are (but if you want to be sure better look it up ;) )

juusechec: No. The tests are working. Need a enchacement. But it could go in a newer PR like a refactor.

juusechec: it's ready https://github.com/opengisch/projectgenerator/pull/157/commits/3003a94f160a8139f1b3f2f9499c5d46b5a16a9c

projectgenerator/tests/test_projectgen_generic_databases.py
```diff
@@ -66,9 +66,10 @@ def test_postgres_db_with_empty_schema(self):
         cur.execute("CREATE SCHEMA IF NOT EXISTS empty_schema;")
 
         generator = Generator('ili2pg', uri, 'smart1', 'empty_schema')
-        self.assertEqual(len(generator.layers()), 0)
-
-        cur.execute("DROP SCHEMA empty_schema CASCADE;")
+        try:
```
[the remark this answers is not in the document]
juusechec: you're right

PR #160, opened 2018-03-29
Use relation.id instead of referencedlayer.name as identificator

signedav:
In the QGIS VectorLayerProperties at AttributesForm-configuration as an item in the drop-list, there will be the relation.id as label and identificator. 

Fix of Redmines #18290

**I had not the yet possiblity to test**

### Why that?
In that form we work always with relation-id and not with the relation-name (altough it's saved under "FieldNameRole" - this is because the other fields use the name as unique identificator). That's why we exepct in the name-variable of `mLayer->editFormConfig().tabs()``the relation-id. This is why the projectgenerator would write after this commit there the relation-id and not the referencedlayer-name. It would have worked if it would have written the relation-name (but only because it's equal to relation-id).

### Wieso das?
*Same explanation in german:*
In den LayerProperties wird immer mit der RelationId gearbeitet, nicht mit dem RelationName (selbst wenn das Feld, wo diese Id gespeichert wird wärend der Konfiguration unter FieldNameRole abgespeichert wird - dies weil die normalen Felder den Name als Identifikation verwenden, der - im Gegensatz zum RelationName - eindeutig ist). Diese Funktionsweise ist schon bei 2.18 *

Somit soll im QgsAttributeEditorElement ( aus dem mLayer->editFormConfig().tabs() ) auch in der name-Variable dieser als Id verwendete Name des Feldes bzw. die Id der Relation stehen. Dort steht aber im Falle eines Projektes, das vom Projektgenerator erstellt wird immer der ReferencingLayerName. Würde hier der RelationName drin stehen (was auch falsch wäre), würde es Funktionieren, da der Projektgenerator die RelationId als RelationName nimmt.

/* man könnte darüber nachdenken, den Name anstelle der Id in den TreeItems anzuzeigen. Müsste diesen aber zusätzlich zwischenspeichern.

### Possible QGIS improvements
We could think about to have the relation-name as the labeling of the items in the drag and the drop list. But then we would need to save them additionally in the item struct. It should be a second priority improvement because it's not fixing any bug. It's just the look.

### DIFF
--- a/projectgenerator/libqgsprojectgen/dataobjects/form.py
+++ b/projectgenerator/libqgsprojectgen/dataobjects/form.py
@@ -96,11 +96,15 @@ class FormFieldWidget(object):
 
 class FormRelationWidget(object):
 
-    def __init__(self, name, relation):
-        self.name = name
+    def __init__(self, relation):
         self.relation = relation
 
     def create(self, parent, layer):
-        widget = QgsAttributeEditorRelation(
-            self.name, self.relation.id, parent)
+        try:
+            widget = QgsAttributeEditorRelation(
+               self.relation.id, parent)
+        except TypeError:
+            # Feed deprecated API for 3.0.0 and 3.0.1
+            widget = QgsAttributeEditorRelation(
+               self.relation.id, self.relation.id, parent)
         return widget
--- a/projectgenerator/libqgsprojectgen/dataobjects/layers.py
+++ b/projectgenerator/libqgsprojectgen/dataobjects/layers.py
@@ -89,8 +89,7 @@ class Layer(object):
             for relation in project.relations:
                 if relation.referenced_layer == self:
                     tab = FormTab(relation.referencing_layer.name)
-                    widget = FormRelationWidget(
-                        relation.id, relation)
+                    widget = FormRelationWidget(relation)
                     tab.addChild(widget)
                     self.__form.add_element(tab)
         else:
--- a/projectgenerator/libqgsprojectgen/dataobjects/layers.py
+++ b/projectgenerator/libqgsprojectgen/dataobjects/layers.py
@@ -90,7 +90,7 @@ class Layer(object):
                 if relation.referenced_layer == self:
                     tab = FormTab(relation.referencing_layer.name)
                     widget = FormRelationWidget(
-                        relation.referencing_layer.name, relation)
+                        relation.id, relation)
                     tab.addChild(widget)
                     self.__form.add_element(tab)
         else:


m-kuhn:
Thanks @signedav 

I think I do not completely follow. What I understand is this: Within the method which is called here, a new [QgsAttributeEditorRelation is created](
https://github.com/opengisch/projectgenerator/blob/3d66530b35cf096a7587f967e7118d64820a0425/projectgenerator/libqgsprojectgen/dataobjects/form.py#L105). Its constructor signature is [QgsAttributeEditorRelation::QgsAttributeEditorRelation(const QString &name, const QString &relationId, QgsAttributeEditorElement *parent](https://qgis.org/api/classQgsAttributeEditorRelation.html#a847c72df31782417bafd516a87f1922e).

With this patch here, we are forcing the first parameter to be the id (and no longer the name). Before: `QgsAttributeEditorRelation(name, id, parent)` after: `QgsAttributeEditorRelation(id, id, parent)`. If the API needs to be used like this, this looks to me like a bug in the QGIS API and not like a visual glitch.

Am I missing something?

signedav:
The situation is that we use the name as the identification in the AttributesForm-Config (and in 2.18 in Fields&Forms). In the Fields (DB Fields) this is the normal thing, because the name is unique. I assume the relation-id has been used as "name" there, because it's unique (what relation-name is not). That's why we use the relation-id as the "name" there, and that's why we should deliver the id as the name and not as the id.

Actually in the current implementation it would work as well when we deliver the relation-name as the name because the relation-name is equal the relation-id. 

I do agree, that it's quite confusing and not solved that nice in QGIS. But it's only the naming there and so I'm not sure if we should change that.

Here how it's currently working with this Relations in the Dragndrop-Configuration:

relations

The fields name in "Available Widgets" have the same naming like in the "Form Layout".
But in the Relations the id is used as displayed name in "Available Widgets" and usually in "Form Layout" as well. But the generated items have as name (and the identification) the name of the referencedLayer.

signedav:
It still could probably be solved by QGIS, when we read the relation-id instead of the name when loading the configuration. But then at least I would take the relationname here instead of the referencedlayer-name. No?

m-kuhn:
The issue I assume on QGIS side is that the configuration somehow loads the name and saves it as the ID and therefore corrupts data. Is this correct?

> It still could probably be solved by QGIS, when we read the relation-id instead of the name when loading the configuration. But then at least I would take the relationname here instead of the referencedlayer-name. No?

We are generating data in here, so we need to check if we also generate the relation name in some other place within the plugin.

signedav:
> The issue I assume on QGIS side is that the configuration somehow loads the name and saves it as the ID and therefore corrupts data. Is this correct?

It's usually done like that: 
The given name is used as itemname. But because in every case (except this one here) the given name is the relation-id, we could say: The given id is used as itemname. While this itemname is actually the identificator of the item, so itemname is not really a good naming, but anyway:
Because the layerpropertyconfigurator assumes that everyone has written the id in this variable, it writes this back in the end as the id. Usually no problem, but here the it receives not the reference-id but the referencedlayername as this itemname (itemidentification) and writes this referencedlayername back to the reference-id, what corrupts the data.

So I see two options to solve and I would prefer to make both to have it proper in the end:
- QGIS should take (in case of relation) not the name of the `QgsAttributeEditorElement` in case of relation but the relation.id as itemname.
- projectgenerator should not define the referencedlayername as the itemname. 

One sollution would solve the issue. But I would make both.

signedav:
Just to clarify: 
In the current version, when the project is generated with projectgenerator and then you configure your widgets on the form by yourself. Everything works fine. The only issue on projectgenerator is how it configures the widgets on the form.

m-kuhn:
On hold. Waiting for upstream fixes in the QGIS API and a subsequent fix on a clean API (potentially with a backwards compatible fix in projectgenerator).

signedav:
This should no be merged before that https://github.com/qgis/QGIS/pull/6716

PR #163, opened 2018-04-02
Simple script to execute tests locally

juusechec:
This are able to been executed since any directory:
**./scripts/run-tests.sh** or **./run-tests.sh** or **../scripts/run-tests.sh**

### DIFF
--- a/scripts/run-tests.sh
+++ b/scripts/run-tests.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+MY_PATH="`dirname \"$0\"`"              # relative
+MY_PATH="`( cd \"$MY_PATH\" && pwd )`"  # absolutized and normalized
+
+cd "$MY_PATH"/..
+
+export TRAVIS_BUILD_DIR=$PWD
+docker-compose -f .docker/docker-compose.travis.yml run qgis /usr/src/.docker/run-docker-tests.sh
+docker-compose -f .docker/docker-compose.travis.yml rm -s -f
--- a/scripts/run-tests.sh
+++ b/scripts/run-tests.sh
@@ -1,9 +1,6 @@
 #!/bin/bash
 
-ROOTPATH=$(git rev-parse --show-toplevel)
-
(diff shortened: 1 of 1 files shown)


scripts/run-tests.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+MY_PATH="`dirname \"$0\"`"              # relative
+MY_PATH="`( cd \"$MY_PATH\" && pwd )`"  # absolutized and normalized
```
m-kuhn:
```
ROOTDIR=$(git rev-parse --show-toplevel)
```

and then

```
pushd $ROOTPATH
...
popd
```

Helps to keep things a bit more clean and tidy :)

juusechec: - with this recursive path algorithm I can execute the tests even outside a git repo.
- pushd is more used to save all locations and browse them, like good alternative to subshells http://tldp.org/LDP/abs/html/subshells.html. I thing if i only move one time cd is best, less lines to do the same.

m-kuhn: good points.
I thought cd will also affect the path of outside the script, but apparently not, so no reason for `pushd`.
Concerning path, what do you think about this snippet (also less lines ;) )

```
DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
```

juusechec: it's ok for me
what do you think about this?
```bash
#!/bin/bash

cd $(dirname $0)/..
export TRAVIS_BUILD_DIR=$PWD
docker-compose -f .docker/docker-compose.travis.yml run qgis /usr/src/.docker/run-docker-tests.sh
docker-compose -f .docker/docker-compose.travis.yml rm -s -f
```

juusechec:
the suggested changes are made https://github.com/opengisch/projectgenerator/pull/163/commits/e48f17bc357dfad5551c2c1a16f250b8737242bd

PR #165, opened 2018-04-02
Update psycopg2

juusechec:
Updating this, warnings of psycopg2 are solved.
It is in issue: https://github.com/psycopg/psycopg2/issues/498
More detailed in https://github.com/AgenciaImplementacion/projectgenerator/issues/3

### DIFF
--- a/.docker/Dockerfile
+++ b/.docker/Dockerfile
@@ -3,7 +3,13 @@ FROM  qgis/qgis:${QGIS_TEST_VERSION}
 MAINTAINER Matthias Kuhn <matthias@opengis.ch>
 
 RUN apt-get update && \
-    apt-get -y install default-jre 
+    apt-get -y install default-jre
+
+RUN apt-get -y install python3-pip
+
+RUN pip3 install --upgrade pip
+
+RUN pip3 install --upgrade psycopg2
 
 ENV LANG=C.UTF-8
 


juusechec:
This was successfully added in Asistente-LADM_COL https://github.com/AgenciaImplementacion/Asistente-LADM_COL/pull/28

m-kuhn:
Sounds good, but shouldn't the version be added to `requirements.txt` (and that one installed if not yet done?) instead?

https://github.com/opengisch/projectgenerator/blob/master/requirements.txt

juusechec:
https://github.com/opengisch/projectgenerator/blob/master/requirements.txt file is installing dependencies outside of docker qgis container (in travis instance). https://github.com/opengisch/projectgenerator/blob/master/.travis.yml#L12

The psycopg2 are installed in a previous step of docker image generation https://github.com/qgis/QGIS/blob/master/.docker/qgis3-build-deps.dockerfile#L92

juusechec:
Optionally, we can use https://github.com/opengisch/projectgenerator/blob/master/requirements.txt as a list in https://github.com/opengisch/projectgenerator/blob/master/.docker/Dockerfile#L7 to provide dependencies to the tests container. But the process is "update" (or install if not installed) for container, and not only pure "install".

Some better idea?

juusechec:
I'll work in that

juusechec:
An approach to do that https://github.com/opengisch/projectgenerator/pull/180

PR #174, opened 2018-04-11
Improve layer order by taking rasters and tables into account

gacarrillor:
+ Add your suggestions from #169 
 + Take into account rasters and tables
 + If this situation occurs in the layer tree:
     + Table/raster (or whatever lower than polygons in the desired order)
     + Polygon layer

    And we want to insert a Line layer, don't insert it right before the polygon, but right before the table/raster. This works in a generic way, regardless of the types used in the example above.

### DIFF
--- a/projectgenerator/utils/qgis_utils.py
+++ b/projectgenerator/utils/qgis_utils.py
@@ -17,43 +17,75 @@
  ***************************************************************************/
 """
 
-from qgis.core import QgsLayerTreeNode, QgsWkbTypes
+from qgis.core import QgsLayerTreeNode, QgsWkbTypes, QgsMapLayer
 
-layer_order = [QgsWkbTypes.PointGeometry,
-               QgsWkbTypes.LineGeometry,
-               QgsWkbTypes.PolygonGeometry,
-               QgsWkbTypes.UnknownGeometry]
+layer_order = ['point',  #QgsWkbTypes.PointGeometry
+               'line',  #QgsWkbTypes.LineGeometry
+               'polygon',  #QgsWkbTypes.PolygonGeometry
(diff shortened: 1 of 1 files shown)


projectgenerator/utils/qgis_utils.py
```diff
@@ -17,43 +17,75 @@
...
+    index = None
+    for layer_type in layer_order[layer_order.index(get_layer_type(layer)):]:  # Slice from current until last
+        index = get_first_index_for_layer_type(layer_type, group)
         if index is not None:
-            break
+            indices.append(index)
+
+    if indices:
+        index = min(indices)
+    print("Index for", layer.name(), -1 if index is None else index)
```
m-kuhn: Can you guard that in an `if debug`? `print`'s sometimes can crash QGIS when executed at bad times. I recently proposed to smear some lamb blood on the door everytime one uses print in a QGIS plugin :P

m-kuhn:
Looks much better than my approach!

m-kuhn:
> Don't fully understand the problem of having tables "behind" rasters.

Tables are the data one works with (should be easy to find -> on top) whereas basemaps are just there to look good, so they can go anywhere in the legend.

projectgenerator/utils/qgis_utils.py
gacarrillor: Yeap, my fault, I should have deleted that line.

gacarrillor:
Oh, I see. To be honest I've always used tables at the bottom. Fortunately, we have our own group for tables, because we have too many of them. I think that layer type order could even be customizable, but not for Project Generator :D

m-kuhn:
It was just a personal feeling. Maybe others are more used to "your" style. Let's see if someone complains ;)

PR #180, opened 2018-04-17
Using requirements.txt inside docker

juusechec:
Using requirements.txt to update or install python packages.

This is related with https://github.com/opengisch/projectgenerator/pull/165

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-psycopg2
+psycopg2==2.7.4
 nose2
 pyyaml
 future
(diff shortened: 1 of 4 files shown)


juusechec:
The changes in docker-compose is a solution due problems generated in "context" of docker by having a separated `docker-compose.travis.yml` https://github.com/moby/moby/issues/34986

requirements.txt
```diff
@@ -1,4 +1,4 @@
-psycopg2
+psycopg2==2.7.4
```
m-kuhn: Is there a problem with newer versions?

m-kuhn:
I like :+1:

requirements.txt
juusechec: No, it can be psycopg2>=2.7.4

juusechec: This is the latest version available, released on Feb 6.

PR #181, opened 2018-04-19
Set related layer on n-m cardinality

### LINKED ISSUE #24: n-m relations
Check if it is available in ili2pg metadata and then do it

signedav:
In the attribute form configuration we set the FK of the layer referencing to the same layer (networktable) as the current layer is.

So this would look like this:
cardinality1
Instead of "Many to one" relation everywhere.

While the hinweisweiteredokumente_hinweis_fkey points to hinweisweiteredokumente_ursprung_fkey and hinweisweiteredokumente_ursprung_fkey to hinweisweiteredokumente_hinweis_fkey:
```
      <widgets>
        <widget name="hinweisweiteredokumente_hinweis_fkey">
          <config type="Map">
            <Option name="nm-rel" type="QString" value="hinweisweiteredokumente_ursprung_fkey"/>
          </config>
        </widget>
        <widget name="hinweisweiteredokumente_ursprung_fkey">
          <config type="Map">
            <Option name="nm-rel" type="QString" value="hinweisweiteredokumente_hinweis_fkey"/>
          </config>
        </widget>
        <widget name="multilingualuri_dokument_textimweb_fkey">
          <config type="Map">
            <Option name="nm-rel" type="QString" value="multilingualuri_amt_amtimweb_fkey"/>
          </config>
        </widget>
        <widget name="typ_dokument_vorschrift_fkey">
          <config type="Map">
            <Option name="nm-rel" type="QString" value="typ_dokument_typ_fkey"/>
          </config>
        </widget>
      </widgets>
```
Looks like this:
cardinality2

Because I'm not to familiar with the project, can you tell me if this is what was requested @m-kuhn ?

Fix #24

**No tests yet in this PR**

### DIFF
--- a/projectgenerator/libqgsprojectgen/dataobjects/layers.py
+++ b/projectgenerator/libqgsprojectgen/dataobjects/layers.py
@@ -24,7 +24,7 @@ from qgis.PyQt.QtCore import QCoreApplication
 
 class Layer(object):
 
-    def __init__(self, provider, uri, name, geometry_column=None, wkb_type=QgsWkbTypes.Unknown, alias='', is_domain=False):
+    def __init__(self, provider, uri, name, geometry_column=None, wkb_type=QgsWkbTypes.Unknown, alias='', is_domain=False, is_nmrel=False):
         self.provider = provider
         self.uri = uri
         self.name = name
@@ -34,6 +34,7 @@ class Layer(object):
         self.__layer = None
         self.fields = list()
         self.is_domain = is_domain
+        self.is_nmrel = is_nmrel
         self.__form = Form()
 
     def dump(self):
--- a/projectgenerator/libqgsprojectgen/dataobjects/layers.py
+++ b/projectgenerator/libqgsprojectgen/dataobjects/layers.py
@@ -34,11 +34,11 @@ class Layer(object):
         self.__layer = None
         self.fields = list()
         self.is_domain = is_domain
-        """
-        If is_nmrel is set to true it is a junction table in a N:M relation.
-        Or in ili2db terms, the table is marked as ASSOCIATION in t_ili2db_table_prop.settings.
-        """
+
         self.is_nmrel = is_nmrel
+        """ If is_nmrel is set to true it is a junction table in a N:M relation.
+        Or in ili2db terms, the table is marked as ASSOCIATION in t_ili2db_table_prop.settings. """
+
         self.__form = Form()
 
     def dump(self):
--- a/projectgenerator/libqgsprojectgen/dataobjects/layers.py
+++ b/projectgenerator/libqgsprojectgen/dataobjects/layers.py
@@ -34,6 +34,10 @@ class Layer(object):
         self.__layer = None
         self.fields = list()
         self.is_domain = is_domain
+        """
+        If is_nmrel is set to true it is a junction table in a N:M relation.
+        Or in ili2db terms, the table is marked as ASSOCIATION in t_ili2db_table_prop.settings.
+        """
         self.is_nmrel = is_nmrel
         self.__form = Form()
 
@@ -42,6 +46,7 @@ class Layer(object):
         definition['provider'] = self.provider
         definition['uri'] = self.uri
         definition['isdomain'] = self.is_domain
+        definition['isnmrel'] = self.is_nmrel
         definition['form'] = self.__form.dump()
         return definition
 
@@ -49,6 +54,7 @@ class Layer(object):
         self.provider = definition['provider']
         self.uri = definition['uri']
         self.is_domain = definition['isdomain']
+        self.is_nmrel = definition['isnmrel']
         self.__form.load(definition['form'])
 
     def create(self):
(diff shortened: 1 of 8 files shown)


signedav:
Dies wird übrigens auch für die Relations zur Tabelle **multilingualuri** gemacht, auch wenn sie nicht als ` ch.ehi.ili2db.tableKind=ASSOCIATION` in `T_ILI2DB_TABLE_PROP` aufgelistet ist sondern als `STRUCTURE`. Die hier commitete Lösung schaut aber nicht auf diesen Wert sondern wählt den direkten Key sobald die Möglichkeit dazu besteht.

m-kuhn:
In case we have a table with various value lists attached, should they be treated as N:M relations?

Let's say we have a table `Buildings` with attributes `status` and `owner`, both of which are foreign keys to other tables (value lists/domains).

I think the current solution would propose a N:M relation between status and owner?

signedav:
Yes @m-kuhn in the current situation it links the relations whenever possible. I agree that in the mentioned case this is not what we want. So I guess we need to care about the `tableKind=ASSOCIATION`.

signedav:
Two questions on the latest push:
1. should we rename `is_domain` in general because the `settings` from `t_ili2db_table_prop` are not used anymore only for `is_domain` but for `is_nmrel` as well. Or is that clear enough?
2. I make another loop here https://github.com/opengisch/projectgenerator/pull/181/commits/e0898f7b507bba504521ea9530af46646479333a#diff-293e9e374ed6cbd37b23e9aa1b03e914R126 to get  the `Layer()` object matching the `QgsVectorLayer` `other_relation.referenced_layer` to be able to read the `is_nmrel`. Is there a nicer way to do that?

projectgenerator/libqgsprojectgen/dataobjects/layers.py
```diff
@@ -34,6 +34,7 @@ def __init__(self, provider, uri, name, geometry_column=None, wkb_type=QgsWkbTyp
         self.__layer = None
         self.fields = list()
         self.is_domain = is_domain
+        self.is_nmrel = is_nmrel
```
m-kuhn: If this is true, this is a junction table? I.e. the generated table with the two foreign keys to the tables that have the N:M relation?

signedav: Yes. According to `t_ili2db_table_prop.settings` are `ASSOCIATION`.

m-kuhn: Could you add a comment on this variable?

```
self.is_nmrel = is_nmrel
""""If a layer has is_nmrel set to true it is a junction table in a N:M relation. Or in ili2db terms, the table (or two foreign keys on this table?) is marked as ASSOCIATION in t_ili2db_table_prop.settings."""
```
(Please review the docstring, I'm not sure it's correct!!)

signedav:
1. Well that would be an idea as well, but what I meant was to rename the column-alias (`p.setting AS is_domain` etc.) because it's not used anymore only for the boolean is_domain. But it's a detail.
2. I originally wanted to implement it in layers.py or even form.py but I got problems there to set the editFormConfig (`qgis_project.mapLayer(layer.layer.id()).editFormConfig()`).

m-kuhn:
1. Ah, I see. Sounds good to rename for clarity.
2. Strange, isn't the form config set within layers.py (`Layer.create_form`)?

signedav:
1. Will do it.
2. Yes it is. It's called right after I put the nm-rel into the editFormConfig. You would put it into that function `Layer.create_form`? 
What I meant is, that I had problems  to set the editFormConfig in like `Layer.post_generate` because there is not a qgis_project. And I don't know if it makes sense to pass it all the way there.

m-kuhn:
`Layer.create_form` is called just right after the code you added with the `qgis_project`, I guess it could safely be done in there.

https://github.com/opengisch/projectgenerator/commit/e0898f7b507bba504521ea9530af46646479333a#diff-293e9e374ed6cbd37b23e9aa1b03e914R133

projectgenerator/libqgsprojectgen/dataobjects/layers.py
```diff
@@ -34,6 +34,10 @@ def __init__(self, provider, uri, name, geometry_column=None, wkb_type=QgsWkbTyp
         self.__layer = None
         self.fields = list()
         self.is_domain = is_domain
+        """
+        If is_nmrel is set to true it is a junction table in a N:M relation.
+        Or in ili2db terms, the table is marked as ASSOCIATION in t_ili2db_table_prop.settings.
+        """
         self.is_nmrel = is_nmrel
```
m-kuhn: Docstrings in python come after the variable. If you find this as unintuitive to read as me, add an empty line before and after

```
self.is_domain = is_domain

self.is_nmrel = is_nmrel
"""
If is_nmrel is set to true it is a junction table in a N:M relation.
Or in ili2db terms, the table is marked as ASSOCIATION in t_ili2db_table_prop.settings.
"""

```

projectgenerator/tests/test_projectgen.py
```diff
@@ -258,16 +258,98 @@ def test_ranges_geopackage(self):
 
         self.assertEqual(count, 1)
 
+    def test_rmrel_postgis(self):
```
m-kuhn: rmrel or nmrel?

PR #183, opened 2018-04-26
Do not freeze when a table has more than two geometries

### LINKED ISSUE #178: Project Generator v3.0.6 freezes loading relations for tables with more than one geometry column
v3.0.5 works fine.

The problem we face is with `Terreno` table (it has polygon and point geometries) from the LADM_COL [Catastro_Registro_Nucleo](https://github.com/AgenciaImplementacion/LADM_COL/tree/master/Catastro_Multiproposito) model.

m-kuhn:
Fix #178

Tested with the Nucleo model


gacarrillor:
Confirmed, it works now, thanks!

Issue #135, opened 2018-01-12
referenced objects not via t_id (FK)


zigertiger:
maybe already in the pipeline...

object referencing is established by choosing t_id (FK) which is not feasible. referenced objects should be referencable via defined field values of the referenced table (e.g. type names).

![referenced-objects](https://user-images.githubusercontent.com/25387992/34865371-bf5a0c12-f778-11e7-81fc-8194f3fa85b7.png)

m-kuhn:
Depends on https://github.com/opengisch/projectgenerator/issues/128

signedav:
It's about to load the meta attribute for the display expression defined in the interlis file (expl. https://github.com/opengisch/projectgenerator/blob/master/projectgenerator/tests/testdata/ilimodels/ExceptionalLoadsRoute_V1.ili)
```
!!@ dispExpression="TypeOfRoute"
CLASS ...
```
This would be written into an ili2db table, where the projectgenerator form-part can read it from.

**Note to myself:**
To have the project select in the projectgenerator the Model `ExceptionalLoadsRoute_LV95_V1` and import the Interlis Transfer Files `/home/david/doc/clients/SH/ExceptionalLoadsRoute_Catalogue_V1.xml`

m-kuhn:
> This would be written into an ili2db table, where the projectgenerator form-part can read it from.

Not exactly this one, it's another ili2db meta table

signedav:
`t_ili2db_meta_attr`

signedav:
Done with PR #189

PR #189, opened 2018-05-03
Read meta attrs from the t_ili2db_meta_attrs table

signedav:
And use the dispExpression (if existing) as the displayExpression of the relation reference widget.

Not completed: 
- Needs more implementation
- Needs tests

#135

### DIFF
--- a/projectgenerator/libqgsprojectgen/dataobjects/layers.py
+++ b/projectgenerator/libqgsprojectgen/dataobjects/layers.py
@@ -24,7 +24,7 @@ from qgis.PyQt.QtCore import QCoreApplication
 
 class Layer(object):
 
-    def __init__(self, provider, uri, name, geometry_column=None, wkb_type=QgsWkbTypes.Unknown, alias='', is_domain=False, is_nmrel=False):
+    def __init__(self, provider, uri, name, geometry_column=None, wkb_type=QgsWkbTypes.Unknown, alias='', is_domain=False, is_nmrel=False, display_expression=''):
         self.provider = provider
         self.uri = uri
         self.name = name
@@ -39,6 +39,8 @@ class Layer(object):
         """ If is_nmrel is set to true it is a junction table in a N:M relation.
         Or in ili2db terms, the table is marked as ASSOCIATION in t_ili2db_table_prop.settings. """
 
+        self.display_expression = display_expression
+
         self.__form = Form()
 
     def dump(self):
@@ -47,6 +49,7 @@ class Layer(object):
         definition['uri'] = self.uri
         definition['isdomain'] = self.is_domain
         definition['isnmrel'] = self.is_nmrel
+        definition['displayexpression'] = self.display_expression
         definition['form'] = self.__form.dump()
         return definition
 
@@ -55,6 +58,7 @@ class Layer(object):
         self.uri = definition['uri']
         self.is_domain = definition['isdomain']
         self.is_nmrel = definition['isnmrel']
+        self.display_expression = definition['displayexpression']
         self.__form.load(definition['form'])
 
     def create(self):
@@ -63,7 +67,8 @@ class Layer(object):
             self.__layer = QgsVectorLayer(self.uri, layer_name, self.provider)
             if self.is_domain:
                 self.__layer.setReadOnly()
-
+            if self.display_expression:
+                self.__layer.setDisplayExpression(self.display_expression)
         for field in self.fields:
             field.create(self)
 
--- a/projectgenerator/libqgsprojectgen/dbconnector/db_connector.py
+++ b/projectgenerator/libqgsprojectgen/dbconnector/db_connector.py
@@ -56,8 +56,22 @@ class DBConnector:
                 srid
                 type  (Geometry Type)
                 kind_settings
+                ili_name
                 table_alias
                 model
+        '''
+        return []
+
+    def get_meta_attr(self, ili_name):
+        '''
+        Info about meta attributes of a given ili element
+
+        Return:
+            Iterable allowing to access rows, each row should allow to access
+            specific columns by name (e.g., a list of dicts {column_name:value})
+            Expected columns are:
+                attr_name
+                attr_value
         '''
         return []
 
--- a/projectgenerator/libqgsprojectgen/dbconnector/gpkg_connector.py
+++ b/projectgenerator/libqgsprojectgen/dbconnector/gpkg_connector.py
@@ -25,7 +25,6 @@ from ..generator.config import GPKG_FILTER_TABLES_MATCHING_PREFIX_SUFFIX
 
 GPKG_METADATA_TABLE = 'T_ILI2DB_TABLE_PROP'
 
-
 class GPKGConnector(DBConnector):
 
     def __init__(self, uri, schema):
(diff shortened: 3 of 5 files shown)


projectgenerator/libqgsprojectgen/dataobjects/layers.py
```diff
@@ -24,7 +24,7 @@
 
 class Layer(object):
 
-    def __init__(self, provider, uri, name, geometry_column=None, wkb_type=QgsWkbTypes.Unknown, alias='', is_domain=False, is_nmrel=False):
+    def __init__(self, provider, uri, name, geometry_column=None, wkb_type=QgsWkbTypes.Unknown, alias='', is_domain=False, is_nmrel=False, display_expression=''):
```
m-kuhn: In most cases `None` is preferable over `''`

projectgenerator/tests/testdata/ilimodels/ExceptionalLoadsRoute_V1.ili
```diff
@@ -10,7 +10,7 @@ VERSION "2017-02-08"  =
 
   TOPIC TypeOfRouteCatalogue
   EXTENDS CatalogueObjects_V1.Catalogues =
-
+    !!@ dispExpression="t_id||type"
```
m-kuhn: While technically possible, it's probably better to use something like `COALESCE(Type, ' (', RoadWidth, ' m)')` to promote good practice (Especially the `t_id` is meaningless except for the database backend).

signedav: Okay. Do we use this file for other things than tests?

m-kuhn: nope, not actually.

projectgenerator/gui/ili2db_options.py
```diff
@@ -33,24 +38,46 @@ def __init__(self, parent=None):
 
         self.buttonBox.accepted.disconnect()
         self.buttonBox.accepted.connect(self.accepted)
+        self.buttonBox.rejected.disconnect()
+        self.buttonBox.rejected.connect(self.rejected)
+        self.toml_file_browse_button.clicked.connect(
+            make_file_selector(self.toml_file_line_edit, title=self.tr('Open TOML File'),
+                               file_filter=self.tr('Toms Obvious Minimal Language File (*.toml)')))
```
m-kuhn: I think a user is more interested in the content of the file. Something like "Interlis Extra Meta Attributes" (other proposals also welcome).

projectgenerator/gui/ili2db_options.py
```diff
@@ -60,3 +87,5 @@ def restore_configuration(self):
             self.smart1_radio_button.setChecked(True)
         else:
             self.smart2_radio_button.setChecked(True)
+        self.toml_file_line_edit.setText(
+            settings.value('QgsProjectGenerator/ili2db/tomlfile'))
```
m-kuhn: Thinking about it, this probably shouldn't be saved in the app options. The toml file is model specific (while the inheritance type is more of a user preference).

 - Either not save it at all
 - or save it with the model name

Would also be good to show in the main interface that a toml file is used (with a read only label).

projectgenerator/gui/ili2db_options.py
signedav: I agree. But isn't it clear to everyone that it's an *Interlis* Extra Meta Attributes? 
What do you think about "Extra Attributes File (*.toml)"?

projectgenerator/gui/ili2db_options.py
signedav: Not convinced about the saving input. Doesn't the user expects that the configuration is saved when the filepath is entered and okay is pressed? Would't he be confused when it's gone when reopening the dialog? What do you mean with "save it with the model name"?

I do agree regarding the information label. I thought about something myself. But when we already need the space there, why still have the configuration in "Advanced Options"?

m-kuhn: > Not convinced about the saving input. Doesn't the user expects that the configuration is saved when the filepath is entered and okay is pressed? Would't he be confused when it's gone when reopening the dialog? What do you mean with "save it with the model name"?

Let's say this morning I import the *exceptional loads route* model and specify a toml file. When a couple of hours later this afternoon I import *kataster belasteter standorte*, the toml file from this morning is meaningless and shouldn't be there by default.
Instead it could be saved in the key `QgsProjectGenerator/ili2db/tomlfile/ExceptionalLoadsRoute_LV95_V1` and only be loaded when I import the schema for the same model again.

> I do agree regarding the information label. I thought about something myself. But when we already need the space there, why still have the configuration in "Advanced Options"?

I think from a UX perspective, line edits (and other editable things) receive more attention, so they should not be placed prominently for things which are only used 1 out of 30 times.

signedav: IMHO when we import later on something else, it makes neither sense to keep the model etc. As long it does keep the other information I would expect that my toml-configuration is still saved as well. 
Another option would be to clean toml-path after changing the Interlis file or the model. What do you think?

Anyway because you know the concept and the use cases much better than me, I am open to do it your way.

> I think from a UX perspective, line edits (and other editable things) receive more attention, so they should not be placed prominently for things which are only used 1 out of 30 times.

Agree.

m-kuhn: >  Another option would be to clean toml-path after changing the Interlis file or the model. What do you think?

That's pretty similar to what I had in mind. I'd propose to load the last toml used with the new model instead of cleaning.

Example workflow:

1. Load model ExceptionalLoadsRoute
2. Set toml file elr.toml
3. Generate
4. Open dialog again
  - model: ExceptionalLoadsRoute / toml: elr.toml
5. Change model to KbS
  - toml is cleared
6. set toml file to kbs.toml
7. Generate
8. Open again
  - model: KbS / toml: kbs.toml
9. Change model to ExceptionalLoadsRoute
  - toml: elr.toml

signedav: Okay I will do that. 
Since I will intergrate the information-label on the main dialog as well, this solution would work for me. So the user would notice that it selected something for him (because saved the last time...).

projectgenerator/gui/generate_project.py
```diff
@@ -538,6 +544,12 @@ def fill_models_line_edit(self):
         self.ili_models_line_edit.setText(
             self.multiple_models_dialog.get_models_string())
 
+    def fill_toml_file_info_label(self):
+        text = None
+        if self.ili2db_options.toml_file():
+            text = 'Ext. Attributes File: '+('...'+self.ili2db_options.toml_file()[len(self.ili2db_options.toml_file())-60:]) if len(self.ili2db_options.toml_file()) > 60 else 'Ext. Attributes File: '+self.ili2db_options.toml_file()
```
m-kuhn: text = `self.tr('Ext. Attributes File: {}').format(self.ili2db_options.toml_file())` or similar... And for the ellipsis use `…` ;)

m-kuhn:
@signedav is this ready for the final review?

signedav:
Not yet. I have it ready but not sure about the naming. Should I take **Extra Meta Attributes File** or do we think if it will need a more explicit name because the content of the file is not only concerning attributes?

Issue #184, opened 2018-04-27
Allow specifying an external meta attributes file (--iliMetaAttrs)


m-kuhn:
When importing a new model/schema. With `--iliMetaAttrs`

See https://github.com/claeis/ili2db/pull/149

signedav:
It's about that if we have the meta attribute for the display expression mentioned in #135 in a toml file, this will be imported as well. But what we need is the possibility to select the toml file (define the path) in the project generator (in the Advanced Settings) and load it from there.

In future it would be nice when the project generator would take the toml file from a repository (interlis2) if it recognizes a Model like `ExceptionalLoadsRoute_LV95_V1`.

Together with this implementations (#135 and this one) we need to habe test with configured toml- and with configured interlis-file. 

And I have to update the documentation in `projectgenerator/docs` including a toml-sample.

m-kuhn:
@romefi @pramisters @zigertiger (and whoever interested)

We are thinking about a name for the toml file to use consistently throughout application, documentation, repository, ... (thanks @signedav )

 * Extra Attributes File
 * Meta Attributes File
 * Extra Meta Attributes File
 * Interlis Plus File
 * Funky Toms Obvious Minimal Language File

Any favorites?

pramisters:
Extra Meta Attributes File

romefi:
Even if "Funky Toms Obvious Minimal Language File" is my personal favorite, I'd think "Extra Meta Attributes File" would do it. How about something more generic like "Extra Model Information File". It does not only concern the attributes (on long-term)?

m-kuhn:
I think the idea was that it's about "meta attributes" for entities (not about "meta" for "attributes") so the term is accurate. But possibly misleading because "attribute" may often be mistaken for "class attributes".

"Extra Model Information File" sounds better in this respect.

signedav:
Okay. I'll take **Extra Model Information File**.

But next time it has to be something with "Funky" in its name ;-)

PR #190, opened 2018-05-07
Specify external meta attributes file in GUI

signedav:
Specify an external meta attributes file (.toml) in Advanced Options and provide it to ili2db by `--iliMetaAttrs`

projectgen

Issue #184

Missing:
- [ ] Test
- [ ] Docs
- [ ] loading toml-file from repository

 (Plus small fix, that on Advanced Options with "Cancel" the settings are not stored)

### DIFF
--- a/projectgenerator/gui/generate_project.py
+++ b/projectgenerator/gui/generate_project.py
@@ -348,6 +348,7 @@ class GenerateProjectDialog(QDialog, DIALOG_UI):
 
         configuration.epsg = self.epsg
         configuration.inheritance = self.ili2db_options.get_inheritance_type()
+        configuration.tomlfile = self.ili2db_options.get_toml_file()
 
         configuration.base_configuration = self.base_configuration
         if self.ili_file_line_edit.text().strip():
--- a/projectgenerator/gui/generate_project.py
+++ b/projectgenerator/gui/generate_project.py
@@ -347,8 +347,8 @@ class GenerateProjectDialog(QDialog, DIALOG_UI):
             configuration.dbfile = self.gpkg_file_line_edit.text().strip()
 
         configuration.epsg = self.epsg
-        configuration.inheritance = self.ili2db_options.get_inheritance_type()
-        configuration.tomlfile = self.ili2db_options.get_toml_file()
+        configuration.inheritance = self.ili2db_options.inheritance_type()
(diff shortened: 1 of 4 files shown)


projectgenerator/gui/generate_project.py
```diff
@@ -348,6 +348,7 @@ def updated_configuration(self):
 
         configuration.epsg = self.epsg
         configuration.inheritance = self.ili2db_options.get_inheritance_type()
+        configuration.tomlfile = self.ili2db_options.get_toml_file()
```
m-kuhn: Can you remove the `get_` prefix (we also never do that on qgis side). Also remove it on the inheritance type while on it.

projectgenerator/gui/ili2db_options.py
```diff
@@ -18,6 +18,14 @@
  ***************************************************************************/
 """
 
+from projectgenerator.utils.qt_utils import (
+    make_file_selector,
+    make_save_file_selector,
+    Validators,
+    FileValidator,
+    NonEmptyStringValidator,
+    OverrideCursor
```
m-kuhn: That's a lot of imports there, of which I think some are not used. (e.g. can't find `OverrideCursor`)
PyCharm will list them in gray

projectgenerator/libili2db/ili2dbconfig.py
```diff
@@ -161,6 +162,9 @@ def to_ili2db_args(self, hide_password=False):
         if self.ilimodels:
             args += ['--models', self.ilimodels]
 
+        if self.tomlfile:
+            args += ["--iliMetaAttrs", self.tomlfile]
```
m-kuhn: @marioba do you remember, is this parameter for all ili2db commands, or only schemaimport?

m-kuhn:
Thanks @signedav looks really good

m-kuhn:
sidenote: toml file from repository is not yet on the todo list, this can be skipped for now.

projectgenerator/gui/ili2db_options.py
signedav: Yes. I don't use `make_save_file_selector` and `OverrideCursor` - I forgot to remove them after copy this part...

signedav:
Okay, so I will make the tests and docs and then we communicate before I would start with repo-stuff. Thanks.

Issue #119, opened 2017-12-13
Attributes of type ENUMs:  Forms are not created correctly (no drop-down lists or similar)  because of missing relations or other reasons


pramisters:
If an attribute with ENUM type is defined in an abstract class, projectgenerator doesn't create a relation in QGIS although the ENUM table is marked as "ENUM" in the metadatatable t_ili2db_table_prop.

Therefore, the widget type in the form Definition is not defined as "Relation reference", but "Text Edit".

In the attached example the abstract class "Hecken_Feldgehoelze_Basis" (with several attributes of type "ENUM") is extended to the classes "Hecken_Feldgehoelze_Linie" and "Hecken_Feldgehoelze_Flaeche" where the QGIS relation definitions to the ENUM tables are missing.

[ZG_Hecken_und_Feldgehoelze_V1_0.txt](https://github.com/opengisch/projectgenerator/files/1556215/ZG_Hecken_und_Feldgehoelze_V1_0.txt)

gacarrillor:
Could you please confirm if this happens in Project Generator v2.0.0?

pramisters:
Yes, it works now for ENUM type attributes in ABSTRACT classes! Thanks!

But there is another case where the QGIS relations are missing (although the entries in the t_ili2db_table_prop are present): 
**If ENUMs are defined inline (directly within the CLASS Definition)** 

Try the attached ili-model "Abfallsammelstellen_ZEBA_LV95_V1". For the Inline-ENUMs
- Abfallsammelstellen.Eigentum
- Abfallsammelstellen.Grundeigentum
- Abfallsammelstellen.OeffentlicheSammelstelle

the QGIS relations are missing, although the corresponding entries in t_ili2db_table_prop are present.

[ZG_Abfallsammelstellen_ZEBA_V1.txt](https://github.com/opengisch/projectgenerator/files/1568055/ZG_Abfallsammelstellen_ZEBA_V1.txt)

m-kuhn:
Is this a duplicate of 
https://github.com/opengisch/projectgenerator/issues/96 ?

pramisters:
Kind of, yes.
#96 was related to the missing foreign key. Now the implementation is different (work around with t_ili2db_table_prop). From my point of view, #96 can be closed as it has been replaced by this issue.

pramisters:
The issue persists:
[...]

PR #191, opened 2018-05-07
Domain-class relation improvements

gacarrillor:
Solves:

1. Issues with `Hazard_Mapping_V1_2.txt` (now Project Generator creates 41 domain-class relations instead of 0).
2. Issues with `ZG_Hazard_Mapping_V1_2.txt` (now Project Generator creates 41 domain-class relations instead of 0).
3. Nothing to do here.
4. Issues with `ZG_Abfallsammelstellen_ZEBA_V1.txt` (now Project Generator creates 8 domain-class relations, 3 of them are from inline ENUMs and 2 of them from the INTERLIS base model).

New test for `Hazard_Mapping_V1_2.ili`
Additions in existing tests to test inline ENUMs and domains form INTERLIS base model.

References #119


gacarrillor:
Since this is working well (we have several tests for it and keep improving it when something is missing), an upstream fix is not a priority for us here.

m-kuhn:
I'm afraid writing a proper parser is not easy, especially based on regex.
Not that I'm very afraid this is super fragile, but it would be very easy to exploit

E.g. (and there are other ways as well ... )

```
/* comment */ /*
broken
*/
```

Do you know the status of the upstream fix? It would be much cleaner

gacarrillor:
It is expected for ili2pg v4.

m-kuhn:
I was a bit surprised because I understood from @pramister s results that all the information is already in the meta tables.
Please continue, it's much appreciated, but at the same time I think we should revisit the clean fix to get rid of our internal parser mid-term.

gacarrillor:
Yeap, that's what I'm telling you above, we have a couple of requirements for Claude that are much more urgent for us, like proper GPKG support for our model.

pramisters:
For us the proper handling of ENUMs are 1st priority. As long as I have to configure all the ENUMs in the forms after having run the project generator, it is not of real help at all.

pramisters:
@m-kuhn 
In the meta table (t_ili2db_table_prop) we currently have the information which table is an ENUM table. That's correct. It's enough to create the relations in QGIS but it's not enough for the forms. You need to know which attributes are actually of which ENUM type. This information has to be supplied by ili2db or , if not available we need a kind of parser as @gacarrillor explained.

m-kuhn:
Thanks for clarification.

@romefi should we schedule an agenda item for this in the next meeting?
@gacarrillor would it be easy to refactor the model parser code into its own file (modelparser.py or similar)?

romefi:
Actually https://github.com/claeis/ili2db/issues/19 should solve the problem, right? Claude is supposed to fix it, he offered it to Columbia, Kt. GL and us and we accepted. @zigertiger, maybe you know something about the time schedule?
Anyway, it seems a matter of time till ili2db can handle it. As it's also for us a very high priority, I'd appreciate a workaround with parsing the model till ili2db supplies the information.

gacarrillor:
This PR is complete. I've also moved the parser class to its own file.

What's still pending to solve? 
- [This report](https://github.com/opengisch/projectgenerator/issues/119#issuecomment-381644131) of a two-level inheritance issue.
- Bag of ENUMs, which already has its own issue. 

I'd rather deal with those pending issues in separate PRs.

m-kuhn:
Good job, thanks a lot !

PR #194, opened 2018-05-10
Make PostgreSQL estimated metadata parameter optional in Generator class (by default True)

gacarrillor:
Several tools from our LADM plugin simply don't work as expected because the number of features in the database doesn't match the number of features QGIS returns when `estimatedmetadata` is `True`.

This PR allows us to switch that parameter off, while avoiding to alter the current behaviour for the rest of Project Generator users.

I would say this is a critical issue for us.

**Example 1:**

 + A layer in the database contains several features but QGIS returns `0` when calling `layer.featureCount()`.
 + A tool initially checks whether the layer has features or not. If no features are found, the tool does not continue.
 + Since QGIS returns `0` for `layer.featureCount()`, the tool does nothing beyond the first check, generating confusion for the user, who may be even seeing spatial data on QGIS canvas.

**Example 2:**

 + There is no data stored in the layer, but QGIS says (because of the estimated metadata parameter) the layer does have data! 

estimatedmetadata_1

 + When a tool does its check, the layer passes it and the tool continues (QGIS says there is data), but when the tool does its real job, for example, creating a Spatial Index, QGIS throws an exception (`QSpatialIndex` requires a layer with data).

estimatedmetadata_2

### DIFF
--- a/projectgenerator/libqgsprojectgen/generator/generator.py
+++ b/projectgenerator/libqgsprojectgen/generator/generator.py
@@ -33,11 +33,12 @@ from .config import IGNORED_SCHEMAS, IGNORED_TABLES, IGNORED_FIELDNAMES, READONL
 class Generator:
     """Builds Project Generator objects from data extracted from databases."""
 
-    def __init__(self, tool_name, uri, inheritance, schema=None):
+    def __init__(self, tool_name, uri, inheritance, schema=None, pg_estimated_metadata=True):
         self.tool_name = tool_name
         self.uri = uri
         self.inheritance = inheritance
         self.schema = schema or None
+        self.pg_estimated_metadata = 'true' if pg_estimated_metadata else 'false'
         if self.tool_name == 'ili2pg':
             self._db_connector = pg_connector.PGConnector(uri, schema)
         elif self.tool_name == 'ili2gpkg':
@@ -65,9 +66,10 @@ class Generator:
             if self.tool_name == 'ili2pg':
                 provider = 'postgres'
                 if record['geometry_column']:
-                    data_source_uri = '{uri} key={primary_key} estimatedmetadata=true srid={srid} type={type} table="{schema}"."{table}" ({geometry_column})'.format(
+                    data_source_uri = '{uri} key={primary_key} estimatedmetadata={estimated_metadata} srid={srid} type={type} table="{schema}"."{table}" ({geometry_column})'.format(
                         uri=self.uri,
                         primary_key=record['primary_key'],
+                        estimated_metadata=self.pg_estimated_metadata,
                         srid=record['srid'],
                         type=record['type'],
                         schema=record['schemaname'],


m-kuhn:
I think we should try to explore other possibilities before going this route, the bigger your tables will become (I'm not sure but I assume you'll have some users with larger tables?) the slower everything becomes.

 * Did you try to run `ANALYZE` and/or work with [`autovacuum`](https://www.postgresql.org/docs/9.5/static/routine-vacuuming.html), we might want to run this on QGIS, plugin or database server side once in a while
 * Alternatively we could add a `QgsVectorLayer::isEmpty()` method which could run much more performant than a full count if it's just to know if the layer is empty

gacarrillor:
The current situation is certainly undesired. While we come to the best solution (yeap, I agree it should be done in QGIS), we need to make our plugin work as expected.

If things become slow (only for us, this PR won't impact current Project Generator users) it would still be better than the current situation, when things simply don't work.

Do you think we can, at least temporarily, merge this PR and then come to the best solution for this use case? Additionally, we could revert this single commit when we have the best solution we can come up with.

m-kuhn:
Yes, we can do that.

Are you ok with working on a better solution in parallel after merging this?
I'm convinced we can find a cleaner solution than the two options we have now!

gacarrillor:
If you mean work together on that, yeap!

m-kuhn:
> If you mean work together on that, yeap!

Yes, that's what I meant. Do you have thoughts on [my comments above](https://github.com/opengisch/projectgenerator/pull/194#issuecomment-388112269)?

gacarrillor:
Yeah, that's actually the way we solve the problem, running VACUUM on the database table. Since autovacuum seems to be periodic, it wouldn't necessarily update the statistics before a user runs tools on layers. So, I guess the problem would remain.

I think the `QgsVectorLayer::isEmpty()`  makes a lot of sense if it can ask directly to the actual database content, instead of relying on cached statistics.

m-kuhn:
> Yeah, that's actually the way we solve the problem, running VACUUM on the database table. 

Is there a problem with this approach? It sounds not too bad to me from a conceptual perspective. The main question is, at what time you are running `VACUUM`?

> I think the QgsVectorLayer::isEmpty() makes a lot of sense

I'll work on that in this case :+1:

gacarrillor:
We run vacuum when we are in trouble with QGIS feature count. I think users shouldn't have to be worried about that. It should just work for them with no additional actions in the database.

> I'll work on that in this case :+1: 

Great!

m-kuhn:
> We run vacuum when we are in trouble with QGIS feature count.

Can you define that?

> I think users shouldn't have to be worried about that. It should just work for them with no additional actions in the database.

Users not, but potentially a DB admin ;)

gacarrillor:
When we encounter an unexpected behaviour on our tools (e.g., a topology check that is not made because QGIS says the layer has no features), we check layer's feature count in the layer tree. If we find that it is in fact a wrong feature count, we go to pgAdmin3 and run Vacuum Analyze.

m-kuhn:
>  If we find that it is in fact a wrong feature count, we go to pgAdmin3 and run Vacuum Analyze.

Oh, manually. I thought that was automated somehow. Well, yes, manually is not acceptable actually ;)

gacarrillor:
Yes, that's what I said that anything was better than what we had to do until v3.2.0!

m-kuhn:
Here's an incomplete but testable pull request in case you have time:

https://github.com/qgis/QGIS/pull/7016

PR #195, opened 2018-05-15
Run tests on specified QGIS version

m-kuhn:
Previously all tests would run on master and none on release-3_0.


m-kuhn:
Currently master builds are broken and introduce the travis failures. At least release-3_0 is now testing against the proper build.

PR #198, opened 2018-05-17
Class-domain relations: Support abstract classes that extend from other abstract clases

gacarrillor:
References #119 . Namely, see https://github.com/opengisch/projectgenerator/issues/119#issuecomment-381644131

Pending: Add test


m-kuhn:
Nice.
Do you have a test coming for this as well?

gacarrillor:
Now it's done.

PR #203, opened 2018-05-24
Skip  t_ili2db_meta_attrs if it does not exist 

m-kuhn:


projectgenerator/libqgsprojectgen/dbconnector/pg_connector.py
```diff
@@ -142,6 +157,9 @@ def get_tables_info(self):
         return []
 
     def get_meta_attrs(self, ili_name):
+        if not self._table_exists('t_ili2db_meta_attrs'):
```
signedav: You could use `PG_METAATTRS_TABLE` here.

projectgenerator/tests/test_dump.py
```diff
@@ -0,0 +1,50 @@
...
+
+start_app()
+
+
+class TestCustomDump(unittest.TestCase):
+
+    def test_ili2pg_dump_without_metattr(self):
+        myenv = os.environ.copy()
+        myenv['PGPASSWORD'] = 'docker'
+        call(["pg_restore", "-Fc", "-hpostgres", "-Udocker", "-dgis", testdata_path("dumps/_nupla_dump")], env=myenv)
```
signedav: I always make spaces between parameters, like `-h postgres` but I guess this works as well.

m-kuhn: With `pg_restore` also? I normally also make those spaces, but had trouble here.

signedav: Also on postgresql.org/docs there is the example for `pg_restore` with spaces. I don't know where the problems come from, but I suppose when it works here without them, then it's fine as well.

m-kuhn: I also had to use an environment variable for the password, might be related to an old version of pg_restore. I'll keep this working approach and merge.

Thanks for the review

PR #216, opened 2018-06-22
Default values for geopackage

m-kuhn:
Sqlite does not allow the definition of function based default values, so there is no `nextval` or similar that could be used to predict upcoming serial values on form opening.

Client side default values to the rescue, we manage things via the same `T_KEY_OBJECT` table as ili2gpkg does.

Two issues remain for the moment:

 * The expression function is defined on first form opening, so the very first time, it's not yet there. Project macros or requiring clients to install a plugin are a way around that but they both come at an expense. Macros give warnings, having to install a plugin makes the project less portable.
Or we try to use the form API to inject the value into a hidden widget.

 * There seems to be a timeout (due to locking?) on the database when acquiring more serial values in subsequent digitizing within a single transaction. Needs some investigation.

### DIFF
--- a/projectgenerator/libqgsprojectgen/dataobjects/fields.py
+++ b/projectgenerator/libqgsprojectgen/dataobjects/fields.py
@@ -18,17 +18,20 @@
  ***************************************************************************/
 """
 
-from qgis.core import QgsEditorWidgetSetup
+from qgis.core import (
+    QgsEditorWidgetSetup,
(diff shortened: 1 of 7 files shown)


m-kuhn:
Fixes https://github.com/opengisch/projectgenerator/issues/120

m-kuhn:
Mini-plugin for providing the serial logic

PR #219, opened 2018-07-06
[WIP] Hide field if meta-attribute 'hidden' is present

gacarrillor:
Example in TOML:

```
["curso_qfield.catastro_registro.Punto.Nombre_Punto"]
hidden="True"
```

### DIFF
--- a/projectgenerator/libqgsprojectgen/dbconnector/pg_connector.py
+++ b/projectgenerator/libqgsprojectgen/dbconnector/pg_connector.py
@@ -213,11 +213,13 @@ class PGConnector(DBConnector):
             unit_join = ''
             text_kind_join = ''
             disp_name_join = ''
+            full_name_join = ''
 
             if self.metadata_exists():
                 unit_field = "unit.setting AS unit,"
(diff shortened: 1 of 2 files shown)


gacarrillor:
I have some troubles with the new test. Apparently, the newly added form layout has duplicated fields in its tab 'General'. 

This is what I get in QGIS ('tipo' shouldn't be there as it is specified in the TOML file used in the test):

duplicated_fields


And these are the attributes from such tab printed:

`['numero_predial', 'fmi', 'avaluo1', 'nombre', 'nombre', 'tipo', 'tipo', 'u_espacio_de_nombres', 'u_local_id', 'comienzo_vida_util_version', 'comienzo_vida_util_version', 'comienzo_vida_util_version', 'fin_vida_util_version', 'fin_vida_util_version', 'fin_vida_util_version', 'geometria']
`

Do you have any clue on what might be happening with those duplicated fields in the form layout?

I used [this model](https://github.com/opengisch/projectgenerator/blob/master/projectgenerator/tests/testdata/ilimodels/CIAF_LADM/CIAF_LADM.ili) with smart2Inheritance.

PR #226, opened 2018-07-18
Allow ignoring groups (nodes) when ordering new layers 

gacarrillor:
This PR makes the ordering of new layers more flexible, e.g., ignoring certain groups that should be always on top, such as a group of topology errors. 

The introduced parameter is optional, so current users won't see any change in behaviour and also there won't be anything to adjust in other parts of Project Generator.

From the 'Asistente LADM_COL' plugin we pass the group name we would like to ignore when creating the legend and that's all we need.

### DIFF
--- a/projectgenerator/utils/qgis_utils.py
+++ b/projectgenerator/utils/qgis_utils.py
@@ -27,7 +27,7 @@ layer_order = ['point',  #QgsWkbTypes.PointGeometry
                'unknown']  # Anything else like geometry collections or plugin layers
 
 
-def get_first_index_for_layer_type(layer_type, group):
+def get_first_index_for_layer_type(layer_type, group, ignore_node_names=[]):
     """
     Finds the first index (from top to bottom) in the layer tree where a
     specific layer type is found. This function works only for the given group.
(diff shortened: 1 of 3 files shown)


m-kuhn:
Thanks for the pull request, travis is passing now, I'll have a look at the code soon, but the description sounds good.

> From the 'Asistente LADM_COL' plugin we pass the group name we would like to ignore when creating the legend and that's all we need.

Is there a deeper meaning in "that's all we need."?

gacarrillor:
> Is there a deeper meaning in "that's all we need."?

The only change for somebody wanting to use this PR is to pass a parameter (`ignore_node_names`) to a `generator.legend()` call:

```
# Where 'Validation errors' is the name of the group that should be kept at the top
generator.legend(layers, ignore_node_names=["Validation errors"])
```

projectgenerator/utils/qgis_utils.py
```diff
@@ -48,7 +52,7 @@ def get_first_index_for_layer_type(layer_type, group):
     return None
 
 
-def get_suggested_index_for_layer(layer, group):
+def get_suggested_index_for_layer(layer, group, ignore_node_names=[]):
```
m-kuhn: Better avoid mutable defaults (i.e. use `ignore_node_names=None`)
See https://docs.python-guide.org/writing/gotchas/#what-you-should-do-instead

gacarrillor: Thanks for the suggestion, should be fixed now.

PR #228, opened 2018-07-25
Fix inheritance config

m-kuhn:


m-kuhn:
@gacarrillor is it ok for you to remove `ilidataimporter.DataImporter`? It was a 95% copy of `Importer` which I am afraid is getting out of sync more and more.

gacarrillor:
It seems like a good move and skimming the code changes I think it's safe. 

Something to keep in mind is that [ili2db will make it optional](https://github.com/claeis/ili2db/issues/112) to run an `import` from a `data import` command (something that we do in Project Generator).

m-kuhn:
Thanks, I will merge it then. Let's keep it in mind and handle this situation when we upgrade to an ili2db version with this behavior implemented.

PR #230, opened 2018-07-31
Avoid showing General tab as default in settings

gacarrillor:
That always means an extra click for users, since we have no options in General.


m-kuhn:
Hehehe, don't you like it ;)

gacarrillor:
Not always :( 
:+1:

PR #232, opened 2018-08-09
Improve usability while adding paths in custom model dir dialog

wfpinedar:
A minor modification is made to improve the usability in the add custom model dir functionality, so that in a single click to button "Add" we create a new list element and open the file explorer dialog to immediately set a path for it. 

This saves a couple of clicks for users.


m-kuhn:
Nice one!

PR #238, opened 2018-09-14
Improvements for BAG OF ENUM

gacarrillor:
+ [x] Configure BAG OF ENUM for tables with more than one geometry column.
 + [x] Use the meta-attribute `mapping=ARRAY` from meta-attrs table.
 + [x] Add more checks for bags of enum configuration for the case where a list of layers is given. 
 + [x] Adjust the exposed `create_project` method to work with bags of enums.

### DIFF
--- a/projectgenerator/qgs_project_generator.py
+++ b/projectgenerator/qgs_project_generator.py
@@ -165,6 +165,21 @@ class QgsProjectGeneratorPlugin(QObject):
         return Generator
 
     def create_project(self, layers, relations, bags_of_enum, legend, auto_transaction=True, evaluate_default_values=True):
+        """
+        Expose the main functionality from Project Generator to other plugins,
+        namely, create a QGIS project from objects obtained from the Generator
+        class.
+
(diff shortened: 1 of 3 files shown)


projectgenerator/qgs_project_generator.py
```diff
@@ -164,10 +164,11 @@ def show_about_dialog(self):
     def get_generator(self):
         return Generator
 
-    def create_project(self, layers, relations, legend, auto_transaction=True, evaluate_default_values=True):
+    def create_project(self, layers, relations, bags_of_enum, legend, auto_transaction=True, evaluate_default_values=True):
```
m-kuhn: Can you add some docstrings here?

m-kuhn:
I don't know about the whole parser code, but backing that up with a good set of unit tests as done here looks good.

projectgenerator/qgs_project_generator.py
gacarrillor: Done!

PR #240, opened 2018-09-17
Make create import tid optional

seralra96:
This allows users to choose between create import tid column (`t_ili_tid`) on the data base (`True` by default):

imagen


m-kuhn:
Why does this need to be optional? (No opinion yet, but a new option needs a good reason ;) )

PR #241, opened 2018-09-17
Set map canvas extent to one obtained from the model (t_ili2db_column_prop table)

### LINKED ISSUE #83: Map extent from INTERLIS models
Since INTERLIS models have coordinate ranges that play a role when validating data (i.e., avoiding data export when features are outside of it), couldn't we read the allowed extent and do things like the following?

 + Set initial map extent 
 + Create a spatial bookmark called "Allowed extent" (and open the Spatial Boormarks Panel each time a project is generated)
 + Show a warning message when saving features out of the allowed extent

gacarrillor:
We pick the first layer extent for setting the canvas extent. We could have used the union of all extents in the `t_ili2db_column_prop` table, but we didn't since the base geometry model is almost set in stone and all layers use the same definitions from such model. 

@m-kuhn What do you think about such approach? Do Swiss models use several base geometry models or extent definitions? 

2 tests added.

Fix #83

### DIFF
--- a/projectgenerator/libqgsprojectgen/dbconnector/db_connector.py
+++ b/projectgenerator/libqgsprojectgen/dbconnector/db_connector.py
@@ -58,7 +58,7 @@ class DBConnector:
                 type  (Geometry Type)
                 kind_settings
                 ili_name
-                extent (insert coordinates xmin, ymin, xmax, ymax)
+                extent [a string: "xmin;ymin;xmax;ymax"]
                 table_alias
                 model
         '''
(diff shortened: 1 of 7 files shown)


projectgenerator/libqgsprojectgen/dbconnector/db_connector.py
```diff
@@ -58,6 +58,7 @@ def get_tables_info(self):
                 type  (Geometry Type)
                 kind_settings
                 ili_name
+                extent (insert coordinates xmin, ymin, xmax, ymax)
```
m-kuhn: needs to be semicolon, not comma separated, reading the code?

m-kuhn:
I like it :+1:

projectgenerator/libqgsprojectgen/dbconnector/db_connector.py
gacarrillor: Right, done.

gacarrillor:
Just added two tests for this. It should be ready now.

Issue #151, opened 2018-03-09
multiple model versions


zigertiger:
SITUATION:
There can be multiple versions of a data model with the same model name. Obsolete model(file)s are stored in the "obsolete/" directory of models.geo.admin.ch but remain listed in ilimodels.xml. Each model has a name and a version date that is stored in ilimodels.xml

Example: 
[1 - obsolete model version]
   name: Laermempfindlichkeitsstufen_LV95_V1_1
   file: obsolete/Laermempfindlichkeitsstufen_LV03_V1_1_o0.ili
   version: 2017-02-16

[2 - actual model version]
   name: Laermempfindlichkeitsstufen_LV95_V1_1
   file: BAFU/Laermempfindlichkeitsstufen_V1_1.ili
   version: 2018-02-16
   precursorVersion: 2017-02-16

PROBLEM:
The projectgenerator apparently handles only model names but not versions. When I choose the given model to be implemented, I get two models listed (attachment).

SOLUTION:
The projectgenerator shall evaluate model names and versions as well. Possible extension: the projectgenerator filters obsolete model version.
At least, model name AND version shall be listed in the models field of the gui dialog.

![models](https://user-images.githubusercontent.com/25387992/37196866-bf5ef2aa-2378-11e8-89eb-5a65a64e6076.png)

zigertiger:
I'll check that with Claude Eisenhut (what ili2pg is actually doing with precursor model versions).

m-kuhn:
@zigertiger do we have an update on this?

zigertiger:
just asked Claude via E-Mail

claeis:
Only one ili file is current. 

The ``precursorVersion`` element from ilimodels.xml entries must be used to build an ordered list of models with the same name to know which one is the current.

m-kuhn:
Does this mean only the current version can be loaded with ili2db and listing any other version in a gui is not meaningful?

claeis:
Yes, only the current version can be loaded. (Would it not be better, if the projectgenerator uses the existing code (in ili2c.jar) to read the (current) model list from the repositories?)

claeis:
``--help,...but the option --listModels that provides the functionality to the user  is not yet there (claeis/ili2c#15)

m-kuhn:
It would be nice from a perspective, that all code is in a single place.
[...]

PR #242, opened 2018-09-18
Do not load duplicate model entries in GUI

signedav:
IMHO this is a solid solution, but I'm not sure if it could be nicer with something like `if any(model['name'] in s.data(int(Qt.DisplayRole)) for s in self):`...

#151

### DIFF
--- a/projectgenerator/libili2db/ilicache.py
+++ b/projectgenerator/libili2db/ilicache.py
@@ -258,12 +258,6 @@ class IliModelItemModel(QStandardItemModel):
                 if any(model['name'] in s for s in names):
                     continue
 
-                # for modelname in self.data(int(Qt.DisplayRole)):
-                #     if modelname == model['name']:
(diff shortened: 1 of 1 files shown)


m-kuhn:
Yes, looks like a solid solution. Since (at least to me) it's not intuitively readable, a comment on what it does would be good

m-kuhn:
Hmm... I think the `fix #xyz` part was missing in the commit message. Can you close the issue?

signedav:
So it needs to be in the commit message, right? Sorry for that. Wouldn't happen again :slightly_smiling_face:

m-kuhn:
Maybe it also works if it's in the description, but it needs to have `fix` or another [supported keyword](https://help.github.com/articles/closing-issues-using-keywords/) on the same line

signedav:
I added the keyword with edit, later on. So probably it does not work like this. Thanks.

Issue #130, opened 2018-01-05
Icon in Database Toolbbar


m-kuhn:
Draft by @zigertiger

![projectgenerator-icon-databasetoolbar](https://user-images.githubusercontent.com/588407/34607339-65f72bec-f213-11e7-97c7-605c19c2b23b.png)

[projectgenerator-icon.zip](https://github.com/opengisch/projectgenerator/files/1606413/projectgenerator-icon.zip)

zigertiger:
icon created.
[projectgenerator-icon.zip](https://github.com/opengisch/projectgenerator/files/1619175/projectgenerator-icon.zip)

m-kuhn:
Icon is introduced for the plugin itself.

For the toolbar, should this be used for the schemaimport? Should we also have one for import and export?

m-kuhn:
@zigertiger can you have a creative session for import/export functionalities

signedav:
@zigertiger It looks like this. 
![icons_projgen](https://user-images.githubusercontent.com/28384354/45673401-dddbf300-bb2a-11e8-943e-d83058c03a2e.gif)

When your icons are ready and I will add them :slightly_smiling_face:

zigertiger:
comin' right up!
[projectgenerator-offerte-opengisch-20180713.pdf](https://github.com/opengisch/projectgenerator/files/2391988/projectgenerator-offerte-opengisch-20180713.pdf)

m-kuhn:
@zigertiger shall we put the offer into the plugin ;P ?

zigertiger:
haha. wrong pdf

signedav:
Pushed them to the #243 PR.
I like the icons, they are well-made. Still I wonder, if the "subicon" that makes the difference between import/export should be bigger or at least different colored because it's hard to see a difference on this size. 
![projectgenerator icons](https://user-images.githubusercontent.com/28384354/45678096-02899800-bb36-11e8-9436-5d24e325a94d.png)

But IMHO we could go on still with this version because you have the tooltip anyway...

m-kuhn:
@zigertiger the icons are not in the database toolbar. I hope you don't mind.

PR #243, opened 2018-09-18
Icons in Toolbar and Menu

signedav:
Added the icons in the toolbar and the menu. 
icons_projgen

The icons are still needed.
#130


m-kuhn:
Thanks @signedav 
Ping @zigertiger for icons

m-kuhn:
@signedav after thinking again, I would prefer not to use resources to avoid the step to create resources. The reason for this is, that it either needs to be done manually (which make development more complex) or it needs to be done on release which means the plugin can no longer just be checked out from the repository and be tested without running a make script.

projectgenerator/qgs_project_generator.py
```diff
@@ -97,6 +103,11 @@ def initGui(self):
         self.iface.addPluginToDatabaseMenu(
             self.tr('Project Generator'), self.__about_action)
 
+        self.toolbar = self.iface.addToolBar('Project Generator')
```
m-kuhn: `self.tr('Project Generator')`

I think we also need to more lines to properly remember the visible state automatically and show it in the toolbar context menu.

```
self.toolbar.setObjectName("ProjectGeneratorToolbar")
self.toolbar.setToolTip(self.tr('Project Generator Toolbar'))
```

m-kuhn:
:ship: Ship it :shipit:

m-kuhn:
I think new strings should appear on transifex in the next couple of minutes and can then be translated so we have them in the next release.

Issue #215, opened 2018-06-20
At export propose possible models


romefi:
If possible use the one used chosen at schema import as default

m-kuhn:
--exportModels

m-kuhn:
You can create a database with multiple models inside and only export part of this.
The available models are listed in `t_ili2db_models` ( modelname up to the first `{` )

signedav:
They often looks like this `CHAdminCodes_V1 AdministrativeUnits_V1{ CHAdminCodes_V1 InternationalCodes_V1 Dictionaries_V1 Localisation_V1 INTERLIS} AdministrativeUnitsCH_V1{ CHAdminCodes_V1 InternationalCodes_V1 LocalisationCH_V1 AdministrativeUnits_V1 INTERLIS}` so I assume not only up to the first `{` but instead:  CHAdminCodes_V1, AdministrativeUnits_V1 and AdministrativeUnitsCH_V1. Right?

m-kuhn:
Yes, you are right.
The ones inside the curly braces are dependencies, maybe there is a smart way to automatically enable them along when something is activated?

signedav:
After several tests I'm pretty sure that there is no difference if I choose one model or another or none. The `t_ili2db_models` table always contains all from the chosen file. But I'll check if they are stored elsewhere...

signedav:
Does by chance anyone have some appropriate test data for this use case? Eg. @zigertiger

zigertiger:
Important use case: polymorphic export.

Nutzungsplanung: cantonal model (GL_Nutzungsplanung_V1_3) as an extension of federal model (Nutzungsplanung_LV95_V1_1). Export either according cantonal or federal model.
As m-kuhn mentioned, ili2db provides the option `--exportmodels` for these cases.

[1632-nup-20180809.zip](https://github.com/opengisch/projectgenerator/files/2397400/1632-nup-20180809.zip)
(at the moment, you need to import the data using the option `--ver4-noSchemaImport`)

signedav:
Hi @zigertiger 
Since I'm not that experienced with these ili2pg command I have the question, how to import with the option `--ver4-noSchemaImport` - because it gives me back `--ver4-noSchemaImport: unknown option`.

Would your use case with thge cantonal and federal model lead to the command part like `--models Nutzungsplanung_LV95_V1_1 --exportModels GL_Nutzungsplanung_V1_3` 
Means you have to be able to choose models and exportModels? Or should it be the same?

[...]

PR #249, opened 2018-09-20
Export models from models stored in db

signedav:
The available models are read from the table `t_ili2db_model` and provided as a list to the GUI.
At the moment you can choose several models to export.

It looks like this:
export

It's not finished. Some parts are missing.

- [x] read from db
- [x] reload when parameter changed
- [x] timer on reload
- [x] checkboxes in ListView
- [x] avoid crash on selecting invalid GPKG file
- [ ] store the selection for each schema (because at the moment the selection is not stored)

It's about #215

### DIFF
--- a/projectgenerator/gui/export.py
+++ b/projectgenerator/gui/export.py
@@ -60,7 +60,7 @@ class ExportModels(QStringListModel):
                     for modelname in regex.split(db_model['modelname']):
                         if modelname:
                             modelnames.append(modelname.strip())
-        except:
+        except TypeError:
             pass
         self.setStringList(modelnames)
 
--- a/projectgenerator/gui/export.py
+++ b/projectgenerator/gui/export.py
@@ -23,14 +23,14 @@ import webbrowser
 import os.path
 import re
 
-from projectgenerator.gui.options import OptionsDialog, CompletionLineEdit
+from projectgenerator.gui.options import OptionsDialog, ModelListView
 from projectgenerator.gui.multiple_models import MultipleModelsDialog
 from projectgenerator.libili2db.iliexporter import JavaNotFoundError
 from projectgenerator.libili2db.ilicache import IliCache, ModelCompleterDelegate
 from projectgenerator.utils.qt_utils import make_save_file_selector, Validators, \
     make_file_selector, FileValidator, NonEmptyStringValidator, make_folder_selector, OverrideCursor
 from qgis.PyQt.QtGui import QColor, QDesktopServices, QFont, QValidator
-from qgis.PyQt.QtWidgets import QDialog, QDialogButtonBox, QApplication, QCompleter, QMessageBox, QListView
+from qgis.PyQt.QtWidgets import QDialog, QDialogButtonBox, QApplication, QCompleter, QMessageBox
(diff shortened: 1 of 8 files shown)


m-kuhn:
The CheckStateRole had to be initialized.
What's still left to do is to find out, how to react on clicks and SpaceBar, I hope that's not something to handle manually.

signedav:
Looks like this:
model_demo_export

signedav:
Now the checking/unchecking with pressing space is implemented as well.

I'm still confused about the `--exportmodels` part in your comments @zigertiger and @m-kuhn . Until now it worked with `--models`, why do we have to change that?

projectgenerator/gui/export.py
```diff
@@ -21,24 +21,75 @@
...
+                self._db_connector = pg_connector.PGConnector(uri, schema)
+
+            if self._db_connector.db_or_schema_exists():
+                db_models = self._db_connector.get_models()
+                for db_model in db_models:
+                    regex = re.compile(r'(?:\{[^\}]*\}|\s)')
+                    for modelname in regex.split(db_model['modelname']):
+                        if modelname:
+                            modelnames.append(modelname.strip())
+        except:
```
m-kuhn: Can you catch specific errors here? Catch-all exceptions should be avoided.

projectgenerator/gui/export.py
```diff
@@ -21,24 +21,89 @@
...
+
+            if self._db_connector.db_or_schema_exists():
+                db_models = self._db_connector.get_models()
+                for db_model in db_models:
+                    regex = re.compile(r'(?:\{[^\}]*\}|\s)')
+                    for modelname in regex.split(db_model['modelname']):
+                        if modelname and modelname not in blacklist:
+                            modelnames.append(modelname.strip())
+        except TypeError:
+            pass
```
m-kuhn: Can you add a comment why this error is expected and ignored

projectgenerator/gui/export.py
```diff
@@ -117,11 +169,53 @@ def __init__(self, base_config, parent=None):
...
+        self.gpkg_file_line_edit.textChanged.connect(self.request_for_refresh_models)
+
+        self.export_models_model = ExportModels(None, None, None)
+        self.refreshed_export_models_model()
+        self.export_models_view.setModel(self.export_models_model)
+        self.export_models_view.clicked.connect(self.export_models_model.check)
+        self.export_models_view.spaced.connect(self.export_models_model.check)
+
+    def request_for_refresh_models(self):
+        #hold refresh back one second
```
m-kuhn: It's probably worth removing the "one second" part of this comment ;)

projectgenerator/gui/options.py
```diff
@@ -129,4 +129,20 @@ def focusInEvent(self, e):
...
     def mouseReleaseEvent(self, e):
         super(CompletionLineEdit, self).mouseReleaseEvent(e)
-        self.punched.emit()
\ No newline at end of file
+        self.punched.emit()
+
+
+class ModelListView(QListView):
+
+    spaced = pyqtSignal(QModelIndex)
```
m-kuhn: Interesting word. Does it mean `space_pressed`?

signedav: Sorry, I guess my liguaphile alter-ego overcame me. I will change it for the good of understandable code..

PR #251, opened 2018-10-02
Avoid QGIS crash by calling processEvents() only when required

gacarrillor:
Also: Fix conditionals that weren't really working.

The crash happened on data Import, throwing this error: 

```
Stacktrace (piped through c++filt):
src/core/qgsmessagelog.cpp: 29: (logMessage) [1900ms] 2018-10-01T16:13:57 Qt[1] QSettings::value: Empty key passed
src/core/qgsmessagelog.cpp: 29: (logMessage) [120642ms] 2018-10-01T16:15:58 Python error[1] RecursionError: maximum recursion depth exceeded

Fatal Python error: Cannot recover from stack overflow.

Current thread 0x00007fa96cec9340 (most recent call first):
  File "/home/germap/.local/share/QGIS/QGIS3/profiles/default/python/plugins/projectgenerator/libili2db/iliimporter.py", line 133 in stderr_ready
  File "/home/germap/.local/share/QGIS/QGIS3/profiles/default/python/plugins/projectgenerator/gui/import_data.py", line 224 in on_stderr
```

### DIFF
--- a/projectgenerator/gui/import_data.py
+++ b/projectgenerator/gui/import_data.py
@@ -221,7 +221,6 @@ class ImportDataDialog(QDialog, DIALOG_UI):
         self.txtStdout.setTextColor(QColor('#2a2a2a'))
         self.txtStdout.append(text)
         self.advance_progress_bar_by_text(text)
-        QCoreApplication.processEvents()
 
     def on_process_started(self, command):
         self.disable()
@@ -392,7 +391,9 @@ class ImportDataDialog(QDialog, DIALOG_UI):
                 "https://opengisch.github.io/projectgenerator/docs/user-guide.html#import-an-interlis-transfer-file-xtf")
 
     def advance_progress_bar_by_text(self, text):
-        if text.strip() == 'Info: compile models…':
+        if text.strip() == 'Info: compile models...':
             self.progress_bar.setValue(50)
-        elif text.strip() == 'Info: create table structure…':
+            QCoreApplication.processEvents()
+        elif text.strip() == 'Info: create table structure...':
             self.progress_bar.setValue(75)
+            QCoreApplication.processEvents()


projectgenerator/gui/import_data.py
```diff
@@ -392,7 +391,9 @@ def help_requested(self):
                 "https://opengisch.github.io/projectgenerator/docs/user-guide.html#import-an-interlis-transfer-file-xtf")
 
     def advance_progress_bar_by_text(self, text):
-        if text.strip() == 'Info: compile models…':
```
m-kuhn: Is there something wrong with using the ellipsis?

projectgenerator/gui/import_data.py
```diff
@@ -392,7 +391,9 @@ def help_requested(self):
...
+        if text.strip() == 'Info: compile models...':
             self.progress_bar.setValue(50)
-        elif text.strip() == 'Info: create table structure…':
+            QCoreApplication.processEvents()
+        elif text.strip() == 'Info: create table structure...':
             self.progress_bar.setValue(75)
+            QCoreApplication.processEvents()
```
m-kuhn: Does that really fix the crash? Or just reduce the risk?
If it helps something, I am ok to merge this. But I think we should rather send this to a background thread and do this properly (sooner or later).

gacarrillor: I think it is fair to say that it fixes the crash because instead of calling `QCoreApplication.processEvents()` hundreds (potentially thousands) of times, we now just call it twice at the most.

projectgenerator/gui/import_data.py
m-kuhn: Of course there is...

projectgenerator/gui/import_data.py
m-kuhn: No criticism, it's a good move in the right direction, but it's still more a band aid than a real fix :)

projectgenerator/gui/import_data.py
gacarrillor: It wasn't fulfilling the condition, that is, the progress bar was never showing the 50% and 75% indicators. Could you please confirm if the ellipsis actually works for you?

projectgenerator/gui/import_data.py
m-kuhn: ... as we actually loose real time progress reporting

gacarrillor: Could you please tell me why? I still don't get it.

gacarrillor: Oh, the message log is being updated in a normal way for the user, it doesn't hang or something.

m-kuhn: Two things:

 * we still have `processEvents()` calls in there, so we can still have crashes
 * we only update the log upon certain events instead of whenever we get new text to show, which can make user wonder if the process is still alive

m-kuhn: Hmm... might be that Qt fixed a couple of things. I think mac is still affected by the no update. And there's still a risk of crashes, it's just decreased a bit (not sure if it decreases by the number of calls to `processEvents()` or by the amount of time between the last "real" event loop and the last `processEvents()` call though)

PR #253, opened 2018-10-09
DB superuser options in settings

signedav:
There are new configuration options in the "Settings"
settings
You can enter there superuser login information for the DB.

On "Generate Project" in the PostgreSQL section you can enter the login information you want to use for the project itself. In case you want to use the superuser login from the "Settings" to make the ili2pg command (means create Schema and everything) you can select the checkbox below.
generatoproject

### DIFF
--- a/projectgenerator/gui/generate_project.py
+++ b/projectgenerator/gui/generate_project.py
@@ -147,6 +147,8 @@ class GenerateProjectDialog(QDialog, DIALOG_UI):
         self.pg_user_line_edit.textChanged.connect(
             self.validators.validate_line_edits)
         self.pg_user_line_edit.textChanged.emit(self.pg_user_line_edit.text())
+        self.pg_use_super_login.setText(
+            "Use logins from settings (User: {}) for schema generation".format(base_config.super_pg_user))
         self.ili_models_line_edit.textChanged.connect(
             self.validators.validate_line_edits)
         self.ili_models_line_edit.textChanged.emit(
@@ -242,7 +244,7 @@ class GenerateProjectDialog(QDialog, DIALOG_UI):
                     return
 
             try:
-                generator = Generator(configuration.tool_name, configuration.uri(False), configuration.uri(True),
+                generator = Generator(configuration.tool_name, configuration.uri(False),
                                       configuration.inheritance, configuration.dbschema)
                 self.progress_bar.setValue(50)
             except OperationalError:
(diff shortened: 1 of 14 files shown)


projectgenerator/gui/generate_project.py
```diff
@@ -147,6 +148,8 @@ def __init__(self, iface, base_config, parent=None):
         self.pg_user_line_edit.textChanged.connect(
             self.validators.validate_line_edits)
         self.pg_user_line_edit.textChanged.emit(self.pg_user_line_edit.text())
+        self.pg_use_super_login.setText(
+            "Generate schema with superuser login from settings ({})".format(base_config.super_pg_user))
```
m-kuhn: translation

projectgenerator/gui/import_data.py
```diff
@@ -123,6 +124,8 @@ def __init__(self, base_config, parent=None):
         self.pg_user_line_edit.textChanged.connect(
             self.validators.validate_line_edits)
         self.pg_user_line_edit.textChanged.emit(self.pg_user_line_edit.text())
+        self.pg_use_super_login.setText(
+            "Generate schema with superuser login from settings ({})".format(base_config.super_pg_user))
```
m-kuhn: translation

projectgenerator/libili2db/ili2dbconfig.py
```diff
@@ -107,6 +116,7 @@ def __init__(self):
         self.dbhost = ''
         self.dbpwd = ''
         self.dbusr = ''
+        self.dbusesuperlogin = False
```
m-kuhn: can you use some `_` for readability in this variable name? The others are the way they are to be consistent with ili2db options (and this one is not an ili2db option if I am not wrong)

projectgenerator/libili2db/ili2dbconfig.py
```diff
@@ -115,14 +125,18 @@ def __init__(self):
         self.ilimodels = ''
         self.tomlfile = ''
 
-    @property
-    def uri(self):
+    def uri(self, su = False):
```
m-kuhn: API docs for `su` are missing.
Another approach would be a property `super_user_uri`.

signedav: Just for information or should a second property be preferred?

I decided for method because of avoiding to have double written code here. But I don't really mind.

m-kuhn: Another property would be nice to have cleaner code. To avoid duplicate code, this method can be called internally.

signedav: You mean two properties calling the method? 
Okay, I can do that...

m-kuhn: Yes, and the method prefixed with a `_` to mark it as private and a short API doc on top to explain the parameter.

signedav:
Thanks @m-kuhn :taco:

PR #257, opened 2018-10-17
Pass filename via parameter

### LINKED ISSUE #256: Error opening Generate dialog using Project Generator v3.3.2
There is an error on both Windows 10 and GNU/Linux when opening the Generate dialog using current (from today 2018.10.16) QGIS master. It seems to be while attempting to download models from the repositories.

Project Generator version: 3.3.2

error_project_generator

m-kuhn:
Fix #256

### DIFF
--- a/projectgenerator/utils/qt_utils.py
+++ b/projectgenerator/utils/qt_utils.py
@@ -20,6 +20,7 @@
 
 import os.path
 import inspect
+import functools
 from qgis.PyQt.QtWidgets import (
     QFileDialog,
     QApplication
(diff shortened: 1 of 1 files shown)


m-kuhn:
@gacarrillor it's a random attempt to solve issue #256 , do you have a chance to test this updated file on an affected machine?

I cannot reproduce the issue locally. One thing I could imagine is the recent update of Python on Windows to version 3.7 (from 3.6 which I also have locally to test), but I didn't find anything in the release notes that could explain this problem.

gacarrillor:
In fact, it also doesn't occur on my machine.

But on @seralra96's machine it does. He tried with your fix and now gets: 
error_project_generator_02

pramisters:
Tested with replaced qt_utils.py:
Same error, but 3 lines later (117) with variable 'reply' instead of 'filename' (line 114).

PR #267, opened 2018-11-13
Keep references to ongoing requests

### LINKED ISSUE #266: Several Interlis models are never loaded from Internet repositories
In the Generate and Import dialogs, some models are never shown in the `Models` field. This affects both Swiss (like `Abfallsammelstellen_ZEBA_LV95_V1`) and Colombian models (like `Catastro_Registro_Nucleo_v2_2_1`). 

This error occurs when using Python v3.7 (we have seen it on Windows and GNU/Linux).

m-kuhn:
if this is not done, only one request will ever be kept alive (garbage collector?)

fix #266


gacarrillor:
Thank you Matthias, this solves the issue!

Issue #176, opened 2018-04-13
path


aweb777:
The java path was not found on my machine. I set the path to e.g. "C:/ProgramData/Oracle/Java/javapath" which didn't work either. Then I set the path to "C:/ProgramData/Oracle/Java/javapath/java.exe" which worked fine (including "/java.exe"). Maybe the annotation "Java Path" in the setting dialog could have a hint/example "C:/ProgramData/Oracle/Java/javapath/java.exe".

m-kuhn:
The placeholder text in the widget can be improved for this.

In general, we would like to make the tool work with 0 configuration. Could you open a python console in your QGIS installation and run the following script, so we can check if/why it is not working without any configuration:

```
import os
print('Path: {}'.format(os.environ['PATH']))
print('Java Home: {}'.format(os.environ['JAVA_HOME']))
p = QProcess()
p.readyReadStandardOutput.connect(lambda: print('{}: {}'.format(proc, p.readAllStandardOutput())))
p.readyReadStandardError.connect(lambda: print('{}: {}'.format(proc, p.readAllStandardError())))
for proc in ['C:/ProgramData/Oracle/Java/javapath/java.exe', 'java', 'java.exe', 'gugus']:
    print('')
    print('-----------')
    p.start(proc, ['-version'])
    print('{} started: {}'.format(proc, p.waitForStarted(3000)))
    p.waitForFinished(3000)
    print('{} exited with {}'.format(proc, p.exitCode()))
```

aweb777:
Does this help?
Python Konsole 
QGIS-API ist über iface zu benutzen. Geben Sie help(iface) ein, um mehr Informationen zu erhalten
import os
Traceback (most recent call last):
  File "C:\PROGRA~1\QGIS3~1.0\apps\Python36\lib\code.py", line 64, in runsource
    code = self.compile(source, filename, symbol)
  File "C:\PROGRA~1\QGIS3~1.0\apps\Python36\lib\codeop.py", line 168, in __call__
    return _maybe_compile(self.compiler, source, filename, symbol)
  File "C:\PROGRA~1\QGIS3~1.0\apps\Python36\lib\codeop.py", line 99, in _maybe_compile
    raise err1
  File "C:\PROGRA~1\QGIS3~1.0\apps\Python36\lib\codeop.py", line 87, in _maybe_compile
    code1 = compiler(source + "\n", filename, symbol)
[...]

Issue #45, opened 2017-06-01
JAVA path


pramisters:
subform for java path selection doesn't open when pressing the button.
Only manual input is possible.

pramisters:
Path selection form works now!
But the automatic search ("standardmässig wird %PATH ... durchsucht") doesn't seem to work, although the JAVA path is present in %PATH.

m-kuhn:
Can you run the following snippet in your python console in QGIS:

```py
from PyQt5.QtCore import QProcess
p = QProcess()
p.start('java')
print(p.exitCode())
```

for me this prints `1`

If I run it for a non-existent command, it prints `255`.

pramisters:
for me this prints  0

m-kuhn:
Can you try it with some non-existent file (e.g. `p.start('jjjjjj')`?

pramisters:
`p.start('gugus')` results in 0 as well

m-kuhn:
What about
`p.start('java.exe')`
and
`p.start('C:/full/path/to/java.exe')`

pramisters:
`p.start('C:\Program Files (x86)\Java\jre1.8.0_144\bin\java.exe')`
results in:
`>>> exec(open(u'C:/Users/KANU/AppData/Local/Temp/tmpcxw7e7ju.py'.encode('utf-8')).read())`
`0`

m-kuhn:
Not a lot of information so far, looks like we need to improve the script a bit

```py
p = QProcess()
p.readyReadStandardOutput.connect(lambda: print('{}: {}'.format(proc, p.readAllStandardOutput())))
p.readyReadStandardError.connect(lambda: print('{}: {}'.format(proc, p.readAllStandardError())))
for proc in ['java', 'java.exe', 'gugus']:
    p.start(proc, ['-version'])
    p.waitForFinished(3000)
    print('{} exited with {}'.format(proc, p.exitCode()))
```

m-kuhn:
@pramisters did you have a chance to run the last test?

pramisters:
java exited with 0
java.exe exited with 0
gugus exited with 0

pramisters:
-----------
java started: False
java exited with 0

-----------
java.exe started: False
java.exe exited with 0

-----------
gugus started: False
gugus exited with 0

m-kuhn:
Thanks
What about the one with the full path annd the environment variable?

pramisters:
I'm not sure what you are referring to.
[...]

PR #282, opened 2019-02-06
[fix] Include double check as java can be found on different paths

lacardonap:
Include double check as java can be found on different paths.


m-kuhn:
Looks good.

Does this fix https://github.com/opengisch/QgisModelBaker/issues/176 and #45 ?

lacardonap:
I checked the  issues and couldn't reproduce them.

PR #293, opened 2019-04-03
Use global env to find java path on Windows

### LINKED ISSUE #283: Java Path
The «standard» Java path works just fine with the «Generate» part of ModelBaker. However, the «Export Interlis» part throws an error saying the Java path be not defined. The user then needs to paste the Java path manually into the ModelBakers settings.

m-kuhn:
Because the QGIS startup script overwrites the env variables they need to be accessed in a different way.

This pull request:

 * Deduplicates code and puts common code blocks into ili2dbutils.
 * Extracts the original Windows `PATH` variable to look for the java executable
 * Checks the java version and raises an error if people have installed an incompatible version

Fix #283

### DIFF
--- a/QgisModelBaker/gui/export.py
+++ b/QgisModelBaker/gui/export.py
@@ -25,9 +25,11 @@ import re
 
 from QgisModelBaker.gui.options import OptionsDialog, ModelListView
 from QgisModelBaker.gui.multiple_models import MultipleModelsDialog
-from QgisModelBaker.libili2db.iliexporter import JavaNotFoundError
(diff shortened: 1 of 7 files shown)


m-kuhn:
@lacardonap you have been fiddling with java paths on Windows too recently. Could you have a look at this pull request and check if this works for you?
Thanks a lot!

lacardonap:
@m-kuhn Thank you!!! Everything works out fine for me. I only have one problem that occurs when I don't have java installed on my computer.

cmd_java_no_found

java_no_install

m-kuhn:
Thank you @lacardonap , that was still an issue indeed.
It should all be fixed now, and make it a lot easier for people to get started.

PR #295, opened 2019-04-11
Fix the error message in file download

yesidpol:
- Allow redirect in file download
- Fix the error message in file download


m-kuhn:
Thanks!

PR #301, opened 2019-05-08
New cleaner icons

m-kuhn:
Big thanks to @zigertiger

image


zigertiger:
volentieri :-)

Let's please just wait with this until after our next meeting (May 23)! We will discuss the new icons in the steering group for they are not yet as good as it gets...

m-kuhn:
Ok, I'll leave it open.
My 2 cents: what about a folder (open) and floppy (save) icon instead of arrows as used in many applications?

m-kuhn:
image

PR #315, opened 2019-06-11
Support disabling data validation on XTF import/export

gacarrillor:
By default it will be false anyways, but this enables us to use the option from the Asistente LADM_COL plugin.

Highly required feature for interchanging (incomplete) data among different actors.


m-kuhn:
Nice move!

Especially for importing data it's super annoying that you don't have any possibility to even load invalid data to fix it.

I was thinking that export/import could propose this after a failed attempt (alongside a nice warning that the data quality is not good) and if the user agrees the import/export step can be repeated with validation disabled.

m-kuhn:
The GUI changes will be done in a follow up step

PR #319, opened 2019-07-08
Avoid critical message that always appears when installing Model Baker

jhonalex06:
This pull requests aims to avoid displaying this message

image

Instead, it displays this message: "The plugin Project Generator is not installed, so it was not removed." to the QGIS log.


m-kuhn:
Thanks for the fix :+1: 
I wonder if it's even worth the message that project generator was not found in the log?

gacarrillor:
In the end it is the same for us, we just wanted to get rid of the red-scary message in the Plugin Manager.

The message made sense in the transition phase from Project Generator to Model Baker. Now, after more than half a year, it has lost relevance.

m-kuhn:
Totally agree, that's why I'd get rid of the message altogether :+1:

PR #324, opened 2019-07-18
Add Sql Server support

yesidpol:
This branch adds support for SQL Server based on pyodbc.

This implementation allows users:

- Load layers with interlis schema
- Load layers without interlis schema. Geometry type is got from stored data in the source similar to how Qgis does it.

Besides, Config panel for SQL Server provides to users a way to choose the right odbc driver in their systems.


QgisModelBaker/gui/panel/mssql_config_panel.py
```diff
@@ -0,0 +1,175 @@
...
+        lbl_schema = QLabel(self.tr("Schema"))
+        lbl_user = QLabel(self.tr("User"))
+        lbl_password = QLabel(self.tr("Password"))
+
+        self.mssql_odbc_driver = QComboBox()
+
+        for item_odbc_driver in get_odbc_drivers():
+            self.mssql_odbc_driver.addItem(item_odbc_driver)
+
+        self.mssql_host_line_edit = QLineEdit()
```
3nids: I believe the parent should be set to `self`:
```suggestion
        self.mssql_host_line_edit = QLineEdit(self)
```

and in following lines the same

QgisModelBaker/gui/panel/mssql_config_panel.py
```diff
@@ -0,0 +1,175 @@
...
+        self.mssql_schema_line_edit = QLineEdit()
+        self.mssql_schema_line_edit.setPlaceholderText(self.tr("[Leave empty to load all schemas in the database]"))
+
+        self.mssql_user_line_edit = QLineEdit()
+        self.mssql_user_line_edit.setPlaceholderText(self.tr("Database Username"))
+
+        self.mssql_password_line_edit = QLineEdit()
+        self.mssql_password_line_edit.setEchoMode(QLineEdit.Password)
+
+        layout = QGridLayout(self)
```
3nids: but not in the case of layout
```suggestion
        layout = QGridLayout()
```

QgisModelBaker/gui/panel/mssql_config_panel.py
```diff
@@ -0,0 +1,175 @@
...
+
+        layout.addWidget(self.mssql_odbc_driver, 0, 1)
+        layout.addWidget(self.mssql_host_line_edit, 1, 1)
+        layout.addWidget(self.mssql_instance_line_edit, 2, 1)
+        layout.addWidget(self.mssql_port_line_edit, 3, 1)
+        layout.addWidget(self.mssql_database_line_edit, 4, 1)
+        layout.addWidget(self.mssql_schema_line_edit, 5, 1)
+        layout.addWidget(self.mssql_user_line_edit, 6, 1)
+        layout.addWidget(self.mssql_password_line_edit, 7, 1)
+
```
3nids: and I guess you should do here instead
```suggestion
self.setLayout(layout)
```

QgisModelBaker/libqgsprojectgen/dataobjects/layers.py
```diff
@@ -27,14 +27,15 @@
...
+    QgsCoordinateReferenceSystem
 )
-from qgis.PyQt.QtCore import QCoreApplication
+from qgis.PyQt.QtCore import QCoreApplication, QSettings
 
 
 class Layer(object):
 
-    def __init__(self, provider, uri, name, extent, geometry_column=None, wkb_type=QgsWkbTypes.Unknown, alias=None, is_domain=False, is_structure=False, is_nmrel=False, display_expression=None):
+    def __init__(self, provider, uri, name, srid, extent, geometry_column=None, wkb_type=QgsWkbTypes.Unknown, alias=None, is_domain=False, is_structure=False, is_nmrel=False, display_expression=None):
```
3nids: a bit offtopic, but it would be nice to start using typehints:
```suggestion
    def __init__(self, provider: str, uri: str, name: str, srid: int, extent: QgsRectangle, geometry_column:str = None, wkb_type: QgsWkbTypes.Type = QgsWkbTypes.Unknown, alias:??? = None, is_domain:bool = False, is_structure:bool = False, is_nmrel:bool = False, display_expression:??? = None):
```

QgisModelBaker/libqgsprojectgen/db_factory/db_factory.py
```diff
@@ -121,3 +121,6 @@ def get_specific_messages(self):
         }
 
         return messages
+
+    def customize_widget_editor(self, field, data_type: str):
```
3nids: can you add documentation to the method?

QgisModelBaker/libqgsprojectgen/dbconnector/mssql_connector.py
```diff
@@ -0,0 +1,511 @@
...
+                query = self._def_cursor(query)
+
+            cur.execute(query)
+
+            if cur.description:
+                columns = [column[0] for column in cur.description]
+
+                for row in cur.fetchall():
+                    my_rec = dict(zip(columns, row))
+                    # TODO type != simple_type
```
3nids: left-over? still todo?

QgisModelBaker/libqgsprojectgen/dbconnector/mssql_connector.py
```diff
@@ -0,0 +1,511 @@
...
+                """.format(schema=self.schema, table=table_name)
+            
+            constraints_cur.execute(query)
+
+            # Create a mapping in the form of
+            #
+            # fieldname: (min, max)
+            constraint_mapping = dict()
+            for constraint in constraints_cur:
+                m = re.match(r"\(\[(.*)\]>=\(([+-]?[0-9]+(?:\.[0-9]+)?(?:[eE][+-]?[0-9]+)?)\) AND \[(.*)\]<=\(([+-]?[0-9]+(?:\.[0-9]+)?(?:[eE][+-]?[0-9]+)?)\)\)", constraint[0])
```
3nids: can you explain in words what does this regex and/or add a link to something like regex101 to show an example?

yesidpol: This method (get_constraints_info) gets the intervals of some number fields. These intervals are stored as **check constraints** into the database.

The query in this method returns the **clause check** intervals that are similar to:

([**numero_pisos**]>=(**1**) AND [**numero_pisos**]<=(**100**))

So, the regex takes the query results and gets the field name, the minimum value, and the maximum value of the field for each register.

QgisModelBaker/gui/panel/mssql_config_panel.py
yesidpol: That was moved to ui file df4dfa1d453b0e1bef7301bf17265a872a4cfc18

QgisModelBaker/gui/panel/mssql_config_panel.py
yesidpol: That was also moved to an ui file df4dfa1d453b0e1bef7301bf17265a872a4cfc18

QgisModelBaker/libqgsprojectgen/dbconnector/mssql_connector.py
3nids: perfect, I believe this would help to write this down in the code as a comment

3nids:
looks good to me

PR #329, opened 2019-07-30
Check that ili file exists before refreshing cache

jhonalex06:
There is an issue when the user has a .xtf loaded in the menu (Import Interlis Transfer File .xtf), but the file does not exist anymore.

image

Gif of the issue:

gif_issue


m-kuhn:
Thanks

jhonalex06:
> Thanks

You are welcome

PR #330, opened 2019-08-05
[fix] Not trying to get models when the scheme is not interlis.

lacardonap:


m-kuhn:
Thanks, can you add a description what problem this solves?

lacardonap:
When trying to get the models of a schema that does not have INTERLIS the following error is generated because the table "t_ili2db_model" does not exist.

get_models_pg

m-kuhn:
Thanks!

PR #331, opened 2019-08-05
[gui] load panel settings from ui file

lacardonap:
This PR moves the GUI part from Python code to UI files for easing maintenance.

### DIFF
--- a/QgisModelBaker/gui/panel/gpkg_config_panel.py
+++ b/QgisModelBaker/gui/panel/gpkg_config_panel.py
@@ -17,10 +17,6 @@
  ***************************************************************************/
 """
 from qgis.PyQt.QtCore import pyqtSignal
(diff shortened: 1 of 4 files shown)


m-kuhn:
Thanks, does this change the UI?
Can you add a screenshot in this case?

lacardonap:
The UI does not change.

**PostgreSQL settings panel**
pg_settings_panel

**Example: Import data with PostgreSQL**

import_interlist_data_pg


**GeoPackage settings panel**
gpkg_settings_panel

**Example: Import data with GeoPackage**

import_interlist_data_gpkg

m-kuhn:
> The UI does not change.

Then, what does it change?

gacarrillor:
@m-kuhn, this PR moves the GUI part from Python code to UI files for easing maintenance.

PR #347, opened 2019-10-21
Update SQL Server support to Ili2db4

yesidpol:
This PR updates SQL server support to ili2db4 and also adds the required comments for the regex (see https://github.com/opengisch/QgisModelBaker/pull/324#issuecomment-530759530)


QgisModelBaker/libqgsprojectgen/dbconnector/mssql_connector.py
```diff
@@ -0,0 +1,409 @@
...
+
+        return False
+
+    def get_tables_info(self):
+        res = []
+
+        if self.schema:
+            metadata_exists = self.metadata_exists()
+
+            ln = "\n"
```
3nids: I tend to prefer to write SQL in Python as follows
```python
           stmt = """
SELECT distinct
    tbls.TABLE_SCHEMA AS schemaname
  , tbls.TABLE_NAME AS tablename
                  """.format(
                      var1, var2, …
                  )

```

I find it easier to read as the SQL pops out but that's up to you.

QgisModelBaker/tests/test_export.py
```diff
@@ -132,6 +132,50 @@ def test_export_postgis(self):
         self.compare_xtfs(testdata_path(
             'xtf/test_ciaf_ladm.xtf'), obtained_xtf_path)
 
+    def test_export_mssql(self):
+        # Schema Import
```
3nids: the import/export tests look very similar. Could you maybe make a base class (or method?) for them and just call them with the specification for each provider?

QgisModelBaker/tests/test_import.py
```diff
@@ -178,6 +179,85 @@ def test_import_geopackage(self):
         cursor.close()
         conn.close()
 
+    def test_import_mssql(self):
+
+        # Schema Import
+        importer = iliimporter.Importer()
```
3nids: the import/export tests look very similar. Could you maybe make a base class (or method?) for them and just call them with the specification for each provider?

PR #351, opened 2019-11-07
Support uppercase file extensions

signedav:
Uppercase extensions are valid now.

Otherwise it's hard to find the file, when its extension is written in uppercase letters.
The general validation is still case-sensitive if the local file system is case sensitive (e.g. unix afaik).

### DIFF
--- a/QgisModelBaker/gui/export.py
+++ b/QgisModelBaker/gui/export.py
@@ -107,7 +107,7 @@ class ExportModels(QStringListModel):
 
 
 class ExportDialog(QDialog, DIALOG_UI):
-    ValidExtensions = ['xtf', 'itf', 'gml', 'xml']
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -28,11 +28,13 @@ jobs:
     - stage: deploy
       if: tag IS present
       install:
+        - apt install python3-pyodbc
```
m-kuhn: What's this for?

signedav: it has been a test.
travis deploy is failing because of pyodbc in the requirement.txt.
At my local machine I had to install it like this, to make it run. So I tried it out here as well. 
Maybe you know how to fix it correctly.

m-kuhn: I think it's for the mssql tests. It makes sense I guess.
But... if it's required for master to pass it should already be in there, that's what's confusing to me.

signedav: Yes. Thought about that as well. But if there are no changes of travis needed because everything should work fine, why it fails on the requirement script?

PR #353, opened 2019-11-11
Try with getdefaultlocale() if getlocale() fails

marioba:
Fix importing error on Windows if the current locale is different than `UTF-8` and `locale.gelocale()` returns `None`.
Seems that `locale.getdefaultlocale()` (that let python try to determine the correct locale) works on these situations.

### DIFF
--- a/QgisModelBaker/libili2db/iliimporter.py
+++ b/QgisModelBaker/libili2db/iliimporter.py
@@ -61,7 +61,7 @@ class Importer(QObject):
             self.configuration = SchemaImportConfiguration()
         self.encoding = locale.getlocale()[1]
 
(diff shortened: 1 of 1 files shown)


QgisModelBaker/libili2db/iliimporter.py
```diff
@@ -60,6 +60,11 @@ def __init__(self, dataImport=False, parent=None):
         else:
             self.configuration = SchemaImportConfiguration()
         self.encoding = locale.getlocale()[1]
+
+        # Let's python try to determine the default locale
```
m-kuhn:
```suggestion
        # Lets python try to determine the default locale
```

PR #354, opened 2019-11-15
Refresh layers (including feature count) after data import

signedav:
After a data import it goes through all the layers, resets the data source and featureCount with the QGIS method `reloadData()` https://github.com/qgis/QGIS/pull/33259 and reloads the layer in the legend (to trigger the getter of the feature count).

**Before:**
After data import, canvas are not refreshed (still empty until first zoom or whatever) and the feature count has been - even after the features are loaded on the canvas - the same like before the data import...

### DIFF
--- a/QgisModelBaker/gui/import_data.py
+++ b/QgisModelBaker/gui/import_data.py
@@ -213,7 +213,7 @@ class ImportDataDialog(QDialog, DIALOG_UI):
     def refresh_layers(self):
         # refresh layers
         for layer in self.iface.mapCanvas().layers():
-            layer.setDataSource(layer.source(), layer.name(), layer.providerType())
+            layer.dataProvider().forceReload()
         self.iface.layerTreeView().layerTreeModel().recursivelyEmitDataChanged()
 
     def print_info(self, text, text_color='#000000'):
--- a/QgisModelBaker/gui/import_data.py
+++ b/QgisModelBaker/gui/import_data.py
@@ -214,8 +214,7 @@ class ImportDataDialog(QDialog, DIALOG_UI):
         # refresh layers
         for layer in self.iface.mapCanvas().layers():
             layer.setDataSource(layer.source(), layer.name(), layer.providerType())
-            self.iface.layerTreeView().setCurrentLayer(layer)
-        self.iface.layerTreeView().setCurrentLayer(None)
+        self.iface.layerTreeView().layerTreeModel().recursivelyEmitDataChanged()
 
     def print_info(self, text, text_color='#000000'):
         self.txtStdout.setTextColor(QColor(text_color))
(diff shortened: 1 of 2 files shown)


signedav:
Made it a little bit better by using `self.iface.layerTreeView().layerTreeModel().recursivelyEmitDataChanged()` to reload the legend instead of imitate a selecting of every single legend-node. 

Still, I don't find a better solution to reload the dataSource at all, than using this hack:
`layer.setDataSource(layer.source(), layer.name(), layer.providerType())`

This would load it, that they appear on the canvas:
```
    layer.setLegend(QgsMapLayerLegend.defaultVectorLegend(layer))
    layer.dataSourceChanged.emit()
    layer.triggerRepaint()
```
But it does not work to reload the legend after it.

Do you have a good idea @m-kuhn ?

I decided against creating a function like `dataSourceChanged` on QGIS containing the commands above, since it did not solved it. If they would, a function like this would maybe make sense.

m-kuhn:
@signedav what is the status of this?
I think I don't have all the context to make a decision, please let me know if we need to talk this though again.

signedav:
Hi @m-kuhn 

I would merge it and invest more time later if you agree or if you see a better approach.

Like I mentioned in the last comment, I'm happy triggering `recursivelyEmitDataChanged` to reload the legend (and the feature count) but this is only the small part of it. The main issue is that it does not reload the data to the layer at all (it never did). It's done in QGIS in `setDataSource` so I just call this function with the current data. I agree it's not that nice.
So I tried out to call the function that are needed to have the data in the layer:
```
    layer.setLegend(QgsMapLayerLegend.defaultVectorLegend(layer))
    layer.dataSourceChanged.emit()
    layer.triggerRepaint()
```
But this does not make it. Maybe I missed something. But before investing more time in finding out what I missed I ask you if you would agree on making a new QGIS function called something like "dataSourceChange" reloading everything with the current data (without setting the data source) or if you prefer to have it like it is right now.

Thanks

m-kuhn:
If your proposal is to add

```
QgsVectorLayer::forceReloadDataSource()
{
  setLegend(QgsMapLayerLegend.defaultVectorLegend(layer))
  dataSourceChanged.emit()
  triggerRepaint()
}
```

I am very much in favor :+1:

signedav:
Needs https://github.com/qgis/QGIS/pull/33259

signedav:
@m-kuhn how can we make this release reliable on the latest QGIS?

signedav:
@m-kuhn What is the plan with this? Can we make it reliable on the latest QGIS?

signedav:
IMO we can merge it - though it wont do it with QGIS earlier than https://github.com/qgis/QGIS/pull/33259

m-kuhn:
This will raise an exception on earlier versions?
You probably wanted to say depend/rely instead of reliable? I agree. Just put a `try: except AttributeError:` around it and it should be good to go.i

signedav:
Sorry yes. Mixed up with rely and reliable. Should have asked Frau Binder instead :-)

PR #363, opened 2020-01-27
Bag Of creating Value Relations with ili2db 4.4.0

### LINKED ISSUE #182: Constraint for arrays that need to contain at least one element
If a BAG-OF-construct requires at least one element (i.e. BAG {1..*} OF...) the projectgenerator should configure the form so that at least one element has to be selected before the green tick appears.
Currently the green tick appears even if no values are selected. Hence data can be saved that is not ili-model-compliant.

bag_of_1_n

The prerequisite seem to be missing in the PostGIS metatables:
https://github.com/claeis/ili2db/issues/171

### LINKED ISSUE #359: ili2db4 - BAG OF Value Relations not supported anymore
With ili2pg4 the BAG_OF columns creating Value Relation Widgets are no longer suported.

How comes?
While with ili2db3 the `domain_relation_generator` creates the domain relations, with ili2db4 that's not needed anymore since the relations are created by ili2db. But the BAG_OF_ENUM list is not created, since this has been part of the `domain_relation_generator`. 
So first thing would be to bring the functionality back to get the BAG_OF_ENUM list.

If so, then there is another challenge. 
While the `domain_relation_generator` used the string-column `ilicode` as the key column in BAG_OF_ENUM (and then the Value Relation Widget) the column for the BAG_OF in ili2db4 is created is not a string-array anymore (`bigint[]` instead of `character varying(255)[]`). There is a suggestion to implement this column as a JSON-field, but not really cared about claeis/ili2db#218 Could we just use the `t_id` instead, or why we decided for `ilicode`? 
Using another key could lead to issues on data migration from ili2db3 schema to ili2db4 schema...

signedav:
Information for Bag-Of-Structures are read from the meta tables:
- t_ili2db_column_prop
- t_ili2db_meta_attrs
- t_ili2db_classname

The cardinality influences the Value Relation settings like this:
- Minimum Value: if 1 the constraint expression is set like `array_length("valuerelation_1")>0` means at least one item should be selected
- Maximum Value: no influece since it should be always * If not, then it's not an ARRAY and not created as a value relation

`Allow Null` and `Multi Selection` are always true, since the list is never NULL and if it should not be a multi selection the widget will be a Relation Reference.

- [x] Simple tests to check if the expression is created regarding the cardinality

Fix #359 
Fix #182

### DIFF
--- a/QgisModelBaker/libqgsprojectgen/dataobjects/project.py
+++ b/QgisModelBaker/libqgsprojectgen/dataobjects/project.py
@@ -139,27 +139,28 @@ class Project(QObject):
                 key_field = bag_of_enum_info[3]
                 value_field = bag_of_enum_info[4]
 
-                allow_null = cardinality.startswith('0')
-                allow_multi = cardinality.endswith('*')
+                minimal_selection = cardinality.startswith('1')
 
                 current_layer = layer_obj.create()
 
                 field_widget = 'ValueRelation'
                 field_widget_config = {
-                    'AllowMulti': allow_multi,
+                    'AllowMulti': True,
                     'UseCompleter': False,
                     'Value': value_field,
                     'OrderByValue': False,
-                    'AllowNull': allow_null,
+                    'AllowNull': True,
                     'Layer': domain_table.create().id(),
                     'FilterExpression': '',
                     'Key': key_field,
                     'NofColumns': 1
                 }
-
                 field_idx = current_layer.fields().indexOf(attribute)
                 setup = QgsEditorWidgetSetup(field_widget, field_widget_config)
                 current_layer.setEditorWidgetSetup(field_idx, setup)
+                if minimal_selection:
+                    constraint_expression = 'array_length("{}")>0'.format(attribute)
+                    current_layer.setConstraintExpression(field_idx, constraint_expression, 'The minimal selection is 1')
 
         for layer in self.layers:
             layer.create_form(self)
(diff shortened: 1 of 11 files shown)


m-kuhn:
Does this fix #359 ?

signedav:
Yes. And the initial issue reported by @romefi with the "Periodische Arbeiten" as well.

QgisModelBaker/libqgsprojectgen/dataobjects/project.py
```diff
@@ -139,27 +139,28 @@ def create(self, path: str, qgis_project: QgsProject):
...
                     'Key': key_field,
                     'NofColumns': 1
                 }
-
                 field_idx = current_layer.fields().indexOf(attribute)
                 setup = QgsEditorWidgetSetup(field_widget, field_widget_config)
                 current_layer.setEditorWidgetSetup(field_idx, setup)
+                if minimal_selection:
+                    constraint_expression = 'array_length("{}")>0'.format(attribute)
+                    current_layer.setConstraintExpression(field_idx, constraint_expression, 'The minimal selection is 1')
```
m-kuhn: Translate?

QgisModelBaker/libqgsprojectgen/generator/generator.py
```diff
@@ -236,8 +236,22 @@ def relations(self, layers, filter_layer_list=[]):
                 layers)
             relations = relations + domain_relations
         else:
+            """Create the bags_of_enum structure"""
```
m-kuhn:
```suggestion
            # Create the bags_of_enum structure
```

The other notation is mainly for docstrings

m-kuhn:
Can you add `Fix #359` to the PR description so it's closed and linked?

PR #370, opened 2020-01-31
Fix get models

jhonalex06:


m-kuhn:
Any chance to get a description of what was wrong / is being fixed?

jhonalex06:
Yes, @m-kuhn I will add the description in a few minutes and the solution

jhonalex06:
Current get_models() function does not consider the case when several models are in one .ili file. For example, in [Insumos LADM_COL_V2_9_6](https://github.com/AgenciaImplementacion/LADM_COL/blob/ladmcol2_9_6/Catastro_Multiproposito/Operacion/Insumos_V2_9_6.ili) we have 3 models in one ili file.

When we run a schema import with only one of those 3 models in QgisModelBaker, the export interface shows all 3 models (only one should be shown).

EXAMPLE:

First, we create a project with Datos_Gestor_Catastral_V2_9_6 Project. 

image

Then,  we try to export the _.xtf_ file and we found models that we did not create, like:   
_LADM_COL_V1_3, Datos_SNR_V2_9_6 and  Datos_Integracion_Insumos_ V2_9_6_. 

image

Finally, when we fix the get_models function, we obtain only the models created. 

image

m-kuhn:
Tested and works, thanks

PR #390, opened 2020-03-19
Remove condition when parsing model dirs in path; add tests

gacarrillor:


gacarrillor:
Oops, they ran fine locally. I'll fix that.

gacarrillor:
Transifex issue this time.

m-kuhn:
That step should probably be skipped when a PR is done from a different repo.
Thanks!

PR #396, opened 2020-04-01
Update to ili2db 4.4.2 (ili2mssql to 4.4.1)

gacarrillor:
Some tests were failing, I had to adjust them. 

Updating to ili2mssql 4.4.2 is not safe because of a problem with the instance, so, just update it to 4.4.1 in the meantime.


m-kuhn:
What were the failures?
Something that is now broken?

gacarrillor:
For some reason, updating to ili2db 4.4.2 forced to add an ABSTRACT keyword in a topic on a sample model. Same for one sample GPKG database, since it contains the whole model in an INTERLIS table.

Regarding the XTF, yes, the expected XTF had an empty XML element. It doesn't seem to be the case anymore with ili2db 4.4.2, since the element does not appear after an export.

Finally, I didn't expect the t_ids issue, just adjust them to what the run said and then everything was OK. 

BTW, is travis not running anymore on PRs?

m-kuhn:
@gacarrillor do you have an error message?

@claeis is the new requirement for `ABSTRACT` in the model (https://github.com/opengisch/QgisModelBaker/pull/396/files#diff-85747a963bb3b06bb6a4fb7d28697a03L631) expected and safe or something that needs attention on ili2db side?

We recently migrated the organisation account to travis-ci beta because we experienced some issues with the old setup. This repo was not yet migrated.

gacarrillor:
This is the error message for the ABSTRACT issue: 

`ERROR    root:test_projectgen.py:1192 Error: /usr/src/QgisModelBaker/tests/testdata/ilimodels/CIAF_LADM/Catastro_COL_ES.ili:631:Catastro_COL_ES_V2_1_6.Catastro_Registro must be declared as ABSTRACT. One of its elements (UnidadConstruccion) is abstract.`

claeis:
It is no new requirement, just an improved checking of the ili-model. The model was not correct: the TOPIC Catastro_Registro must be abstract because it contains abstract classes (e.g. Alerta).

PR #404, opened 2020-04-25
Support for other SRS authorities

gacarrillor:
and pass --defaultSrsAuth parameter to ili2db. 

Handles ili2db non-valid SRSs like USER:1234 (only valid for a single machine running QGIS) or srs codes that are not integers.

### DIFF
--- a/QgisModelBaker/gui/generate_project.py
+++ b/QgisModelBaker/gui/generate_project.py
@@ -496,7 +496,7 @@ class GenerateProjectDialog(QDialog, DIALOG_UI):
         if  srs_auth == 'USER':
             self.crs_label.setStyleSheet('color: orange')
             self.crs_label.setToolTip(
-                self.tr('Please select a valid Coordinate Reference System.\nCRSs from USER are valid for a single computer and therefore, not recommended.\nA default EPSG:21781 will be used.'))
+                self.tr('Please select a valid Coordinate Reference System.\nCRSs from USER are valid for a single computer and therefore, a default EPSG:21781 will be used instead.'))
         else:
             self.crs_label.setStyleSheet('')
             self.crs_label.setToolTip(self.tr('Coordinate Reference System'))
--- a/QgisModelBaker/gui/generate_project.py
+++ b/QgisModelBaker/gui/generate_project.py
@@ -372,7 +372,8 @@ class GenerateProjectDialog(QDialog, DIALOG_UI):
         self._lst_panel[db_id].get_fields(configuration)
 
         configuration.tool = mode
-        configuration.epsg = self.epsg
+        configuration.srs_auth = self.srs_auth
+        configuration.srs_code = self.srs_code
         configuration.inheritance = self.ili2db_options.inheritance_type()
         configuration.tomlfile = self.ili2db_options.toml_file()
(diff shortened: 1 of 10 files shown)


QgisModelBaker/gui/generate_project.py
```diff
@@ -484,16 +490,24 @@ def update_crs_info(self):
...
-        if self.crsSelector.crs().authid()[:5] != 'EPSG:':
+        self.srs_auth = 'EPSG'  # Default
+        self.srs_code = 21781  # Default
+        srs_auth, srs_code = self.crsSelector.crs().authid().split(":")
+        if  srs_auth == 'USER':
             self.crs_label.setStyleSheet('color: orange')
             self.crs_label.setToolTip(
-                self.tr('Please select an EPSG Coordinate Reference System'))
-            self.epsg = 21781
+                self.tr('Please select a valid Coordinate Reference System.\nCRSs from USER are valid for a single computer and therefore, not recommended.\nA default EPSG:21781 will be used.'))
```
m-kuhn, line 499: Are they "not recommended" or "will 21781 be used"?
From the description it looks like using USER crs'es should be supported but here it looks like they are not accepted. How does that work exactly?

QgisModelBaker/libili2db/ili2dbconfig.py
```diff
@@ -225,8 +226,11 @@ def to_ili2db_args(self, extra_args=[], with_action=True):
         if self.create_basket_col:
             args += ["--createBasketCol"]
 
-        if self.epsg != 21781:
-            args += ["--defaultSrsCode", "{}".format(self.epsg)]
+        if self.srs_auth != 'EPSG':
+            args += ["--defaultSrsAuth", self.srs_auth]
+
+        if self.srs_code != 21781:
```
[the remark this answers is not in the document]
gacarrillor: Yes! I think we should. I have seen those messages as well when testing our models with a default SRS. Not the case with Swiss models, though.

QgisModelBaker/gui/generate_project.py
gacarrillor: I'm 50%-50% here. The only issue I see is that when using a `USER:XXXX`, the SRS code is not guaranteed to remain in different QGIS installations. Other than that, I'm fine if we remove the `USER` conditional.

m-kuhn: How about a warning (but letting the final choice to the user)? What does ili2db do with that anyway?

gacarrillor: ili2db's schema import fails if the SRS is not found in the spatial reference table of the target DB.
I'm fine with the warning. I'll adjust it.

m-kuhn: Let's remove the "not recommended" part. If ili2db does not support them we should not allow them.

gacarrillor: Done!

QgisModelBaker/libili2db/ili2dbconfig.py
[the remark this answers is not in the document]
gacarrillor: Fixed removing the if, ensuring the defaultSrsCode is always passed to ili2db in Schema Import.

PR #405, opened 2020-04-28
Optionally, add preScript and postScript parameters to (schema) import

gacarrillor:
For both API and GUI.

### DIFF
--- a/QgisModelBaker/gui/ili2db_options.py
+++ b/QgisModelBaker/gui/ili2db_options.py
@@ -67,12 +67,12 @@ class Ili2dbOptionsDialog(QDialog, DIALOG_UI):
                                file_filter=self.tr('SQL script to run after (*.sql *.SQL)')))
 
         self.validators = Validators()
-        self.fileValidator = FileValidator(pattern=['*.' + ext for ext in self.ValidExtensions], allow_empty=True)
-        self.toml_file_line_edit.setValidator(self.fileValidator)
+        self.file_validator = FileValidator(pattern=['*.' + ext for ext in self.ValidExtensions], allow_empty=True)
+        self.toml_file_line_edit.setValidator(self.file_validator)
 
-        self.SQLFileValidator = FileValidator(pattern=['*.' + ext for ext in self.SQLValidExtensions], allow_empty=True)
-        self.pre_script_file_line_edit.setValidator(self.SQLFileValidator)
-        self.post_script_file_line_edit.setValidator(self.SQLFileValidator)
+        self.sql_file_validator = FileValidator(pattern=['*.' + ext for ext in self.SQLValidExtensions], allow_empty=True)
+        self.pre_script_file_line_edit.setValidator(self.sql_file_validator)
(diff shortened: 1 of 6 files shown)


gacarrillor:
Note that we avoid recalling used prescript and postscript files in the ili2db options dialog. 
I think it could be better to let the user configure them each time the scripts are needed.

Let me know if you think we should recall them.

QgisModelBaker/gui/ili2db_options.py
```diff
@@ -48,19 +59,42 @@ def __init__(self, parent=None):
...
+                               file_filter=self.tr('SQL script to run before (*.sql *.SQL)')))
+        self.post_script_file_browse_button.clicked.connect(
+            make_file_selector(self.post_script_file_line_edit, title=self.tr('SQL script to run after (*.sql)'),
+                               file_filter=self.tr('SQL script to run after (*.sql *.SQL)')))
+
         self.validators = Validators()
         self.fileValidator = FileValidator(pattern=['*.' + ext for ext in self.ValidExtensions], allow_empty=True)
         self.toml_file_line_edit.setValidator(self.fileValidator)
 
+        self.SQLFileValidator = FileValidator(pattern=['*.' + ext for ext in self.SQLValidExtensions], allow_empty=True)
```
m-kuhn, line 73:
```suggestion
        self.sql_file_validator = FileValidator(pattern=['*.' + ext for ext in self.SQLValidExtensions], allow_empty=True)
```

pep8

QgisModelBaker/gui/ili2db_options.py
```diff
@@ -48,19 +59,42 @@ def __init__(self, parent=None):
...
+
+        self.bar = QgsMessageBar()
+        self.bar.setSizePolicy(QSizePolicy.Minimum, QSizePolicy.Fixed)
+        self.layout().addWidget(self.bar, 0, 0, Qt.AlignTop)
 
     def accepted(self):
         """ Save settings before accepting the dialog """
+        for line_edit in [self.pre_script_file_line_edit, self.post_script_file_line_edit, self.toml_file_line_edit]:
+            if line_edit.validator().validate(line_edit.text().strip(), 0)[0] != QValidator.Acceptable:
+                self.bar.pushWarning("Warning", "Please fix the '{}' value before saving the options.".format(
```
m-kuhn, line 94: Translate?

QgisModelBaker/gui/ili2db_options.py
gacarrillor: Done!

QgisModelBaker/gui/ili2db_options.py
gacarrillor: Natürlich!

m-kuhn:
About remembering the value for pre- and postscript I agree they are powerful (and dangerous occasionally). Let's stick to the "reconfigure every time" approach you proposed. We can still fiddle and fine tune that if it gets annoying.

PR #407, opened 2020-05-05
Update ili2db tool versions to v4.4.3

gacarrillor:
Notes: 
 + The URL to download ili2db tools has changed to a more standard URL that is going to be maintained.
 + t_ids in a test have been adjusted. I guess that sometimes new ili2db versions impose those adjustments, since it's not the first time it happens in some tests.

--------------

:warning:  8 tests fail due to stronger constraint data type checks in ili2c (see https://github.com/claeis/ili2c/issues/17). 

All 8 tests fail when using the model `ExceptionalLoadsRoute_LV95_V1`, throwing the following error: 

> ERROR    root:test_gpkg_pk.py:95 Error: /usr/src/QgisModelBaker/tests/testdata/ilimodels/CHBase_Part4_ADMINISTRATIVEUNITS_20110830.ili:142:incompatible datatypes

It's a Swiss model, so I wouldn't like to adjust the .ili file before bringing this topic here for some discussion.

### DIFF
--- a/QgisModelBaker/tests/test_import.py
+++ b/QgisModelBaker/tests/test_import.py
@@ -85,7 +85,7 @@ class TestImport(unittest.TestCase):
             """.format(importer.configuration.dbschema))
         record = next(cursor)
         self.assertIsNotNone(record)
-        self.assertEqual(record[0], 38)  # t_id for 'Unidad_Derecho'
+        self.assertEqual(record[0], 6)  # t_id for 'Unidad_Derecho'
         self.assertEqual(record[1], 'POLYGON((1000257.426 1002020.376,1000437.688 1002196.495,1000275.472 1002428.19,1000072.25 1002291.539,1000158.572 1002164.914,1000159.942 1002163.128,1000257.426 1002020.376))')
         self.assertEqual(record[2], 3116)
         predio_id = record[3]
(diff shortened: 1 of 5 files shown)


gacarrillor:
Current status: 

10 tests failing:

- 8 because of data type checks for `CHBase_Part4_ADMINISTRATIVEUNITS_20110830.ili`
- 2 because ili2mssql v3.12.2 cannot be found at https://downloads.interlis.ch/

gacarrillor:
Current status: Only 8 tests are failing because of data type checks for  `CHBase_Part4_ADMINISTRATIVEUNITS_20110830.ili`

@m-kuhn How could we proceed? Do you want me to modify that local (in test resources) model so that tests pass?

m-kuhn:
Hi @gacarrillor I am very sorry for the long delay.

I looked into it and can confirm that there is an issue in the model / test data which is not an issue of model baker itself.
I also agree on your proposal. Can you update this PR as required? Once that's done in here, I'd like to propose this fix upstream.

QgisModelBaker/tests/test_import.py
```diff
@@ -85,7 +85,7 @@ def test_import_postgis(self):
             """.format(importer.configuration.dbschema))
         record = next(cursor)
         self.assertIsNotNone(record)
-        self.assertEqual(record[0], 38)  # t_id for 'Unidad_Derecho'
+        self.assertEqual(record[0], 6)  # t_id for 'Unidad_Derecho'
```
m-kuhn, line 88: Is it meaningful to test these ids?

gacarrillor: Not that much. We can remove them in an coming commit.

claeis:
Update your local copy of CHBase_Part4_ADMINISTRATIVEUNITS_20110830.ili (line 142 should read as ```UpperLevelUnit->Code == #CHE;```)

gacarrillor:
This PR is definitely team work!

PR #408, opened 2020-06-06
Add support for authentication configuration

### LINKED ISSUE #391: Project generation should offer to use authconfig
We should avoid storing usernames/passwords in QGIS projects generated by Model Baker.

Model Baker should offer (by default) to use one of the existing authconfig configurations.

m-kuhn:
This pull requests adds compatibility for the auth management framework in QGIS. It will only store an authid in the QGIS project. The same authid will need to be configured on every QGIS profile that the project is opened.

Peek 2020-06-06 15-25

Fixes #391

TODO:

 - [ ] Auth config **requires** a super user password to be configured in settings. Make this clear and straightforward to configure.
 - [ ] Unit tests (if possible?)
 - [ ] Check the UX with geopackage
 - [ ] Validate import

### DIFF
--- a/QgisModelBaker/gui/export.py
+++ b/QgisModelBaker/gui/export.py
@@ -178,7 +178,7 @@ class ExportDialog(QDialog, DIALOG_UI):
         self.export_models_view.setModel(self.export_models_model)
         self.export_models_view.clicked.connect(self.export_models_model.check)
         self.export_models_view.space_pressed.connect(self.export_models_model.check)
-        self.refresh_models()
+        self.request_for_refresh_models()
 
         self.type_combo_box.currentIndexChanged.connect(self.type_changed)
 
--- a/QgisModelBaker/gui/export.py
+++ b/QgisModelBaker/gui/export.py
@@ -115,7 +115,7 @@ class ExportDialog(QDialog, DIALOG_UI):
         QDialog.__init__(self, parent)
         self.setupUi(self)
         self.db_simple_factory = DbSimpleFactory()
-        QgsGui.instance().enableAutoGeometryRestore(self);
+        QgsGui.instance().enableAutoGeometryRestore(self)
         self.buttonBox.accepted.disconnect()
         self.buttonBox.clicked.connect(self.button_box_clicked)
         self.buttonBox.clear()
--- a/QgisModelBaker/gui/export.py
+++ b/QgisModelBaker/gui/export.py
@@ -203,7 +203,7 @@ class ExportDialog(QDialog, DIALOG_UI):
 
         db_factory = self.db_simple_factory.create_factory(tool)
         config_manager = db_factory.get_db_command_config_manager(configuration)
-        uri_string = config_manager.get_uri()
+        uri_string = config_manager.get_uri(configuration.db_use_super_login)
 
         db_connector = None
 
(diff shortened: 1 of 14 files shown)


m-kuhn:
@schmandr @andreasneumann is this what you had in mind?

andreasneumann:
Yes - that's exactly what we want.

The other thing we would be interested is support for "service" name alternative to the host/port/db combo.

Just like you can define connections in QGIS itself.

Should I open a separate ticket for this enhancement?

m-kuhn:
> Yes - that's exactly what we want.

Thanks

> The other thing we would be interested is support for "service" name alternative to the host/port/db combo.
> 
> Just like you can define connections in QGIS itself.
> 
> Should I open a separate ticket for this enhancement?

Yes, please do

andreasneumann:
> > Should I open a separate ticket for this enhancement?
> 
> Yes, please do

ok - see https://github.com/opengisch/QgisModelBaker/issues/413

m-kuhn:
@andreasneumann @gacarrillor this pull request is ready.

One thing to mention is that import/export xtf does not work with authcfg. These tasks are done through ili2db and we need to have the password for it in clear text.
IMO this considerably decreases usability. The concept of "super user" so far was for management jobs (create schema). In combination with authcfg it's now "anything which is ili2db or database inspection" --> creating a database, doing interlis imports and doing interlis exports.
Is the current behavior ok for your use case?

gacarrillor:
@m-kuhn Just to make things clear. This was/is not a requirement from us.
In fact, we do not use the super user (or authcfg) option.

Having said that and if I understood well, from a user perspective, it would be a bit strange if the schema import can be done with no password in clear text, but on the contrary, both import/export xtf can.

m-kuhn:
> @m-kuhn Just to make things clear. This was/is not a requirement from us (SwissTierras Colombia).

Yes, this was from Solothurn mainly.

> Having said that and if I understood well, from a user perspective, it would be a bit strange if the schema import can be done with no password in clear text, but on the contrary, both import/export xtf can.

That's the kind of input I was looking for anyway, thanks !

andreasneumann:
> Is the current behavior ok for your use case?

Do you see any other options? The auth system probably doesn't allow to extract username/password from the auth-db  with Python - does it? If it does, it could get the credentials from the auth db and forward it to ili2db ?

I am not a security expert and don't know about the internals of the auth-db system.

m-kuhn:
I don't think it's easily feasible to forward the stored password to ili2pg.

From the workflows in your organisation:

 - does it still make sense (is import/export of xtf an administration job)
 - do we want a second password (interactive on import/export) next to the admin password? Does this defeat the purpose of this pull request?

andreasneumann:
> * do we want a second password (interactive on import/export) next to the admin password? Does this defeat the purpose of this pull request?

I think an interactive password prompt for xtf import/export would make sense. It is not a task you do too often, and often it is other people than admins doing import/export. These people wouldn't have the admin credentials.

andreasneumann:
> I don't think it's easily feasible to forward the stored password to ili2pg.

Probably right that it is like it is. It would be a security hole if you could extract auth stuff with Python - although you would still need the master pw.

m-kuhn:
Can I read from your responses that "even if it's not possible to integrate import/export with authcfg, it's still worth this functionality in QGIS Model Baker"?

andreasneumann:
> Can I read from your responses that "even if it's not possible to integrate import/export with authcfg, it's still worth this functionality in QGIS Model Baker"?

Yes, definitely. Storing passwords in QGIS project files is not acceptable.

m-kuhn:
Then I will merge and release and looking forward to your testing. @gacarrillor and others, also interested in your feedback from real world testing !

PR #414, opened 2020-06-26
Import data: Delete checkbox optimization

### LINKED ISSUE #361: Import with delete existing data confusion
On data import there is the nice possibility to delete or keep existing data controlled over the checkbox. I made the mistake by leaving it checked after I imported catalogue data and wanted to import my real data. I kind of assumed  that it would nod delete the catalogue data. My mistake, for sure, but still I think maybe we could improve something here.

At least add a tooltip. But another thing would be to have this option only when there is actually existing data. So the user knows that something will be deleted in fact.
And another step would be that there are options for "delete all data" / "delete catalogue data" / etc...

signedav:
It's about this:
image

- having the checkbox unchecked on open of the dialog to avoid unintended overwriting and mostly you don't need it on second import.

- tooltip to get informed that catalogues are deleted as well

fixes #361

### DIFF
--- a/QgisModelBaker/gui/import_data.py
+++ b/QgisModelBaker/gui/import_data.py
@@ -372,12 +372,11 @@ class ImportDataDialog(QDialog, DIALOG_UI):
 
     def restore_configuration(self):
         settings = QSettings()
-
         self.fill_toml_file_info_label()
         self.xtf_file_line_edit.setText(settings.value(
             'QgisModelBaker/ili2pg/xtffile_import'))
-        self.chk_delete_data.setChecked(settings.value(
-            'QgisModelBaker/ili2pg/deleteData', False, bool))
+        # set chk_delete_data always to unchecked
+        self.chk_delete_data.setChecked(False)
 
         for db_id in self.db_simple_factory.get_db_list(False):
             configuration = iliimporter.ImportDataConfiguration()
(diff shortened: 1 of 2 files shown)


QgisModelBaker/gui/import_data.py
```diff
@@ -372,12 +372,11 @@ def save_configuration(self, configuration):
 
     def restore_configuration(self):
         settings = QSettings()
-
         self.fill_toml_file_info_label()
         self.xtf_file_line_edit.setText(settings.value(
             'QgisModelBaker/ili2pg/xtffile_import'))
-        self.chk_delete_data.setChecked(settings.value(
-            'QgisModelBaker/ili2pg/deleteData', False, bool))
+        # set chk_delete_data always to unchecked
```
m-kuhn, line 378: I am in favor of comments, where they give additional context and reasoning (and where they do not transliterate the code itself into prose).

signedav: The reason for this comment is, because it's quite unusual at this place, that we don't read the configuration but set it just to false. So it's here to say: Hey, this is done on purpose to have the box always unchecked. 

I don't mind to remove it - it's not that important. I will remove when you tell me to without any resistance, just please accept that it's not "transliteration of the code itself into prose" :slightly_smiling_face:

m-kuhn: I was more hoping for a minimal bit of reasoning "set ... always to unchecked because [xxx]"
I was pretty sure that you had a good reason ;)

signedav: "... because otherwise the user could delete the data accidentally" - okay?

m-kuhn: much better :+1:

PR #417, opened 2020-07-01
Keep QGIS enabled on generate, import and export dialog open

signedav:
By using `show()` instead of `exec()`.

While the dialog is open the action toolbutton is checked.
modelbaker2
(in the solution the newly opened dialogs are always on top - this screencast is obsolete)

Another possibility would be to have it like `tool_button.setDown()`, but then it's still enabled in the database menu.

This pretty sure fixes as well #314 because on Windows on focusing on QGIS the dialogs are raised.

### DIFF
--- a/QgisModelBaker/qgismodelbaker.py
+++ b/QgisModelBaker/qgismodelbaker.py
@@ -44,6 +44,11 @@ class QgisModelBakerPlugin(QObject):
         QObject.__init__(self)
         self.iface = iface
         self.plugin_dir = os.path.dirname(__file__)
+
+        self.generate_dlg = None
+        self.export_dlg = None
+        self.importdata_dlg = None
+
         self.__generate_action = None
         self.__export_action = None
         self.__importdata_action = None
@@ -51,6 +56,7 @@ class QgisModelBakerPlugin(QObject):
         self.__help_action = None
         self.__about_action = None
         self.__separator = None
+
         if locale.getlocale() == (None, None):
             locale.setlocale(locale.LC_ALL, '')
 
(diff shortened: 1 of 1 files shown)


m-kuhn:
Can we have it activated instead of disabled, like the snapping button?
(`setCheckable(True)`, `setChecked(active)`)

signedav:
Ah, this is the way I've been looking for. Thanks. (Missed the setCheckable and thought it's only possible to set the toolbutton down)...

signedav:
I added the possiblity to close the window by pressing the unchecked toolbutton (see screencast above).

QgisModelBaker/qgismodelbaker.py
```diff
@@ -150,37 +155,51 @@ def unload(self):
...
-        self.generate_dlg = GenerateProjectDialog(self.iface, self.ili2db_configuration)
-        self.generate_dlg.setAttribute(Qt.WA_DeleteOnClose)
-        self.generate_dlg.setWindowFlags(self.generate_dlg.windowFlags() | Qt.Tool)
-        self.generate_dlg.show()
-        self.generate_dlg.finished.connect(self.generate_dialog_finished)
-        self.__generate_action.setEnabled(False)
+        if self.generate_dlg:
+            self.generate_dlg.reject()
+        else:
+            print('generate dialog')
```
m-kuhn, line 161: print?

QgisModelBaker/qgismodelbaker.py
```diff
@@ -150,37 +155,51 @@ def unload(self):
...
 
     def show_generate_dialog(self):
-        self.generate_dlg = GenerateProjectDialog(self.iface, self.ili2db_configuration)
-        self.generate_dlg.setAttribute(Qt.WA_DeleteOnClose)
-        self.generate_dlg.setWindowFlags(self.generate_dlg.windowFlags() | Qt.Tool)
-        self.generate_dlg.show()
-        self.generate_dlg.finished.connect(self.generate_dialog_finished)
-        self.__generate_action.setEnabled(False)
+        if self.generate_dlg:
+            self.generate_dlg.reject()
```
m-kuhn, line 159: when does this happen? and is rejecting the right decision at this time?

signedav: On clicking on the checked icon (see the screencast). I think yes reject. But I think it does not make a big difference to accept.

QgisModelBaker/qgismodelbaker.py
signedav: Oops.

QgisModelBaker/qgismodelbaker.py
m-kuhn: I understand. I was thinking that we just don't react (or raise it to the foreground) since it's more a "wizard" dialog than a dock one toggles on and off, but this also makes sense.

PR #419, opened 2020-07-14
Fix mandatory fields

yesidpol:
This PR allows specifying to widgets which fields are mandatories. Qgis requires that mandatory fields be set explicitly with some DB engines like SQL Server.

For example:
**SQL Server Table**
```sql
CREATE TABLE test.table_a(
	id int IDENTITY NOT NULL primary key,
	number_a  int NOT NULL,
	description_a varchar(256) NOT NULL,
	date_a datetime NOT NULL);
```
**Form with the changes**

**Currently**

### DIFF
--- a/QgisModelBaker/libqgsprojectgen/dataobjects/fields.py
+++ b/QgisModelBaker/libqgsprojectgen/dataobjects/fields.py
@@ -22,6 +22,7 @@ from qgis.core import (
     QgsEditorWidgetSetup,
     QgsDefaultValue
 )
+from qgis.core import QgsFieldConstraints
 
 
 class Field:
@@ -32,6 +33,7 @@ class Field:
         self.widget = None
         self.widget_config = dict()
         self.default_value_expression = None
+        self.not_null = None
 
     def dump(self):
         definition = dict()
@@ -57,3 +59,7 @@ class Field:
         if self.default_value_expression:
             default_value = QgsDefaultValue(self.default_value_expression)
             layer.layer.setDefaultValueDefinition(field_idx, default_value)
+
+        # Additionally, some DB engines like MSSQL require NotNull constraints being set explicitly
+        if self.not_null:
+            layer.layer.setFieldConstraint(field_idx, QgsFieldConstraints.ConstraintNotNull)
--- a/QgisModelBaker/libqgsprojectgen/db_factory/db_factory.py
+++ b/QgisModelBaker/libqgsprojectgen/db_factory/db_factory.py
@@ -123,7 +123,7 @@ class DbFactory(ABC):
 
         return messages
 
-    def customize_widget_editor(self, field: Field, data_type: str):
+    def customize_widget_editor(self, field: Field, data_type: str, fielddef: dict):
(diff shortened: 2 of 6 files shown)


QgisModelBaker/libqgsprojectgen/dataobjects/fields.py
```diff
@@ -32,6 +33,7 @@ def __init__(self, name):
         self.widget = None
         self.widget_config = dict()
         self.default_value_expression = None
+        self.mandatory = None
```
m-kuhn, line 36: Can we call this `not_null` to be consistent in terminology?

QgisModelBaker/libqgsprojectgen/db_factory/db_factory.py
```diff
@@ -123,14 +123,15 @@ def get_specific_messages(self):
...
-    def customize_widget_editor(self, field: Field, data_type: str):
+    def customize_widget_editor(self, field: Field, data_type: str, fielddef: dict):
         """Allows customizing the way a field is shown in the widget editor.
 
         For instance, a boolean field can be shown as a checkbox.
 
         :param field: The field that will be customized
         :type field: :class:`Field`
         :param data_type: The type of field
+        :param fielddef: The field info from database
```
m-kuhn, line 134: Can you add more information about the expected/supported content of the dict?
Or possibly add a `class FieldDef` instead of the dict which defines the supported properties?

QgisModelBaker/libqgsprojectgen/dataobjects/fields.py
yesidpol: Done!

yesidpol:
> Thank you, some comments inside.
> 
> Am I correct, that this could also be done in QGIS directly (improve detection of constraints on mssql columns so they work like postgres detection works)?

Yes, you are correct!

QgisModelBaker/libqgsprojectgen/db_factory/db_factory.py
yesidpol: Done!

signedav:
@yesidpol What is the status with this PR? Anymore requested changes @m-kuhn ?

gacarrillor:
@signedav As @m-kuhn suggests, this is actually a missing part in QGIS Core. 
This can be a requirement from us for Opengis.ch. By and large, it's better if you can implement it there.

m-kuhn:
@gacarrillor now that this is implemented in QGIS, is there something left on the todo list for thie pull request?

gacarrillor:
> @gacarrillor now that this is implemented in QGIS, is there something left on the todo list for thie pull request?

The only item left in the todo list was for @yesidpol to close this PR after he confirmed everything was working as expected. I expect this PR to be closed (or a new comment from our side) this Friday at most.

m-kuhn:
Thanks for the feedback @gacarrillor and for testing @yesidpol

gacarrillor:
> Thanks for the feedback @gacarrillor and for testing @yesidpol

Thanks to you for the upstream implementation!

PR #421, opened 2020-07-17
Omit unused tables

### LINKED ISSUE #284: Omit tables and fields that do not need to be manipulated (localised/multilingual text)
There are several cases where ili2pg produces database tables (or fields within tables) that do not need to be manipulated at all. To see those tables and attributes in a QGIS project might be confusing.
1)
**Such tables shall be ignored by the ModelBaker. 
E.g. «multilingualtext» (multilingual texts are implemented directly within the respective object tables.**
2)
Such fields shall be ignored by the ModelBaker. 
E.g. A the actual field name from a multilingual text (there are fields added for every language in the case of the usage of LocalisationCH_V1). 
E.g. B the field name «multilingualtext_localisedtext» in the table «localisedtext».
3)
In the table «localisedtext», the field «t_type» gets a constant value of «localisationch_v1_localisedtext» (where LocalisationCH_V1 is implemented (see also Localisation_V1).

### LINKED ISSUE #409: Skip loading of geometry classes (GeometryCHLV95_V1) as layers
Geometry classes derived from GeometryCHLV95_V1 like

- linestructure
- multiline
- multisurface
- surfacestructure

shouldn't be loaded into the project as layers

Tested with loading from Postgresql (data structure manually created with ili2pg)

image

Testmodel: http://geo.so.ch/models/ALW/SO_ALW_Strukturverbesserungen_20190912.ili

signedav:
According the ili2db documentation for the following tables an alternate creation is used.

_Strukturen mit dem Interlis Metaattribut ili2db.mapping=MultiSurface
Strukturen mit dem Interlis Metaattribut ili2db.mapping=MultiLine
Strukturen mit dem Interlis Metaattribut ili2db.mapping=MultiPoint
GeometryCHLV03_V1.MultiSurface
GeometryCHLV03_V1.MultiLine
GeometryCHLV03_V1.MultiDirectedLine
GeometryCHLV95_V1.MultiSurface
GeometryCHLV95_V1.MultiLine
GeometryCHLV95_V1.MultiDirectedLine
CatalogueObjects_V1.Catalogues.CatalogueReference
CatalogueObjects_V1.Catalogues.MandatoryCatalogueReference
LocalisationCH_V1.MultilingualMText
LocalisationCH_V1.MultilingualText
LocalisationCH_V1.LocalisedMText
LocalisationCH_V1.LocalisedText_

This means that no layers for mulitgeometry needs to be created in the project, since the don't need to be changed at all.

We get these tables like this:
If there are `ilielements`  in the `t_ili2db_meta_attrs` table with `attr_name` **_ili2db.mapping_** or has one of the listed names above, then check if it's existing the table t_ili2db_classname`. If so, we ignore it in the creation of the projects layers. Additionally we don't create any relations which are depending on such a layer.

fixes #284 fixes #409


QgisModelBaker/libqgsprojectgen/dbconnector/gpkg_connector.py
```diff
@@ -172,6 +172,19 @@ def _get_tables_info(self):
         cursor.close()
         return complete_records
 
+    def get_meta_attrs_info(self):
```
m-kuhn, line 175: Can you add this method to the base class with a documentation comment?
The content of the class can be `raise NotImplementedError()`

signedav:
@m-kuhn I'm loosing time here. Do you have an idea why the tests are failing? They fail locally as well...

m-kuhn:
https://stackoverflow.com/a/19527487/2319028

QgisModelBaker/libqgsprojectgen/generator/generator.py
```diff
@@ -77,7 +77,42 @@ def append_print_message(self, level, text):
         if message not in self.collected_print_messages:
             self.collected_print_messages.append(message)
 
+    def omitted_layers(self):
```
m-kuhn, line 80: Would it be possible to

 - move this to the connector
 - combine this with `IGNORED_TABLES` -- ideally, this would be made more specific for each provider (e.g. right now it's also listing gpkg specific tables on postgres)

m-kuhn: For consistency I would also like to use either `omitted` or `ignored` as term for this.

signedav: I can include the `IGNORED_TABLES` into the omit tables on creating the layers. But the ignored tables are used as well in `get_tables_info_without_ignored_tables` and I'm not sure if the omitted tables should be there as well. Btw. `get_tables_info_without_ignored_tables` is never called. You know anything about it? Can it be removed?

signedav: But anyway shouldn't we keep the `IGNORED_TABLES` to have the "lists" separated. Just append them in the `def omitted_layers()` to the list of tables to have only one check in creating the layers.

m-kuhn: I don't remember. Can you look at the `git blame` for `get_tables_info_without_ignored_tables`?
It's ok to keep the lists separated (global static, provider specific static, dynamic) and then use them at the same time. What do you think?

signedav: Yes. That's what I prefer.

signedav: I did it but do we really want this stuff in the connector?

QgisModelBaker/libqgsprojectgen/dbconnector/db_connector.py
```diff
@@ -172,6 +173,40 @@ def get_bags_of_info(self):
         '''
         return []
 
+    def get_omitted_layers(self):
```
m-kuhn, line 176:
```suggestion
    def get_ignored_layers(self):
```

Can we agree on one taxonomy?

signedav: I can do that.

QgisModelBaker/libqgsprojectgen/generator/generator.py
signedav: Hm.. I'll have to import the config from the generator to the dbconnector or I need to create a dbconnector config file containing the IGNORED_TABLES. These are defined together with the IGNORED_SCHEMAS etc. so I would like to not split them.

Shouldn't we just handle it in the generator like suggested in the beginning to keep the db_connector clean @m-kuhn ?

m-kuhn: I think it makes sense in here because of provider specific logic. In here we can add custom ignored tables on a provider (connector) basis in subclasses which we cannot easily do without `if connector_pg elif connector_gpkg ...` etc code in the generator. Ignored_schemas are a very good candidate for moving here too because they are very provider specific.
Does that make sense?

signedav: Okay, moving everything is fine as well for me.

signedav:
@m-kuhn Can you have a look and tell me if this is what you meant? I seperated the gpkg tables and the pg / mssql tables but I'm not sure. Where can I check what tables are used for what type of DB? Are they the same for postgres and mssql?

And IGNORED_FIELDNAMES are still handled in the generator - should I move them to the connectors as well?

Thanks a lot.

QgisModelBaker/libqgsprojectgen/dbconnector/config.py
```diff
@@ -0,0 +1,73 @@
+IGNORED_SCHEMAS = [
```
m-kuhn, line 1:
```suggestion
IGNORED_PG_SCHEMAS = [
```

signedav: They are used in the MSSQL implementations as well. I guess they have the same names, no?

signedav:
I merged together the list of ignored tables since I don't know which table name is concerning which database (besides the ones with GPKG in the name). If there is this information easily accessible, then please tell me. Otherwise I wonder if it's worth the effort to find it out.

QgisModelBaker/libqgsprojectgen/dbconnector/config.py
m-kuhn: pg_catalog is postgres specific, information_schema maybe not. I'd add them to both.

signedav:
Merge @m-kuhn ?

QgisModelBaker/libqgsprojectgen/dbconnector/pg_connector.py
```diff
@@ -412,6 +428,36 @@ def get_bags_of_info(self):
             return cur
         return []
 
+    def get_ignored_layers(self):
```
gacarrillor, line 431: This logic deserves to be in the parent class (DBConnector).

m-kuhn:
I think @gacarrillor 's review of this is more than enough, I'll retire for this PR and leave all powers to the two of you. Thanks!

m-kuhn (approved):
Good for me if good for @gacarrillor

QgisModelBaker/libqgsprojectgen/dbconnector/pg_connector.py
signedav: I will do that.

signedav: Done

gacarrillor (approved):

gacarrillor:
@signedav Could we proceed and merge your 2 open PRs?
We plan to create some PRs and would like to do it using a complete master.

signedav:
I had to make a last fix. Now it seems to work. So I'll merge it. Thanks @gacarrillor

QgisModelBaker/libqgsprojectgen/dbconnector/db_connector.py
```diff
@@ -159,6 +175,42 @@ def get_bags_of_info(self):
...
+                    tables.append(record['tablename'])
+                    continue
+            if 'tablename' in record:
+                if record['tablename'] in IGNORED_TABLES:
+                    tables.append(record['tablename'])
+                    continue
+        # get the referencing tables
+        for record in relations_info:
+            if record['referenced_table'] in tables:
+                referencing_tables.append(record['referencing_table'])
```
gacarrillor, line 210: This `if` block is damaging our projects.

If we have basket cols in our DB, we end up with relations like this:

`{'referencing_table': 'terreno', 'referencing_column': 'T_basket', 'referenced_table': 'T_ILI2DB_BASKET', 'referenced_column': 'T_Id', 'constraint_name': 'terreno_T_basket_T_ILI2DB_BASKET_T_Id', 'strength': ''}`

Which then enter into such `if` block, and then ALL (yes, all! :smile:) our classes that are not domains, are added to the referencing_tables list, and therefore, are ignored.

Please give such if clause a second thought.

signedav: Is `T_ILI2DB_BASKET` table contained in the `t_ili2db_meta_attrs` as `mapping`?

So  shouldn't the referencing tables be ignored at all? Or what is the criteria to ignore it or not?

gacarrillor: > Is T_ILI2DB_BASKET table contained in the t_ili2db_meta_attrs as mapping?

Nope, it doesn't appear there.

`T_ILI2DB_BASKET`is in the `DBConnector.config.IGNORED_TABLES` list.

The criterion here is to avoid loading such table (this works fine at the moment), but all tables that are referencing to `T_ILI2DB_BASKET` should still be loaded.

gacarrillor: What I see is that the `if` condition is too broad, as it also takes into account the `DBConnector.config.IGNORED_TABLES`. 

From what I understand, you want to filter out tables contained in the `t_ili2db_meta_attrs` as `mapping` and their referencing tables. Right?

signedav: Ah I see. I did it on purpose. But obviously wrong. So I'll change that. Thanks.

PR #423, opened 2020-07-27
Use ili_name part as table alias

### LINKED ISSUE #48: Table name from iliname
ili2pg writes a comment `@iliname Table.Name` to the database schema. We could use that for the table name in QGIS. Only the last part (`Name` for the example before) as default table name.
Just like for field aliases done already.

signedav:
If no table alias defined in the interlis file it takes the last part of the ili name as table alias.
      
On the table `landcover` with ili name `RoadsSimple.Roads.LandCover` it uses `LandCover` as the table alias.

If the table is a domain table, it takes the two last parts instead of only the last one (since it needs it's parent table to be unique). So the `Type` of `LandCover` ( `RoadsSimple.Roads.LandCover.Type` ) receives the alias `LandCover_Type`.

 fixes #48


QgisModelBaker/libqgsprojectgen/generator/generator.py
```diff
@@ -101,12 +101,19 @@ def layers(self, filter_layer_list=[]):
             if filter_layer_list and record['tablename'] not in filter_layer_list:
                 continue
 
-            alias = record['table_alias'] if 'table_alias' in record else ''
             is_domain = record['kind_settings'] == 'ENUM' or record[
                 'kind_settings'] == 'CATALOGUE' if 'kind_settings' in record else False
             is_structure = record['kind_settings'] == 'STRUCTURE' if 'kind_settings' in record else False
             is_nmrel = record['kind_settings'] == 'ASSOCIATION' if 'kind_settings' in record else False
 
+            alias = record['table_alias'] if 'table_alias' in record else ''
```
m-kuhn, line 109:
```suggestion
            alias = record['table_alias'] if 'table_alias' in record else None
```

Maybe? I always tend to prefer explicit "no information" over an empty string -- except where not possible.

m-kuhn: Actually, wouldn't that be much more readable?

```suggestion
            if 'table_alias' in record:
                alias = record['table_alias']
            else:
```

signedav: On the second one I disagree. On such small if-clause I think even if the if-clause itself is slightly more readable, it goes on costs of the readability of the whole function, since it makes it bigger with a lot of simple ifs.

m-kuhn: Might just be my general aversion against ternary operators :-)

signedav: Yeah, I generally like the if then else way better as well. But for this case it's okay.

QgisModelBaker/libqgsprojectgen/generator/generator.py
```diff
@@ -101,12 +101,19 @@ def layers(self, filter_layer_list=[]):
...
-            alias = record['table_alias'] if 'table_alias' in record else ''
             is_domain = record['kind_settings'] == 'ENUM' or record[
                 'kind_settings'] == 'CATALOGUE' if 'kind_settings' in record else False
             is_structure = record['kind_settings'] == 'STRUCTURE' if 'kind_settings' in record else False
             is_nmrel = record['kind_settings'] == 'ASSOCIATION' if 'kind_settings' in record else False
 
+            alias = record['table_alias'] if 'table_alias' in record else None
+            if not alias:
+                if is_domain:
+                    short_name = record['ili_name'].split('.')[-2] + '_' + record['ili_name'].split('.')[-1] if 'ili_name' in record else ''
```
m-kuhn, line 112: Are we sure they always split into enough parts (and don't suddenly get an array out of bounds exception here?

signedav: You mean in case the ili_name structure changes? What do you suggest? Make a fallback in case of exception?

m-kuhn: I'm not absolutely sure we always do have a dot in each name - maybe you know better? - so if we happen to encounter an ili_name `NAME_WITHOUT_DOT` for whatever reason the user will get nothing but an exception which is quite uncool. So unless you are 100% sure, I'd just do the fallback :+1:

QgisModelBaker/libqgsprojectgen/generator/generator.py
```diff
@@ -101,12 +101,19 @@ def layers(self, filter_layer_list=[]):
...
 
-            alias = record['table_alias'] if 'table_alias' in record else ''
             is_domain = record['kind_settings'] == 'ENUM' or record[
                 'kind_settings'] == 'CATALOGUE' if 'kind_settings' in record else False
             is_structure = record['kind_settings'] == 'STRUCTURE' if 'kind_settings' in record else False
             is_nmrel = record['kind_settings'] == 'ASSOCIATION' if 'kind_settings' in record else False
 
+            alias = record['table_alias'] if 'table_alias' in record else None
+            if not alias:
+                if is_domain:
```
gacarrillor, line 111: I'm not sure of this approach. Take for example this ili_name from one of our domains (obtained from `t_ili2db_classname`):

`Modelo_Aplicacion_LADMCOL_Lev_Cat_V1_0.LC_PuntoTipo`

Where: 

`Modelo_Aplicacion_LADMCOL_Lev_Cat_V1_0`: Model name
`LC_PuntoTipo`: Domain table name

Taking the last part of the split array could be the way to go here too.

Perhaps you're referring to this ili_name (obtained from `t_ili2db_attrname`)?
`Modelo_Aplicacion_LADMCOL_Lev_Cat_V1_0.Levantamiento_Catastral.LC_PuntoControl.PuntoTipo`

Which is the iliname of the attribute that has been defined to store domain values.

signedav: Hm... okay. I assume the sql name of the created table is lc_puntotipo then?

On road simple is that the domain table containing the enumerations is called landcover_type,  because it containes LandCover.Type. So I thought this is the way it works. 

Can you provide me the model so I can see if I find a better solution?

gacarrillor: > Hm... okay. I assume the sql name of the created table is lc_puntotipo then?

Right.

> Can you provide me the model so I can see if I find a better solution?

Sure, you can find it here: https://github.com/SwissTierrasColombia/LADM_COL/releases/download/1.0/Modelo_Aplicacion_LADMCOL_Levantamiento_Catastral_V1_0.zip

The model is called `Modelo_Aplicacion_LADMCOL_Lev_Cat_V1_0` and an example of a domain is: `Modelo_Aplicacion_LADMCOL_Lev_Cat_V1_0.LC_PuntoTipo`.

gacarrillor: Where can I find the Swiss model you're using? It would be good to see if there are differences in the modelling itself.

signedav: https://www.interlis.ch/download/interlis2/Testbeispiel_20160812.zip <- this here.

And we are using it in the tests for Model Baker as well.

signedav: There is sometimes a `Topic` that it integrated into the ili-name but sometimes there isn't.

m-kuhn: Is it possible that there are duplicate names and it adds the Topic when there is a name collision without the topic?

signedav: In the case of the simple road model there would be duplicates if there are no topics. But I don't know if it does not take the topic if there are no duplicates. I would need to check.

signedav: Sorry what I said here https://github.com/opengisch/QgisModelBaker/pull/423#discussion_r478205903 is wrong: There are no duplicates without the topics. The topics are not integrated into to sql-name.

When the iliname is:
`<MODEL>.<TOPIC>.<CLASS>` it creates the sqlname `<CLASS>`
When the iliname is:
`<MODEL>.<CLASS>` it creates the sqlname `<CLASS>`

But I think I compared not the same things above:

While `Modelo_Aplicacion_LADMCOL_Lev_Cat_V1_0.LC_PuntoTipo` is under `DOMAIN` in the ili file of Germán, the  `RoadsSimple.Roads.LandCover.Type` is defined as this:
```
    CLASS LandCover =
      Type: MANDATORY (
        building,
        street,
        water,
        other);
      Geometry: MANDATORY SURFACE WITH (STRAIGHTS)
        VERTEX Point2D WITHOUT OVERLAPS > 0.100;
    END LandCover;
```
So we have to find out where to get the difference from then...

m-kuhn: What I was referring to is https://github.com/claeis/ili2db/blob/master/docs/ili2db.rst

> Option | Beschreibung
> -- | --
> `--disableNameOptimization` | Schaltet die Nutzung von unqualifizierten Klassennamen aus. Für alle  Tabellennamen werden qualifizierte Interlis-Klassennamen  (Model.Topic.Class) verwendet (und in einen gültigen Tabellennamen  abgebildet). (siehe Kapitel Abbildungsregeln/Namenskonventionen)
> `--nameByTopic` | Für alle Tabellennamen werden teilweise qualifizierte  Interlis-Klassennamen (Topic.Class) verwendet (und in einen gültigen  Tabellennamen abgebildet). (siehe Kapitel  Abbildungsregeln/Namenskonventionen)

We have some control over the sql name generation. But it's probably not related to this discussion.

gacarrillor: > While Modelo_Aplicacion_LADMCOL_Lev_Cat_V1_0.LC_PuntoTipo is under DOMAIN in the ili file of Germán, the RoadsSimple.Roads.LandCover.Type is defined as this:

It's an in-line ENUM. We're moving forward.

Proposal:

1. If it's a class and does not have alias: Take the last part from its iliname. 
2. If it's a domain and does not have alias:
   2.a. Check if its `iliname` is also an iliname in `t_ili2db_attrname`, which indicates we have an in-line ENUM. If so, to obtain a proper alias, take the last 2 parts of the iliname, and join them with an underscore (e.g., `LandCover_Type`).
   2.b. If `iliname` is not in `t_ili2db_attrname`, to get a proper alias take the last part of `iliname`, as in the table case (e.g., `LC_PuntoTipo`).

Note: 2.a. is ugly, I know. Perhaps we could consider whether it makes sense to ask Claude to make a difference between in-line ENUMs and non-in-line ENUMs in the `t_ili2db_table_prop` table.

signedav: Just for information regarding https://github.com/opengisch/QgisModelBaker/pull/423#discussion_r478579452 of @m-kuhn : In this case we would have roadssimpleroads_landcover on `--disableNameOptimization` and roads_landcover on `--nameByTopic`

@gacarrillor I don't think it's that ugly. It checks if it's an ENUM and if this ENUM is an attribute (in-line) then it needs to contain the name of the owner table in the table name. I think it's a straight forward solution.

gacarrillor (approved):

PR #425, opened 2020-08-31
Raise warning when schema/db-file exists

### LINKED ISSUE #365: Warning if schema exists
Model Baker should show a warning if a schema already exists.
Most likely it's an error since users don't normally generate multiple models in a single schema.

signedav:
On Schema Import 

image

this resolves #365


QgisModelBaker/gui/generate_project.py
```diff
@@ -190,10 +191,37 @@ def accepted(self):
...
+                    warning_box.setText(self.tr("{db_or_schema}: {db_or_schema_name}\n\nDo you want to "
+                                                "import anyway?").format(
+                        db_or_schema=db_factory.get_specific_messages()['db_or_schema'].capitalize(),
+                        db_or_schema_name=configuration.dbschema or config_manager.get_uri()
+                    ))
+                    warning_box.setStandardButtons(QMessageBox.Yes | QMessageBox.No)
+                    warning_box_result = warning_box.exec_()
+                    if warning_box_result == QMessageBox.No:
+                        return
+        except FileNotFoundError:
```
m-kuhn, line 219: Can we deal with this in the gpkg connector?

QgisModelBaker/gui/generate_project.py
```diff
@@ -190,10 +191,37 @@ def accepted(self):
...
+
+            if db_connector.db_or_schema_exists():
+                if interlis_mode:
+                    warning_box = QMessageBox(self)
+                    warning_box.setIcon(QMessageBox.Warning)
+                    warning_title = self.tr("{} already exists").format(
+                        db_factory.get_specific_messages()['db_or_schema']
+                    ).capitalize()
+                    warning_box.setWindowTitle(warning_title)
+                    warning_box.setText(self.tr("{db_or_schema}: {db_or_schema_name}\n\nDo you want to "
```
m-kuhn, line 210:
```suggestion
                    warning_box.setText(self.tr("{db_or_schema}: {db_or_schema_name} already exists.\n\nDo you want to "
```

QgisModelBaker/gui/generate_project.py
```diff
@@ -190,10 +191,37 @@ def accepted(self):
...
+        try:
+            # raise warning when the schema or the database file already exists
+            config_manager = db_factory.get_db_command_config_manager(configuration)
+            db_connector = db_factory.get_db_connector(
+                config_manager.get_uri(configuration.db_use_super_login) or config_manager.get_uri(), configuration.dbschema)
+
+            if db_connector.db_or_schema_exists():
+                if interlis_mode:
+                    warning_box = QMessageBox(self)
+                    warning_box.setIcon(QMessageBox.Warning)
```
m-kuhn, line 205: Is the "warning" a read circle on your computer? I wonder if we should add this more as an information than a warning... It's just something to be aware of, not something that's bad per se.

QgisModelBaker/gui/generate_project.py
signedav: I would prefer: 
```
{db_or_schema} already exists: 
{db_or_schema_name}

Do you want to import anyway?
```
Because {db_or_schema_name} can be quite long when it's a path to the gpkg.
But then it's the same like the title on the top line, and that's not so nice.

QgisModelBaker/gui/generate_project.py
signedav: It's the exclamation mark sign you see above. We use the same when the xtf already exists on export. For me an information would be okay as well, but then I would make the "xtf already exists" an information as well. What do you think?

QgisModelBaker/gui/generate_project.py
signedav: In the GPKGConnectior it raises this `FileNotFoundError`. So everywhere on calling `get_db_connector` it catches this error.
But we could check for the existence of the file in the `gpkg_factory.get_db_connector` and if it does not exist return `None` - so in the GUI classes we check if db_connector is None. What do you think?

QgisModelBaker/gui/generate_project.py
m-kuhn: I would say, the xtf *always* overwrites information. The import *could* overwrite information or just add more information to the existing data.

signedav: Okay then.

QgisModelBaker/gui/generate_project.py
m-kuhn: Agreed on the long filename.

How the title is displayed depends on the platform, some systems may decide to show the title bar somewhere else than directly around the message box. I would put all the information into the dialog, even if it's duplicate.

QgisModelBaker/gui/generate_project.py
signedav: But anyway. If it would be here it should be `except (DBConnectorError, FileNotFoundError):` since the `FileNotFoundError` is the equivalent of `DBConnectorError`. But anyway. Would the `x_factory.get_db_connector` be the preferred place?

m-kuhn: The reason why I asked was, if there is a `FileNotFoundError` we expect, should we also expect a `NetworkConnectionFailedError`? And a `SchemaDoesNotExistError`?

But if it's already done like this everywhere, no objections to also do it here.

signedav: Okay. But thanks for mentioning. Because of your comment I noticed that I forget to catch `DBConnectorError` :-)

m-kuhn: What my main question is, which errors is this method expected to raise?
If we add more connectors in the future (oracle, ...) do we need to add more catches?

signedav: These system dependent errors are catched in the connector and then the `DBConnectorError` is raised - so from the outside only the `DBConnectorError` needs to be catched (or the `FileNotFoundError`).

QgisModelBaker/gui/generate_project.py
```diff
@@ -190,10 +191,38 @@ def accepted(self):
...
+            if db_connector.db_or_schema_exists():
+                if interlis_mode:
+                    warning_box = QMessageBox(self)
+                    warning_box.setIcon(QMessageBox.Information)
+                    warning_title = self.tr("{} already exists").format(
+                        db_factory.get_specific_messages()['db_or_schema']
+                    ).capitalize()
+                    warning_box.setWindowTitle(warning_title)
+                    warning_box.setText(self.tr("{warning_title}:\n{db_or_schema_name}\n\nDo you want to "
+                                                "import anyway?").format(
```
m-kuhn, line 211:
```suggestion
                                                "import into the existing schema?").format(
```

m-kuhn: I think that would be even nicer to ask :-)

m-kuhn (approved):

signedav:
Thanks @m-kuhn

PR #426, opened 2020-08-31
use version > 2.8.3 for psycopg from python version 3.8

signedav:
https://github.com/psycopg/psycopg2/issues/854#issuecomment-611791946

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,4 @@
-psycopg2==2.7.4; python_version < '3.8'
-psycopg2==2.8.5; python_version >= '3.8'
+psycopg2>=2.7.4
 pytest
 pyodbc
 pyyaml
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
-psycopg2==2.7.4
+psycopg2==2.7.4; python_version < '3.8'
+psycopg2==2.8.5; python_version >= '3.8'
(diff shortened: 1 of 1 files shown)


signedav:
Or could we just remove the version, then it installs the latest? 

Why we defined the version on psycopg at all?

m-kuhn:
Clicking through git blame: https://github.com/opengisch/QgisModelBaker/pull/180#pullrequestreview-113083161

So `psycopg2>=2.7.4` should be the way to go

signedav:
> Clicking through git blame: [#180 (review)](https://github.com/opengisch/QgisModelBaker/pull/180#pullrequestreview-113083161)
> 
> So `psycopg2>=2.7.4` should be the way to go

But we need >=2.8.3 with python 3.8 ... can we do that or might there be a problem, with environments having python versions < 3.8?

signedav:
Or does it install with your suggestion the latest version anyway when using python 3.8?

m-kuhn:
> Or does it install with your suggestion the latest version anyway when using python 3.8?

if it doesn't let's stick to your solution but with a `>=` in the psycopg version as well

signedav:
I can do a test run.

signedav:
And it works :taco: Thanks @m-kuhn I approve your suggestion :-)

PR #429, opened 2020-09-03
Import schema without constraints

### LINKED ISSUE #427: Import schema without validation
This has been discussed here #318 and should be a follow up of #424

Import without validation option in schema import:
- Remove `--createNumChecks` (schemaimport)
- Specify `--sqlEnableNull` (schemaimport)
- Remove `--createFkIdx` (or `--createFk`?) (schemaimport)
- Remove `--createUnique` (schemaimport)

signedav:
Like the possibility to import and export without validation, we have now the possibility to import a **schema** without constraints.

image

If this is started, the ili2db command is run with this parameter:
- `--sqlEnableNull`
and without this parameters
- `--createNumChecks`
- `--createUnique`
- `--createFk`

`--createFkIdx` is still used, since it has nothing to do directly with constraints (it just indexes the column for the fks AFAIK)

resolves #427

Since on importing an xtf File the schema is created (if not existing) the import schema parameters are taken there as well. On "Import without validation" these constraint-parameters are not set.

### DIFF
--- a/QgisModelBaker/libili2db/ili2dbconfig.py
+++ b/QgisModelBaker/libili2db/ili2dbconfig.py
@@ -202,15 +202,15 @@ class SchemaImportConfiguration(Ili2DbCommandConfiguration):
         args += ["--coalesceCatalogueRef"]
         args += ["--createEnumTabs"]
 
-        if not self.disable_constraint_parameters:
-            args += ["--createNumChecks"]
-            args += ["--createFk"]
-            args += ["--createFkIdx"]
-            args += ["--createUnique"]
-
         if self.disable_constraint_parameters:
             args += ["--sqlEnableNull"]
 
+        if not self.disable_constraint_parameters:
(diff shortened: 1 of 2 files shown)


QgisModelBaker/libili2db/ili2dbconfig.py
```diff
@@ -200,16 +200,22 @@ def to_ili2db_args(self, extra_args=[], with_action=True):
 
         args += ["--coalesceCatalogueRef"]
         args += ["--createEnumTabs"]
-        args += ["--createNumChecks"]
+
+        if self.disable_validation:
+            args += ["--sqlEnableNull"]
+
+        if not self.disable_validation:
```
m-kuhn, line 207:
```suggestion
        else:
```

Just a proof that I have actually done a review ;-)

m-kuhn (approved):

PR #434, opened 2020-09-30
Support for scientific notation in GeoPackage numeric ranges

gacarrillor:
And tests for all DB engines.

### DIFF
--- a/QgisModelBaker/tests/test_import.py
+++ b/QgisModelBaker/tests/test_import.py
@@ -85,7 +85,6 @@ class TestImport(unittest.TestCase):
             """.format(importer.configuration.dbschema))
         record = next(cursor)
         self.assertIsNotNone(record)
-        self.assertEqual(record[0], 6)  # t_id for 'Unidad_Derecho'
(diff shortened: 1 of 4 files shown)


QgisModelBaker/tests/test_import.py
```diff
@@ -85,7 +85,6 @@ def test_import_postgis(self):
             """.format(importer.configuration.dbschema))
         record = next(cursor)
         self.assertIsNotNone(record)
-        self.assertEqual(record[0], 6)  # t_id for 'Unidad_Derecho'
         self.assertEqual(record[1], 'POLYGON((1000257.426 1002020.376,1000437.688 1002196.495,1000275.472 1002428.19,1000072.25 1002291.539,1000158.572 1002164.914,1000159.942 1002163.128,1000257.426 1002020.376))')
```
m-kuhn, line 88: I assume these are removed on purpose?

m-kuhn (approved):

QgisModelBaker/tests/test_import.py
gacarrillor: Yes, the second commit does it. We talked about it before. These `t_id` checks are a nightmare because the `t_id` is not guaranteed to be always the same.

PR #436, opened 2020-10-05
Update to ili2db 4 4 4

gacarrillor:
1. Update to ili2db 4.4.4.
 2. Fix a (Colombian) test model that was invalid, and due to improvements in ili2c, now we know it.
 3. Update `CHBase_Part4_ADMINISTRATIVEUNITS_20110830.ili` to the latest version from http://models.geo.admin.ch/CH, which fixes the model. With the update to ili2db 4.4.4, ili2c says the current model in master is invalid.


QgisModelBaker/tests/testdata/ilimodels/CHBase_Part4_ADMINISTRATIVEUNITS_20110830.ili
```diff
@@ -39,10 +40,10 @@ TYPE MODEL CHAdminCodes_V1 (en)
 END CHAdminCodes_V1.
 
 !! ########################################################################
-!!@technicalContact=models@geo.admin.ch
-!!@furtherInformation=http://www.geo.admin.ch/internet/geoportal/de/home/topics/geobasedata/models.html
+!!@technicalContact=mailto:models@geo.admin.ch
```
signedav, line 43: What is the reason for these corrections? 
Would it fail with the old file?

signedav (approved):
But all in all it looks good. Thanks.

QgisModelBaker/tests/testdata/ilimodels/CHBase_Part4_ADMINISTRATIVEUNITS_20110830.ili
gacarrillor: > What is the reason for these corrections?

The official model repository made such adjustments to the model. Probably they realized their model was invalid after an ili2c update. Same happened to a Colombian model.

>  Would it fail with the old file?

Yes, look at the Travis build after the second commit in this PR.

signedav: Okay thanks for the info.

PR #437, opened 2020-10-15
Generate UUIDs on T_Ili_Id of GPKG layers

signedav:
If there is in the table `t_ili2db_column_prop` for a column `T_Ili_Id` the tag `ch.ehi.ili2db.oidDomain` with the value `INTERLIS.UUIDOID` the field will contain a UUID as default value generated by the expression `substr(uuid(), 2, 36)`.


gacarrillor (approved):
It's fine for me. :+1: 

Note that this PR will have conflicts with #438 (in gpkg_connector.py).

PR #438, opened 2020-10-15
Add support for filtering domain values for extended domains

gacarrillor:
With ili2db 4.4.4 we have an update in domain inheritance.

If DomainB extends DomainA in the conceptual model, we end up with a single table in the database containing all values from DomainA plus all values from DomainB (there may be even repeated values if the value is let untouched from the parent domain). 

Example (single table where we have both parent and child domain values):
image

Therefore, for letting users pick a value from DomainB, we need to filter the values from the database.

The way of doing this is the following:

1. Read the `tag` "ch.ehi.ili2db.enumDomain" (new in ili2db 4.4.4) from table `t_ili2db_column_prop` and store it in a Field object.
2. When creating Relation objects, inspect the referenced table to know if we deal with domain-class relation. If so, go to referencing field and get its enumDomain `setting`.
3. If the corresponding `setting` value is NOT found in `t_ili2b_classname` (field `iliname`), then we have the inheritance case (a domain that extends another domain).
  3.a. For that particular case, store a `child_domain_name` in the Relation object.
  3.b. When configuring widgets, take `child_domain_name` into account for domain RelationReferences (`FilterExpression`: `"thisclass" = 'setting_value').
4. Otherwise, the domain will be a single table and will not need any filter.

See more details at https://github.com/claeis/ili2db/issues/362

### DIFF
--- a/QgisModelBaker/libqgsprojectgen/dataobjects/project.py
+++ b/QgisModelBaker/libqgsprojectgen/dataobjects/project.py
@@ -19,14 +19,17 @@
 """
 from typing import List
 
-from QgisModelBaker.libqgsprojectgen.dataobjects.relations import Relation
-from .layers import Layer
-from .legend import LegendGroup
 from qgis.core import (QgsCoordinateReferenceSystem,
                        QgsProject,
                        QgsEditorWidgetSetup,
                        QgsFieldConstraints)
 from qgis.PyQt.QtCore import QObject, pyqtSignal
+
+from QgisModelBaker.libqgsprojectgen.dataobjects.layers import Layer
+from QgisModelBaker.libqgsprojectgen.dataobjects.legend import LegendGroup
+from QgisModelBaker.libqgsprojectgen.dataobjects.relations import Relation
+
+ENUM_THIS_CLASS_COLUMN = "thisclass"
 
 
 class Project(QObject):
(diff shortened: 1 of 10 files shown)


QgisModelBaker/libqgsprojectgen/dataobjects/project.py
```diff
@@ -115,13 +118,19 @@ def create(self, path: str, qgis_project: QgsProject):
             referencing_field_constraints = referencing_layer.fieldConstraints(rel.referencingFields()[0])
             allow_null = not bool(referencing_field_constraints & QgsFieldConstraints.ConstraintNotNull)
 
+            # If we have an extended ili2db domain, we need to filter its values
+            filter_expression = "\"{}\" = '{}'".format(ENUM_THIS_CLASS_COLUMN,
```
signedav, line 122: Do we really need a constant here? It will be never used again, or am I wrong?

signedav (approved):
Beside minor comment it looks good.

QgisModelBaker/libqgsprojectgen/dataobjects/project.py
```diff
@@ -115,13 +118,19 @@ def create(self, path: str, qgis_project: QgsProject):
...
+                                                       relation.child_domain_name) if relation.child_domain_name else ''
```
signedav, line 123: if there is no `relation.child_domain_name` we still enter something into the filterExpression? Or am I missing something?

signedav: Ah sorry. I misread the `if`. Everything is fine...

QgisModelBaker/libqgsprojectgen/dataobjects/project.py
gacarrillor: Not strictly necessary, just thought about a config file with ili2db common names like "thisclass" so that they can be easily spotted and adjusted.

But it's not this PR's bussiness anyway. If you want we can remove the constant here.

signedav: Hm... Okay, I see the point. You can keep it like this.

PR #440, opened 2020-10-27
Support for ordering form widgets via TOML file

gacarrillor:
Layers usually have certain logic attached to them, like an order of the data attributes they hold.

Currently, attribute forms are created disregarding attribute ordering, but when adding a feature, users definitely need the visual guide that a correct arrangement of attributes can give, specially in forms with a large number of attributes.

This PR makes it possible to define an order of attributes in layer forms by using the TOML file, like this:

    ["MODEL.Topic.Class.AttributeA"]
    form_order=2

    ["MODEL.Topic.Class.AttributeB"]
    form_order=1

Note: All attributes get a 999 order index by default. So if you set a single attribute in your TOML (say, with index = 1 or 10), it will appear as the first attribute in the corresponding form.


gacarrillor:
@m-kuhn What name could we use for specifying the attribute order in forms from the TOML file?

I've initially think of attribute_order, attr_order, form_order.

It looks like this: 

```
["LADM_COL_V3_0.LADM_Nucleo.ExtDireccion.Sector_Ciudad"]
form_order=7
```

signedav:
What happens on form order 0? I assume, since you have the default 999, it will appear as the first (before 1).

signedav (approved):

gacarrillor:
> What happens on form order 0? I assume, since you have the default 999, it will appear as the first (before 1).

Right, it counts as the first one then. 
I guess a negative number will have the same behavior.

PR #449, opened 2020-11-16
Support arguments with spaces and quotes

### LINKED ISSUE #444: v6.3.0: Java 8 could not be found
We tried to generate a project with v6.3.0. We always get the message "Java 8 could not be found". Even though the path is defined. On windows. Any suggestions? 

image

signedav:
In the normal call of `QProcess` we use an array of arguments again instead of passing always a command.

If the command is edited, we pass the command. There a slight escaping is done by replacing `"` by `"""` and putting arguments with spaces into doublequotes.

This fixes #444

### DIFF
--- a/QgisModelBaker/libili2db/iliexecutable.py
+++ b/QgisModelBaker/libili2db/iliexecutable.py
@@ -89,17 +89,28 @@ class IliExecutable(QObject, metaclass=AbstractQObjectMeta):
             return self.ILI2DB_NOT_FOUND
         return ["-jar", ili2db_bin]
 
+    def _escaped_arg(self, argument):
+        if '"' in argument:
+            argument = argument.replace('"', '"""')
+        if ' ' in argument:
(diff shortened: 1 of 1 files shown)


QgisModelBaker/libili2db/iliexecutable.py
```diff
@@ -83,56 +83,73 @@ def _args(self, hide_password):
...
         if not ili2db_bin:
             return self.ILI2DB_NOT_FOUND
+        return ["-jar", ili2db_bin]
 
-        ili2db_jar_arg = ["-jar", ili2db_bin]
+    def _escaped_arg(self, argument):
+        if '"' in argument:
+            argument = argument.replace('"', '"""')
+        if ' ' in argument:
+            argument = ('"'+argument+'"')
```
m-kuhn, line 96:
```suggestion
            argument = '"' + argument + '"'
```

m-kuhn (approved):
LGTM

PR #453, opened 2020-11-24
ili2db also works with newer java version

m-kuhn:
No longer force people to use legacy versions of java

### DIFF
--- a/QgisModelBaker/libili2db/ili2dbutils.py
+++ b/QgisModelBaker/libili2db/ili2dbutils.py
@@ -30,6 +30,8 @@ from QgisModelBaker.libs.packaging import version
 from qgis.PyQt.QtCore import QCoreApplication
 from qgis.PyQt.QtGui import QColor
 
+from packaging import version
+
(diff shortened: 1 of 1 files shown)


m-kuhn:
@gacarrillor what do you think?
Works here with 

```
java --version
openjdk 11.0.9 2020-10-20
OpenJDK Runtime Environment 18.9 (build 11.0.9+11)
OpenJDK 64-Bit Server VM 18.9 (build 11.0.9+11, mixed mode, sharing)
```

signedav:
Is this okay? Can we put that into 6.3.2 release?

m-kuhn:
I think so, yes.
It should not break any existing installation (java 8 still works), it is tested here with java 11. Worst that can happen (unlikely IMHO) is that it does not work on some newer java version somewhere and either ili2db receives a fix or we revert this patch.

signedav (approved):
LGTM

signedav:
Mässi :-)

PR #443, opened 2020-11-05
Schema creation on XTF import

signedav:
Re-introduce parameter `--doSchemaImport` when the file (Geopackage) or the db schema (PG, MSSQL) does not exist.

Fixes #349

PR #457, opened 2020-12-16
Open dropped XTF (or similar) file in Import Dialog

signedav:
Valid extensions for dropped files: 'xtf', 'itf', 'pdf', 'xml', 'xls', 'xlsx'

- when exactly one valid file is dropped and when import dialog is already open -> fill up import file field
- when exactly one valid file is dropped and when import dialog not yet open ->  open import dialog and fill up file field
- when one or more invalid files are dropped -> do normal QGIS behavior
- when one or more files are dropped containing at least one valid file -> show user feedback that Model Baker dialog cannot handle multiple files

### DIFF
--- a/QgisModelBaker/gui/import_data.py
+++ b/QgisModelBaker/gui/import_data.py
@@ -505,9 +505,6 @@ class ImportDataDialog(QDialog, DIALOG_UI):
             self.progress_bar.setValue(75)
             QCoreApplication.processEvents()
 
-    def set_xtf_file(self, xtf_file):
-        self.xtf_file_line_edit.setText(xtf_file)
-
     def __db_connector(self, configuration):
         db_factory = self.db_simple_factory.create_factory(configuration.tool)
         config_manager = db_factory.get_db_command_config_manager(configuration)
--- a/QgisModelBaker/gui/import_data.py
+++ b/QgisModelBaker/gui/import_data.py
@@ -505,6 +505,9 @@ class ImportDataDialog(QDialog, DIALOG_UI):
             self.progress_bar.setValue(75)
             QCoreApplication.processEvents()
 
+    def set_xtf_file(self, xtf_file):
+        self.xtf_file_line_edit.setText(xtf_file)
+
     def __db_connector(self, configuration):
         db_factory = self.db_simple_factory.create_factory(configuration.tool)
         config_manager = db_factory.get_db_command_config_manager(configuration)
--- a/QgisModelBaker/qgismodelbaker.py
+++ b/QgisModelBaker/qgismodelbaker.py
@@ -29,7 +29,7 @@ from QgisModelBaker.gui.import_data import ImportDataDialog
 from QgisModelBaker.libqgsprojectgen.dataobjects.project import Project
 from QgisModelBaker.libqgsprojectgen.generator.generator import Generator
 
-from qgis.core import QgsProject
+from qgis.core import Qgis, QgsProject
 from qgis.utils import available_plugins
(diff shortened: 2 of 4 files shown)


signedav:
@m-kuhn I'm thinking about the valid formats: 'xtf', 'itf', 'pdf', 'xml', 'xls', 'xlsx' I'm not sure if the user wants to have the import dialog opened on every PDF he drops. What do you think? Should we do it only for xtf?

m-kuhn:
I would go for xtf and itf, that's where model baker shines.

signedav:
@gacarrillor Do you have an idea why the tests are failing? Like e.g. for `Catastro_COL_ES` etc?

gacarrillor:
Actually, it is strange that tests weren't failing before. I'm not sure of what has changed in this PR.

Could you please replace the models inside `testdata/ilimodels/subparent_dir/not_hidden/` by the corresponding 2 models (i.e., only `CIAF_LADM.ili` and `Catastro_COL_ES.ili`) from `testdata/ilimodels/CIAF_LADM/` and try again?

signedav:
Thanks @gacarrillor On master they failed as well. But anyway it works now :tada:

m-kuhn:
Nice.
One thing that I found surprising is that it reused the last used gpkg name (actually hit "continue" without looking at it, reimporting into an existing gpkg). Could we use a file with the same name and a counter (1,2,3) as suffix in a) the same place as the xtf (if it's a writable location) or b) [QStandardPaths::TempLocation](https://doc.qt.io/qt-5/qstandardpaths.html)

m-kuhn:
... in this case it should also use the `--doSchemaImport` parameter as otherwise it fails

signedav:
I agree on your input. But I'm not yet convinced. My concern is the following:

I think it has the last gpkg name stored because a use case is:
- User generates a project with schema import
- User imports XTF files for this project (and usually it's the just created gpgk or db schema)

If a suffix is appended, the user can be quite confused why the data are not in the open project just generated.

Possibilities are:
- another behavior in open the dialog by the toolbar / menu to the drop XTF
- popping up a message in any case: one "Do you want to import to an exiting database schema?" or the other "The schema / file does not exist. Do you want to create it first?"

At the moment. I would go for having the same behavior without any suffix but having this popup message. But very open for a better solution.

Btw. a `--doSchemaImport` should be done already since #443 or did you observed something else?

m-kuhn:
I think the main use case we are targetting here is a user who downloads/receives by mail/... an xtf. In such a case he normally does not have a pre-created schema and the last used gpkg is completely random.

Therefore I think the default behavior should be a new unique file name (suffix, random file name, ...) which is a predictable behavior with no risk of damaging anything.

The other approach would also be nice, but it should depend on a currently opened project and we need to think what happens if postgres is involved.

signedav:
So you suggest (a) to have another behavior on opening the import dialog when dropping an XTF to open it with the manu/toolbar?
If so, then would you prefer as well, that (b) not the last used db-type (PG, GPKG etc) is chosen but always GPKG instead? 
I think, if we go for (a) to should go for (b) as well...

m-kuhn:
Yes, this behavior sounds very good

signedav:
Okay then. I'll go for the temp folder then.

signedav:
Decided to write directly into the settings instead of passing the values - like this they are stored in case the user closes and reopens the dialog.

And I decided to make a human readable timestamp instead a counter. I think nobody drops file faster than milliseconds.
image

m-kuhn (approved):
Very nice

PR #460, opened 2021-01-12
Drop XTF followup - ask the user if using Model Baker or QGIS default behavior

signedav:
**The behavior is:**
If not XTF / ITF -> QGIS default behavior
If XTF / ITF multiple files -> QGIS default behavior
If XTF / ITF single file -> Dialog asking what to do? -> selected behavior
IF XTF / ITF single file, and selected don't ask again in dialog -> last chosen behavior

> The difference to the last version with multiple files is, that the default behavior is taken instead printing a message: "Model Baker cannot handle multiple dropped files" - this because before the QGIS behavior is "removed" completely and now it's part of the solution, so it should be used still.

When dropping an XTF or an ITF file a dialog is opened:
baker1

On checking the checkbox, the behavior is used without popping up a dialog in future. To get the dialog again, the user can reset the "check" in the settings. 
baker2
Since there is no need for this setting, when it's not checked in the drop-dialog (because it's not decided if it should use the model baker setting or not) it's a disabled widget.
image

### DIFF
--- a/QgisModelBaker/gui/options.py
+++ b/QgisModelBaker/gui/options.py
@@ -19,7 +19,7 @@
 """
 import webbrowser
 
-from QgisModelBaker.libili2db.globals import DbIliMode, displayDbIliMode
+from QgisModelBaker.libili2db.globals import DbIliMode, displayDbIliMode, DropMode
 from QgisModelBaker.libqgsprojectgen.db_factory.db_simple_factory import DbSimpleFactory
 from QgisModelBaker.libili2db.ili2dbconfig import SchemaImportConfiguration, ImportDataConfiguration, \
     ExportConfiguration
@@ -43,6 +43,9 @@ class OptionsDialog(QDialog, DIALOG_UI):
 
         self.pg_user_line_edit.setText(configuration.super_pg_user)
         self.pg_password_line_edit.setText(configuration.super_pg_password)
+        settings = QSettings()
+        print(settings.fileName())
+        self.chk_dontask_to_handle_dropped_files.setChecked(DropMode(settings.value('QgisModelBaker/drop_mode', defaultValue=3)) != DropMode.ask)
 
         self.custom_model_directories_line_edit.setText(
             configuration.custom_model_directories)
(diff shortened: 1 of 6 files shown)


signedav:
@m-kuhn I've been a little unsure regarding the settings. First in the front-end, where this checkbox would fit in. And second in the back-end, where and how it should be stored. Currently some settings are kept in a configuration object and others are always directly read and wrote into the QSettings. Is there a fix structure, or should this be discussed somehow?

QgisModelBaker/ui/drop_message.ui
```diff
@@ -0,0 +1,84 @@
...
+     </property>
+     <property name="openExternalLinks">
+      <bool>true</bool>
+     </property>
+    </widget>
+   </item>
+   <item row="1" column="0">
+    <widget class="QCheckBox" name="chk_dontask">
+     <property name="text">
+      <string>Apply behavior for all XTF / ITF files (you can undo this in the Model Baker settings)</string>
```
m-kuhn, line 40: Could be "Remember my decision and don't ask me again", I think that's the standard wording

QgisModelBaker/gui/options.py
```diff
@@ -83,7 +83,7 @@ def __init__(self, configuration, parent=None):
         self.ili2db_command_reload()
 
         settings = QSettings()
-        drop_mode = DropMode(settings.value('QgisModelBaker/drop_mode', defaultValue=3))
+        drop_mode = DropMode(settings.value('QgisModelBaker/drop_mode', 3, int))
```
m-kuhn, line 86:
```suggestion
        drop_mode = DropMode(settings.value('QgisModelBaker/drop_mode', DropMode.ask, int))
```

?

signedav: I have to store the settings as it's value (integer) because I cannot stor Qt Objects (enums)...

m-kuhn: We could use

`DropMode.ask.value` (3) or even nicer `DropMode.ask.name` (yes)

m-kuhn: Another note, could we use `ASK` (uppercase) -> see https://mail.python.org/pipermail/python-ideas/2016-September/042340.html

signedav: Yes. We store the Value like this:
```
settings.setValue('QgisModelBaker/drop_mode', DropMode.ask.value)
```
So i think 3 is stored, no?
And so we have to read 3 (as int) and then convert to the enum with `DropMode(3)`

Or what am I missing?

m-kuhn (approved):

PR #465, opened 2021-01-25
Fix modeldir support of % values

### LINKED ISSUE #464: Interlis: Custom model directories - problems after modification
Maybe I messed up something else, but I modified the different default paths for "custom model directories" to only "%XTF_DIR". When importing the ITF, I got this log:

```
.....
Info: driverName <SQLiteJDBC>
Info: driverVersion <native>
Info: compile models...
Error: C:\Program Files\QGIS 3.16\bin\156021 Gae Melioration AccessExp.Ili:There is no such file.
Error: 156021 Gae Melioration AccessExp.Ili (Das System kann die angegebene Datei nicht finden)
compiler failed
Finished (1)
```

For me it's strange that it looks in the "bin-folder" - it was not defined.
After deleting my personal settings and activating the default paths for "custom model directories" it worked fine.

Bug or feature?

m-kuhn:
The old approach was locally searching for .ili files in the model directories
and recursively adding all of them. The setting actually supports a couple of
magic strings like `%XTF_DIR` and similar which got lost this way because
we couldn't glob them.

This approach always adds all directories from the setting and adds
additional subdirs only if they contain .ili files

Fixes #464

### DIFF
--- a/QgisModelBaker/libili2db/ili2dbutils.py
+++ b/QgisModelBaker/libili2db/ili2dbutils.py
@@ -105,13 +105,17 @@ def get_ili2db_bin(tool, db_ili_version, stdout, stderr):
 
 def get_all_modeldir_in_path(path, lambdafunction=None):
     all_subdirs = [path[0] for path in os.walk(path)] # include path
-    modeldir = ''
+    # Make sure path is included, it can be a special string like `%XTF_DIR`
+    modeldirs = [path]
     for subdir in all_subdirs:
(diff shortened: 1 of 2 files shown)


m-kuhn:
@juusechec @gacarrillor pinging you as the author of the subdir functionality, how does this look to you?

gacarrillor (approved):
I see the changes are fine. I wonder why we weren't adding the `path` itself.

PR #473, opened 2021-02-17
push ili2pg to 4.4.6

olivierdalang:
Just pushing ili2pg to 4.4.6 to benefit from bugfix https://github.com/claeis/ili2db/issues/381

### DIFF
--- a/QgisModelBaker/libqgsprojectgen/db_factory/gpkg_factory.py
+++ b/QgisModelBaker/libqgsprojectgen/db_factory/gpkg_factory.py
@@ -53,7 +53,7 @@ class GpkgFactory(DbFactory):
         if db_ili_version == 3:
             return '3.11.3'
         else:
-            return '4.4.4'
+            return '4.4.6'
 
     def get_tool_url(self, db_ili_version):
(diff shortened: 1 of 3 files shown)


signedav (approved):

olivierdalang:
@signedav there are failing tests though, with errors like these :

```
self.assertEqual(layer.layer.displayExpression(), '"T_Id"')
E               AssertionError: '"T_Ili_Tid"' != '"T_Id"'
```

Do you know why this is ? I don't think this is expected from changin minor version ?

signedav:
It's the QGIS version. On a QGIS 3.16 version, when I make the project generation with the Model Baker on a Class that has no meta data defined for the display message it used to take the first attribute` T_id (int) `as the display expression value. On a master version (one week old) it takes the first string value as display message` T_Ili_tid (string)`. Actually the tests are here to test the meta data and not the QGIS behavior - so I suggest to change it to t_ili_tid... I can do it, I have it checked out anyway. One moment.

signedav:
Oh you forked. I opened another pr #475 - let's see if the tests work.
Sorry for the git-mess

PR #475, opened 2021-02-25
push ili2pg to 4.4.6 with t_ili_id fix

signedav:
Let's see if tests are happy


m-kuhn (approved):

olivierdalang (approved):
Look good to me too !

Don't really get why displayExpression changed in latest QGIS though, but that's for another story !

m-kuhn:
I guess it's an effect of https://github.com/qgis/QGIS/pull/41372 but didn't check in depth

PR #476, opened 2021-03-12
Fix Layernames for tables with multiple geometry columns

### LINKED ISSUE #469: ILI and ITF for "Melioration_Kt_SHLV95"
My knowledge in QGis is limited, so maybe it is a User-issue..

I have ILI and ITF for "Melioration_Kt_SHLV95", and the ILI contains a 
"    TABLE Masse = .."
The ITF contains a section with 12 objects, beginning with
TABL Masse
OBJE 1 ...

Importing it to QGis, I get a layer displayed as "--- Masse) [12]" - "line" is correct, the bracket) is strange, 12 objects are correct
Looking at the table "Masse)" I have 12 rows, but only with "T_Id" and nothing else ..

Is it a known problem?

signedav:
These commits are cherrypicks from the branch of usabilityhub_toppings - I think it's better to consider it seperately.

**GeoPackage:**
fix layernames on tables that are created twice in geopackage  and contain the iliname with brackets (e.g. KbS_LV95_V1_4.Belastete_Standorte.Belasteter_Standort.Geo_Lage_Punkt(KbS_LV95_V1_4.Belastete_Standorte.Belasteter_Standort)) They are now named **Belasteter_Standort (Geo_Lage_Punkt)** This fixes #469

**Postgres:**
On creating layers from a Postgres dataset where a table has multiplegeometry columns, this table is loaded for each geometry column. To have a unique name, we check for duplicate tablenames and if we found one we check out the t_ili2db_attrname for the geometry column and add it's proper name to the table.

### DIFF
--- a/QgisModelBaker/libqgsprojectgen/generator/generator.py
+++ b/QgisModelBaker/libqgsprojectgen/generator/generator.py
@@ -113,7 +113,13 @@ class Generator(QObject):
                 if is_domain and is_attribute:
                     short_name = record['ili_name'].split('.')[-2] + '_' + record['ili_name'].split('.')[-1] if 'ili_name' in record else ''
                 else:
-                    short_name = record['ili_name'].split('.')[-1] if 'ili_name' in record else ''
+                    if 'ili_name' in record:
+                        match = re.search('([^\(]*).*', record['ili_name'])
+                        if match.group(0) == match.group(1):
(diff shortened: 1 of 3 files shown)


gacarrillor (approved):
It would be great to have a unit test for these cases. Would it be possible, @signedav?

signedav:
Yes. Will do it.

PR #477, opened 2021-03-19
No double loading of layers in layertree

signedav:
On having a project open and importing an additional model into the same schema /db file, the layers are reloaded and double. So the fix is, that it does not reload the already existing layers.

### DIFF
--- a/QgisModelBaker/libqgsprojectgen/dataobjects/legend.py
+++ b/QgisModelBaker/libqgsprojectgen/dataobjects/legend.py
@@ -62,6 +62,8 @@ class LegendGroup(object):
         if not group:
             group = qgis_project.layerTreeRoot()
 
+        existing_layer_names = [found_layer.name() for found_layer in group.findLayers()]
+
         for item in self.items:
             if isinstance(item, LegendGroup):
                 subgroup = group.findGroup(item.name)
@@ -71,8 +73,9 @@ class LegendGroup(object):
                 item.create(qgis_project, subgroup)
             else:
                 layer = item.layer
-                index = get_suggested_index_for_layer(layer, group, self.ignore_node_names) if layer.isSpatial() else 0
-                group.insertLayer(index, layer)
+                if layer.name() not in existing_layer_names:
+                    index = get_suggested_index_for_layer(layer, group, self.ignore_node_names) if layer.isSpatial() else 0
+                    group.insertLayer(index, layer)
 
     def is_empty(self):
         return not bool(self.items)


gacarrillor (approved):
I think this is nice to have, thanks.

Just a question here. For migrating data from database 1 (model A) to database 2 (model A), we load layers from DB 1, and just before loading layers from DB 2, we create a group (say `my_group1`) for all layers and groups from DB 1 (and rename groups from DB 1 because [`findGroup()`](https://github.com/opengisch/QgisModelBaker/blob/56970367e3225a6d16b5dd1cb50027b685a6aa88/QgisModelBaker/libqgsprojectgen/dataobjects/legend.py#L69) is recursive), so we don't get them mixed in the same groups when QMB loads the DB2's layers. 

In that way we can load layers from 2 DBs to migrate data from one to another, but we don't get them mixed. After the migration we can simply drop the `my_group1` and we are done.

Do you think that after this PR, we can still count with such use case?

signedav:
@gacarrillor since it's only considering same named layers in the scope of a group this should still work. But good input, since I have to check if this is still provided on layertree toppings in the other PR.
This one I merge. Thanks.

signedav:
Hi @gacarrillor After double check I found out that I broke your behavior. I'll take care of it...

signedav:
`group.children()` should make it - I'll fix it in the UsabILItyhub PR...

signedav:
Okay. The most proper solution is not to check for layer name existance but for layer source existance:
```
existing_layer_source_uris = [found_layer.layer().dataSource.dataSourceUri() for found_layer in qgis_project.layerTreeRoot().findLayers()]
```
This solves your use case @gacarrillor and the one for usabilityhub.
See https://github.com/opengisch/QgisModelBaker/pull/468/commits/36e481084960583c771d327790cb74ad291e4cfd

gacarrillor:
Thanks @signedav for taking care of the use case I mentioned. 

Your solution looks good. I'm just thinking about edge cases...

I don't know if `dataSourceUri()` would give us the same output depending on some DB connection settings. For instance, what would happen if you load your layers with an authentication configuration set, store the .qgs/z project and later on you use Model Baker to re-attempt to load the same model, but this time without an authentication configuration but with plain password. Would the comparison still work? 

This has happened me before comparing URIs. I used a .qgs/z project that didn't store passwords and then the comparison failed just because of the password (the loaded layer didn't have it but the new layer to be loaded did).

Do you think the solution would work in these cases?

m-kuhn:
@gacarrillor I think what you describe is the limitation of the dataSourceUri approach, (`sslmode`, pg service vs hostname vs ip address, ... being the usual supects).

Do you think this is a huge issue in production?

PR #491, opened 2021-06-01
gh workflow - use published as type

signedav:
because otherwise no workflows are triggered on pre-release


signedav:
> Note: The prereleased type will not trigger for pre-releases published from draft releases, but the published type will trigger. If you want a workflow to run when stable and pre-releases publish, subscribe to published instead of released and prereleased.

Issue #495, opened 2021-06-09
Weird behavior of model (and metaconfig) text edit dropdown


signedav:
We fire the signal `punched` on the text edit to get the popup with all the values on getting focus.

```
    def focusInEvent(self, e):
        super(CompletionLineEdit, self).focusInEvent(e)
        self.punched.emit()
```

This leads to this weird behavior (sometimes):
![emptypopup](https://user-images.githubusercontent.com/28384354/121314876-17bb3c00-c908-11eb-9717-bb541810eece.gif)

I didn't found it out yet why.

PR #496, opened 2021-06-09
Fix unreliable drop down on metaconfig selection widget

signedav:
Loads and unloads the metaconfig model items (UsabILIty Hub Metaconfigurations) reliable.
Marks the box orange when the metaconfig string is not finding a metaconfiguration.

This fixes an unreliable behavior. The only unnice thingy left is #495 - but this is only visually.


signedav:
Merged it. I think it's for sure an improvement.

PR #499, opened 2021-06-16
No need to explicitly upload artifacts

suricactus:
`qgis-plugin-ci` already automatically uploads the artifacts for you.


signedav (approved):
Thanks

PR #508, opened 2021-06-23
Checks if file extension is ili (and not ILI)

### LINKED ISSUE #442: Error message .Ili import (not.ili)
Model baker "projekt generator" recognises ili-files with the false extension .Ili (you can choose the file) but generates in error on import (model not found). Renaming the extension to .ili solves the problem. 
To be more user friendly, two options would be nice :
-  .Ili are not found when browsing, so the user would see the problem
-  Clear error message, for example: "Error: the extension of the interlis file has to be .ili).

signedav:
and reads last error from the validator in case of negative validation
fixes #442

image

### DIFF
--- a/QgisModelBaker/gui/generate_project.py
+++ b/QgisModelBaker/gui/generate_project.py
@@ -101,7 +101,7 @@ COLOR_TOPPING = '#341d5c'
 
 class GenerateProjectDialog(QDialog, DIALOG_UI):
 
-    ValidExtensions = ['ili', 'ILI']
+    ValidExtensions = ['ili']
 
     def __init__(self, iface, base_config, parent=None):
         QDialog.__init__(self, parent)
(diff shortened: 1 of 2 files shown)


QgisModelBaker/gui/generate_project.py
```diff
@@ -274,9 +274,10 @@ def accepted(self, edited_command=None):
                         return
 
                 if self.ili_file_line_edit.text().strip() and \
-                        self.ili_file_line_edit.validator().validate(configuration.ilifile, 0)[0] != QValidator.Acceptable:
+                    self.ili_file_line_edit.validator().validate(configuration.ilifile, 0)[0] != QValidator.Acceptable:
+
                     self.txtStdout.setText(
-                        self.tr('Please set a valid INTERLIS file before creating the project.'))
+                        self.tr('Please set a valid INTERLIS file before creating the project. {}').format(self.ili_file_line_edit.validator().error))
```
m-kuhn, line 280: This will print `None` if unset, is that something to take care of?

m-kuhn (approved):

QgisModelBaker/gui/generate_project.py
signedav: I usually prefer `None` over `''` and in this case it's not important since we do not print this when `error` is empty. But maybe for future use of this `''` would be better in case you attach it to a string...

PR #509, opened 2021-06-25
Allow null in relation references always on domains - when the value …

### LINKED ISSUE #467: Set "Allow NULL" to true at MANDATORY domain related attributes
When a project is generated and a MANDATORY domain related is involved, the "Allow NULL" in the form is deactivated. This leads to the behavior, that the picklist in the form has already a value when a new object is added, the first value in the picklist. So the user is not forced to choose the propper value, may forgets it and saves the wrong value. If the "Allow NULL" is set to true, the user is forced to choose the propper value before he's able to save the form.

Example: Attribute `Datenherr` in https://models.geo.admin.ch/BAFU/Schutzbauten_V1.ili

**Suggestion**: If a domain related attribute is involved, set the "Allow NULL" property always to true.

**Now**:
image

image

**Goal**:
image

image

signedav:
When the value is mandatory the constraints will concern that the user has to choose something. 
image

Otherwise, the first value of the domain table is chosen automatically, what is not what we want.

This fixes #467

### DIFF
--- a/QgisModelBaker/libqgsprojectgen/dataobjects/project.py
+++ b/QgisModelBaker/libqgsprojectgen/dataobjects/project.py
@@ -115,10 +115,6 @@ class Project(QObject):
             assert rel.isValid()
             qgis_relations.append(rel)
 
-            referencing_layer = rel.referencingLayer()
-            referencing_field_constraints = referencing_layer.fieldConstraints(rel.referencingFields()[0])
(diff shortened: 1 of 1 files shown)


signedav:
@gacarrillor and @romefi Should null be allowed only on domain references? 
Shouldn't it allow null on other associations as well?

gacarrillor:
> @gacarrillor and @romefi Should null be allowed only on domain references?
> Shouldn't it allow null on other associations as well?

Sounds good.

PR #515, opened 2021-07-16
Redesign Import Workflow

signedav:
import

## About the structure
We keep information we use for the ili2db commands on top level `workflow_wizard` like eg. the configurations and the models containing the files etc. Because some models need to be on top level, we have all the models there.

## Datasets
- Datasets are created with the dataset manager (or the default dataset on schema import) this means that in case of a basket handling, the data import is *always" an `--update` on importing without any basket handling it's still `--insert`.

### To Do
- [x] More proper logging and feedback 
- [x] Better workflow handling (deactivated next keys or back keys)
- [x] Update Import sessions when changes made in config
- [x] Consider `DbActionType` in database selection and behave accordingly with warnings etc.
- [x] Validate the database before continue
- [x] Big cleanup

### Pending Features
- [x] Data Import functions for DATASETs, BASKETs etc.
- [x] Parsing needed Models from Transfer File
- [x] Export

### Planned as follow up:
- Export of single baskets
- Catalogue Dataset
- File Drop
- Get Catalogues (Transferfiles) from data repo

### Questions
- How should I handle the catalogue dataset?
- Does the xtf model parse makes sense?
- Where should the Transferfile Data Repo be integrated?
- What can be tested in unit tests?


signedav:
To have a little structure, I like to have this PR finalized and then I bring the follow ups with the functionalities:
- Data Import functions for DATASETs, BASKETs etc.
- File Drop
- Parsing needed Models from Transfer File
. Export

@m-kuhn Since to review that can be quite an effort, we might make it together with a phone call.

signedav:
**_WIP again since I will put the dataset part in here as well..._**

PR #516, opened 2021-07-19
Use save file browser for import data

signedav:
It still allows you to select an existing file.

The fix is needed to have the ability to choose a non-existing file, because you can create the GPKG with ili2db Parameter `--doschemaimport`.

Negative point is that because it's a default Qt Component (`QFileDialog`) on providing the ability to choose a non-exisiting file (by using `getSaveFileName`, it displays a warning like this when choosing an existing file:
image

So this could irritate the use, because we do not "replace" the file when importing into an existing GPKG-File.

So I'm not sure about this fix.

### DIFF
--- a/QgisModelBaker/gui/panel/gpkg_config_panel.py
+++ b/QgisModelBaker/gui/panel/gpkg_config_panel.py
@@ -63,7 +63,7 @@ class GpkgConfigPanel(DbConfigPanel, WIDGET_UI):
             validator = self.gpkgSaveFileValidator
             file_selector = make_save_file_selector(self.gpkg_file_line_edit, title=self.tr("Open GeoPackage database file"),
                                         file_filter=self.tr("GeoPackage Database (*.gpkg *.GPKG)"),
-                                                    extensions=['.' + ext for ext in self.ValidExtensions])
+                                                    extensions=['.' + ext for ext in self.ValidExtensions], dont_confirm_overwrite = (self._db_action_type == DbActionType.IMPORT_DATA) )
         else:
             validator = self.gpkgOpenFileValidator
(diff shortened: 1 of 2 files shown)


andreasneumann:
@m-kuhn - no I don't agree. I get this message when creating a new layer in an already existing geopackage:

image

That is quite a different message to the user.

m-kuhn:
We are talking about different dialogs, see this screencast:

Peek 2021-07-19 16-47

signedav:
If it's the known behavior in QGIS I think I'm +1 to do it the same way here.

andreasneumann:
Seems like on Windows this warning message doesn't appear - at least not in QGIS 3.16.

See my screen recording here:

https://user-images.githubusercontent.com/884476/126181970-33d68c55-783a-44f4-aa0d-0494dffff5d7.mp4

m-kuhn:
Oh, I see, then it's probably also not an issue for model baker neither

andreasneumann:
> Oh, I see, then it's probably also not an issue for model baker neither

Probably - you can release an "experimental" version and I can test and report back.

signedav:
@andreasneumann  Can you test the following:
gk

In Generate Project on Model Import select an existing GPKG

signedav:
It's the same in the current version on Schema Import to GKPG.

andreasneumann:
> @andreasneumann Can you test the following:

hm - strange. Here I get the replacement warning. They seem to be different dialogs then.

signedav:
@m-kuhn Just found an info to solve it https://doc.qt.io/qt-5/qfiledialog.html#Option-enum
I'll check it out. Thanks.

signedav:
Actually this message should be avoided in schema import as well since models can be "added" to the GPKG. And there we have the message anyway:
image

m-kuhn (approved):

PR #519, opened 2021-07-30
Basic Basket Dataset Support

signedav:
`--createBasketCol` is active on default - you are still able to deactivate it.

When it's enabled, the following things are done:
- The layers have a `t_basket` column (ili2db) and the fields are integrated in the form.
- The layers `t_ili2db_basket` and `t_ili2db_dataset` are loaded in the project. They are needed to have the reference to them (to fill up `t_basket` in the single layers).
- For the layer `t_ili2db_basket` a display expression is created to provide the user a nice selection in the Relation Reference Widget of `t_basket`.

Questions:
- The "technical" tables `t_ili2db_basket` and `t_ili2db_dataset` are needed to be loaded. But where should they be placed? "Domain" would be a nice group since these are the tables the user usually is not interested in, but it's actually the wrong place, since these are no "Domain" tables.
- And should the tables `t_ili2db_basket` and `t_ili2db_dataset` be editable? It's planned to have a Dataset Manager as Model Baker functionality to edit them. But still it might be interesting for the user to edit them.

To Do:
~~- [ ] In case that the tables `t_ili2db_basket` and `t_ili2db_dataset` are editable, then we might care for `t_id` (and maybe `t_ili_tid`) creation, since that is not done by the DB (neither GeoPackage nor PG).~~
- [x] Filter the baskets for the topic the current table is in
- [x] Tests

### DIFF
--- a/QgisModelBaker/libqgsprojectgen/dbconnector/db_connector.py
+++ b/QgisModelBaker/libqgsprojectgen/dbconnector/db_connector.py
@@ -33,8 +33,11 @@ class DBConnector(QObject):
         self.QGIS_TIME_TYPE = 'time'
         self.QGIS_DATE_TIME_TYPE = 'datetime'
         self.iliCodeName = ''  # For Domain-Class relations, specific for each DB
-        self.tid = ''  # For BAG OF config, specific for each DB
+        self.tid = ''  # For BAG OF config and basket handling, specific for each DB
+        self.tilitid = ''  # For basket handling, specific for each DB
         self.dispName = ''  # For BAG OF config, specific for each DB
+        self.basket_table_name = ''  # For basket handling, specific for each DB
+        self.dataset_table_name = ''  # For basket handling, specific for each DB
 
     def map_data_types(self, data_type):
         '''Map provider date/time types to QGIS date/time types'''
--- a/QgisModelBaker/libqgsprojectgen/dbconnector/db_connector.py
+++ b/QgisModelBaker/libqgsprojectgen/dbconnector/db_connector.py
@@ -185,7 +185,7 @@ class DBConnector(QObject):
         tables_info = self.get_tables_info()
         relations_info = self.get_relations_info()
         meta_attrs_info = self.get_meta_attrs_info()
-        basket_handling_info = self.get_basket_handling_info()
+        basket_handling = self.get_basket_handling()
         mapping_ili_elements = []
         static_tables = []
         detected_tables = []
@@ -207,7 +207,7 @@ class DBConnector(QObject):
                 if record['tablename'] in IGNORED_TABLES:
                     static_tables.append(record['tablename'])
                     continue
-                if not basket_handling_info and record['tablename'] in BASKET_TABLES:
+                if not basket_handling and record['tablename'] in BASKET_TABLES:
                     static_tables.append(record['tablename'])
                     continue
 
@@ -242,7 +242,7 @@ class DBConnector(QObject):
         """
         return None
 
-    def get_basket_handling_info(self):
+    def get_basket_handling(self):
         """
         Returns if there is a basket handling enabled according to the settings table
         """
--- a/QgisModelBaker/libqgsprojectgen/dbconnector/db_connector.py
+++ b/QgisModelBaker/libqgsprojectgen/dbconnector/db_connector.py
@@ -19,7 +19,7 @@
 
 from qgis.PyQt.QtCore import QObject, pyqtSignal
 
-from .config import IGNORED_SCHEMAS, IGNORED_TABLES, IGNORED_ILI_ELEMENTS
+from .config import IGNORED_SCHEMAS, IGNORED_TABLES, IGNORED_ILI_ELEMENTS, BASKET_TABLES
 
 class DBConnector(QObject):
     '''SuperClass for all DB connectors.'''
@@ -185,6 +185,7 @@ class DBConnector(QObject):
         tables_info = self.get_tables_info()
         relations_info = self.get_relations_info()
         meta_attrs_info = self.get_meta_attrs_info()
+        basket_handling_info = self.get_basket_handling_info()
         mapping_ili_elements = []
         static_tables = []
         detected_tables = []
@@ -204,6 +205,10 @@ class DBConnector(QObject):
                     continue
             if 'tablename' in record:
                 if record['tablename'] in IGNORED_TABLES:
+                    static_tables.append(record['tablename'])
+                    continue
+                if not basket_handling_info and record['tablename'] in BASKET_TABLES:
+                    print(f"append table to ignored tables {record['tablename']}")
                     static_tables.append(record['tablename'])
                     continue
 
@@ -237,6 +242,12 @@ class DBConnector(QObject):
         Returns the version of the ili2db application that was used to create the schema
         """
         return None
+
+    def get_basket_handling_info(self):
+        """
+        Returns if there is a basket handling enabled according to the settings table
+        """
+        return False
 
 
 class DBConnectorError(Exception):
--- a/QgisModelBaker/libqgsprojectgen/dbconnector/db_connector.py
+++ b/QgisModelBaker/libqgsprojectgen/dbconnector/db_connector.py
@@ -208,7 +208,6 @@ class DBConnector(QObject):
                     static_tables.append(record['tablename'])
                     continue
                 if not basket_handling_info and record['tablename'] in BASKET_TABLES:
-                    print(f"append table to ignored tables {record['tablename']}")
                     static_tables.append(record['tablename'])
                     continue
 
--- a/QgisModelBaker/libqgsprojectgen/dbconnector/mssql_connector.py
+++ b/QgisModelBaker/libqgsprojectgen/dbconnector/mssql_connector.py
@@ -544,3 +544,14 @@ WHERE TABLE_SCHEMA='{schema}'
             return 3
         else:
             return 4
+
+    def get_basket_handling_info(self):
+        """Entry exists when it's active. Currently we don't use the content of the entry besides that."""
+        if self.schema:
+            cur = self.conn.cursor()
+            cur.execute("""SELECT setting
+                           FROM {schema}.t_ili2db_settings
+                           WHERE tag = 'ch.ehi.ili2db.BasketHandling'
+                        """.format(schema=self.schema))
+            return bool(cur.rowcount)
+        return False
--- a/QgisModelBaker/libqgsprojectgen/dbconnector/mssql_connector.py
+++ b/QgisModelBaker/libqgsprojectgen/dbconnector/mssql_connector.py
@@ -43,7 +43,10 @@ class MssqlConnector(DBConnector):
         self._bMetadataTable = self._metadata_exists()
         self.iliCodeName = 'iliCode'
         self.tid = 'T_Id'
+        self.tilitid = 'T_Ili_Tid'
         self.dispName = 'dispName'
+        self.basket_table_name = 't_ili2db_basket'
+        self.dataset_table_name = 't_ili2db_dataset'
 
     def map_data_types(self, data_type):
         result = data_type.lower()
--- a/QgisModelBaker/libqgsprojectgen/dbconnector/mssql_connector.py
+++ b/QgisModelBaker/libqgsprojectgen/dbconnector/mssql_connector.py
@@ -545,7 +545,7 @@ WHERE TABLE_SCHEMA='{schema}'
         else:
             return 4
 
-    def get_basket_handling_info(self):
+    def get_basket_handling(self):
         """Entry exists when it's active. Currently we don't use the content of the entry besides that."""
         if self.schema:
             cur = self.conn.cursor()
@@ -553,5 +553,7 @@ WHERE TABLE_SCHEMA='{schema}'
                            FROM {schema}.t_ili2db_settings
                            WHERE tag = 'ch.ehi.ili2db.BasketHandling'
                         """.format(schema=self.schema))
-            return bool(cur.rowcount)
+            content = cur.fetchone()
+            if content:
+                return content[0] == 'readWrite'
         return False
--- a/QgisModelBaker/libqgsprojectgen/dbconnector/mssql_connector.py
+++ b/QgisModelBaker/libqgsprojectgen/dbconnector/mssql_connector.py
@@ -27,6 +27,7 @@ from .db_connector import (DBConnector, DBConnectorError)
 
 METADATA_TABLE = 't_ili2db_table_prop'
 METAATTRS_TABLE = 't_ili2db_meta_attrs'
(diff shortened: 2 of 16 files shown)


signedav:
@m-kuhn I do need your feedback on this. Mainly considering the editing of t_ili2db_tables (what is a novum - we never did that before in Model Baker created projects) and the visibility and placing of them. Do you have any concerns about it? And what is your opinion about the questions above? Thanks a lot.

signedav:
After the call with Matthias we decided to a. make them read-only and make a group called "system" (or similar) and maybe "hide" it.

QgisModelBaker/tests/test_export.py
```diff
@@ -393,6 +403,7 @@ def test_ili2db3_export_mssql(self):
             dataImporter.tool, 'ilimodels/CIAF_LADM')
         dataImporter.configuration.ilimodels = 'CIAF_LADM'
         dataImporter.configuration.dbschema = importer.configuration.dbschema
+        dataImporter.configuration.create_basket_col = False
```
gacarrillor, line 406: Why not just leaving it `False` by default? :) 

Just kidding, I guess your users need it active by default.

gacarrillor (commented):
Thanks for requesting a review here.

I think with this PR we are separating MB and _Asistente LADM-COL_ (our plugin in SwissTierras Colombia) a bit, and that's expected because our use cases are simply different. MB is a generic tool and our plugin is much more specific. That means MB needs to give power users all they need and make it available for them, whereas in our plugin we do our best to hide complexity to end users.

Therefore, I see this as a good move for MB, but I'm not sure how we'll end up using MB for our plugin from this PR on. I still would need to test these changes via API (that's how we use MB) and evaluate the impact.

There are a couple of new default modes that are not what we had in mind, but that do make sense for the MB use case. First, we hide `t_basket` column from the user and fill it via application (under the hood). Second, we don't load `t_ili2db_basket` nor `t_ili2db_dataset` tables, but I think this one could be easier to adapt to MB, if you think about hiding them (i.e., making them 'private' layers). I think read-only basket/dataset tables could also represent an issue for us, but I'm not sure if you intend to deal with it via the _Dataset Manager_ you mention in the description.

Also, for a generic approach, I'd thought of MB letting users to add baskets to their DBs and choosing which basket they were going to work on during a QGIS/MB session (I added an inline comment below to confirm that you don't have that in mind). But perhaps that's too generic and you don't want to be that generic, which I'd understand :).

In summary, nice addition to MB since basket handling is a powerful ili2db functionality! 

I haven't tested this PR from a user perspective (neither via GUI or API), but I think what you state in the description (and in the code itself) makes sense.

QgisModelBaker/libqgsprojectgen/dbconnector/db_connector.py
```diff
@@ -33,8 +33,11 @@ def __init__(self, uri, schema, parent=None):
         self.QGIS_TIME_TYPE = 'time'
         self.QGIS_DATE_TIME_TYPE = 'datetime'
         self.iliCodeName = ''  # For Domain-Class relations, specific for each DB
-        self.tid = ''  # For BAG OF config, specific for each DB
+        self.tid = ''  # For BAG OF config and basket handling, specific for each DB
+        self.tilitid = ''  # For basket handling, specific for each DB
```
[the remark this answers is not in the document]
signedav: Maybe it's not good naming but I did it according the existing `self.tid` - this variables do not contain the Tid value but the column name instead. E.g. in Postgres `self.tilitid` is `t_ili_tid` and in Geopackage `T_Ili_Tid` etc. Might need at least a better comment. Like `# DB specific column name, used for BAG OF config and basket handling`

QgisModelBaker/libqgsprojectgen/generator/generator.py
```diff
@@ -114,12 +115,14 @@ def layers(self, filter_layer_list=[]):
             is_attribute = bool(record['attribute_name']) if 'attribute_name' in record else False
             is_structure = record['kind_settings'] == 'STRUCTURE' if 'kind_settings' in record else False
             is_nmrel = record['kind_settings'] == 'ASSOCIATION' if 'kind_settings' in record else False
+            is_basket_table = record['tablename'] == self._db_connector.basket_table_name if 'tablename' in record else False
+            is_dataset_table = record['tablename'] == self._db_connector.dataset_table_name if 'tablename' in record else False
```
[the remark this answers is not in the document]
signedav: Yea, I use them normally but here I did it according to the existing definitions above. But the `get`s are nicer, I agree.

QgisModelBaker/tests/test_export.py
signedav: hehe

signedav:
Thanks @gacarrillor 

> I think with this PR we are separating MB and _Asistente LADM-COL_ (our plugin in SwissTierras Colombia) a bit, and that's expected because our use cases are simply different. MB is a generic tool and our plugin is much more specific. That means MB needs to give power users all they need and make it available for them, whereas in our plugin we do our best to hide complexity to end users.

In Model Baker we are always trying to find the balance between simplicity and power to make what power users need. Hiding complexity is always a good thing. Anyway, the `t_basket` column with the Relation Reference widget (using the tables) is provided so the user can select his basket/dataset on the project no matter if Model Baker is installed or not. But when they have Model Baker installed - default values are provided, so they do not need to struggle around with this field. See this the Dataset Collector Combo for that https://github.com/opengisch/QgisModelBaker/pull/520

> Therefore, I see this as a good move for MB, but I'm not sure how we'll end up using MB for our plugin from this PR on. I still would need to test these changes via API (that's how we use MB) and evaluate the impact.
There are a couple of new default modes that are not what we had in mind, but that do make sense for the MB use case. First, we hide `t_basket` column from the user and fill it via application (under the hood). Second, we don't load `t_ili2db_basket` nor `t_ili2db_dataset` tables, but I think this one could be easier to adapt to MB, if you think about hiding them (i.e., making them 'private' layers). 

What would make these changes "compatible" with your plugin? Or would you like to have kind of "modes"? It's always good to keep synergies - for both sides.
Btw. How is the `t_basket` column filled in your plugin? Where the user sets the dataset/basket to use? Or is there one dataset used per project?

> I think read-only basket/dataset tables could also represent an issue for us, but I'm not sure if you intend to deal with it via the _Dataset Manager_ you mention in the description.

Yes. The thing is the following. A user has usually one basket per topic-dataset-combination. This means usually they want to create a dataset (and are not interested in any baskets at all). With the upcoming dataset manager the user should be able to create a new dataset (name it e.g. "Zurich Region") and then "Create baskets for all the topics" - so the dataset will be created by the user and the basket-table-entries automatically on request. That's why we decided against having the tables editable and want to make them as "untouchable" for the user as possible.

> Also, for a generic approach, I'd thought of MB letting users to add baskets to their DBs and choosing which basket they were going to work on during a QGIS/MB session (I added an inline comment below to confirm that you don't have that in mind). But perhaps that's too generic and you don't want to be that generic, which I'd understand :).

Not sure if I understand you correctly. The users can choose the basket they are working on in the "Dataset Selector Combo" or for each feature in the Relation Reference of `t_basket`. And they can add new ones by the "Dataset Manager". Only thing they cannot do (at the moment) are multiple baskets per topic-dataset combination, what would be technical possible but AFAIK not used in production.

QgisModelBaker/libqgsprojectgen/dbconnector/db_connector.py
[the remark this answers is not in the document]
gacarrillor: You're right!

gacarrillor:
> In Model Baker we are always trying to find the balance between simplicity and power to make what power users need. Hiding complexity is always a good thing. Anyway, the `t_basket` column with the Relation Reference widget (using the tables) is provided so the user can select his basket/dataset on the project no matter if Model Baker is installed or not. But when they have Model Baker installed - default values are provided, so they do not need to struggle around with this field. See this the Dataset Collector Combo for that #520

That's right @signedav, that's a handy way of dealing with baskets. We have a functionality in which we totally hide even the term 'basket' for our users. It's a very specific use case as I mentioned above.

> What would make these changes "compatible" with your plugin? Or would you like to have kind of "modes"? It's always good to keep synergies - for both sides.

Agreed. Well, as I mentioned above, the `t_basket` field is something we don't want end users to deal with (for MB it makes sense, for our plugin it might not). Of course changing the default for such a field will impact others relying on it being hidden.

Also, changing the default for `create_basket_col` has of course consequences since it's not backwards compatible: it will break any version of our plugin that uses a new version of MB including these changes, because only a few of our functionalities deal with baskets and know how to handle them. There are solutions for this of course.

Since we don't want end users to deal with baskets, we also won't like them to have dataset/basket tables visible in their layer tree.

> Btw. How is the `t_basket` column filled in your plugin? Where the user sets the dataset/basket to use? Or is there one dataset used per project? 

We create one basket for each user that acts as a "receiver" (we actually create users via GUI as you can see [here](https://swisstierrascolombia.github.io/Asistente-LADM-COL/_static/captura_de_datos_en_campo/9_Coor_campo_asignacion_6.gif)). A coordinator allocates some parcels to each receiver (see [GIF](https://swisstierrascolombia.github.io/Asistente-LADM-COL/_static/captura_de_datos_en_campo/10_Coor_campo_asignacion_7.gif)), which leads to a subset of related data (based on parcels we allocate related plots, for instance) per receiver. Each subset is exported to a XTF (see [GIF](https://swisstierrascolombia.github.io/Asistente-LADM-COL/_static/captura_de_datos_en_campo/11_Coor_campo_asignacion_89.gif)) that the receiver loads and edits on QField. After their field work, receivers send their modified data to the coordinator, who updates his DB with changes made by every receiver (see [GIF](https://swisstierrascolombia.github.io/Asistente-LADM-COL/_static/captura_de_datos_en_campo/16_Coor_campo_sincronizacion_2.gif)). This is a general view of the use case. The plugin takes care of datasets/baskets under the hood. Neither the coordinator nor the receivers are aware of any "basket". Users choose which role they want to use our plugin with, which tells us which dataset we're going to work on (we have a dataset per role for this functionality).
 
> Yes. The thing is the following. A user has usually one basket per topic-dataset-combination. This means usually they want to create a dataset (and are not interested in any baskets at all). With the upcoming dataset manager the user should be able to create a new dataset (name it e.g. "Zurich Region") and then "Create baskets for all the topics" - so the dataset will be created by the user and the basket-table-entries automatically on request. That's why we decided against having the tables editable and want to make them as "untouchable" for the user as possible.

That's a good move. 

> Not sure if I understand you correctly. The users can choose the basket they are working on in the "Dataset Selector Combo" or for each feature in the Relation Reference of `t_basket`. And they can add new ones by the "Dataset Manager". Only thing they cannot do (at the moment) are multiple baskets per topic-dataset combination, what would be technical possible but AFAIK not used in production.

I'll certainly like to see this in action. As the Dataset Manager was only mentioned in the description, it was hard to figure out how it would work in the end. But this explanation sounds good.

Perhaps we could also take advantage of your implementation in other use cases. I could think about power users as well. However, the use case I explained to you earlier in this comment, will be difficult to be in line with this PR.

If you want we can discuss further via e-mail or a meeting.

gacarrillor:
> Okay. Thanks for the precision. I understand now that this PR brings an implementation that does not overlap with your specific basket implementation.

It does overlap because it brings new default behavior that we won't need (visible fields, expressions, tables to layer tree), but that's expected anyways because of the custom nature of our implementation. That's why I mentioned in the first comment that we'll have to evaluate how we're going to continue using MB, since we won't be able to continue depending on official releases. BTW, the recently added "toppings" were also an extra for us. Since then, I've been thinking on using MB just keeping the minimal set of classes (and functionality as well) that we regularly use, but this time included as a library in our plugin. 

As I said, we'll evaluate such an approach.

> Anyway, would it bring a benefit for your future use when `--createBasketCol` is disabled per default? Because I think this would be something we could handle to persist backwards compatibility (for Model Baker it can set it to true on first initialization or wherever).

Because of what I wrote above, I think we'll recommend MB only to power users, and we expect them to be aware of such new defaults. So, no changes are requested from my side to this PR.

Thanks @signedav.

signedav:
Thanks @gacarrillor 

Could you approve this PR? Since you already had a look at the code etc.

QgisModelBaker/libqgsprojectgen/dbconnector/mssql_connector.py
```diff
@@ -544,3 +548,15 @@ def ili_version(self):
...
+    def get_basket_handling(self):
+        if self.schema and self._table_exists(SETTINGS_TABLE):
+                cur = self.conn.cursor()
+                cur.execute("""SELECT setting
+                            FROM {schema}.{settings_table}
+                            WHERE tag = 'ch.ehi.ili2db.BasketHandling'
+                            """.format(schema=self.schema, settings_table=SETTINGS_TABLE))
+                content = cur.fetchone()
+                if content:
+                    return content[0] == 'readWrite'
```
m-kuhn, line 561: Where does `readWrite` come from?

QgisModelBaker/libqgsprojectgen/generator/generator.py
```diff
@@ -109,17 +110,18 @@ def layers(self, filter_layer_list=[]):
...
+            is_nmrel = record.get('kind_settings') == 'ASSOCIATION'
+            is_basket_table = record.get('tablename') == self._db_connector.basket_table_name
+            is_dataset_table = record.get('tablename') == self._db_connector.dataset_table_name
 
             alias = record['table_alias'] if 'table_alias' in record else None
             if not alias:
                 short_name = None
                 if is_domain and is_attribute:
-                    short_name = record['ili_name'].split('.')[-2] + '_' + record['ili_name'].split('.')[-1] if 'ili_name' in record else ''
+                    short_name = record['ili_name'].split('.')[-2] + '_' + record['ili_name'].split('.')[-1] if 'ili_name' in record and record['ili_name'] else ''
```
m-kuhn, line 124: Can we have a comment here explaining the rationale for this logic?

QgisModelBaker/libqgsprojectgen/dbconnector/mssql_connector.py
signedav: When `--createBasketCol` is set, this value is `readWrite`
I don't know if it ever diverts from that (besides of this column is not existing when `--createBasketCol` is not set).
Anyway I made a search in the ili2db code and I only find tests considering the `readWrite` value and no other.

So I think it's most stable to consider this value to get the information, if there is a basket handling or not...

m-kuhn (approved):
Looks good, some comments and suggestions about code complexity mostly.
Thanks a lot

gacarrillor (approved):
Well, I just skimmed the code and didn't actually test it via GUI. So, take my approval as a +1, but not as an exhaustive code review.

Other than that, I think what you've described/implemented makes a lot of sense for Model Baker. 

Great feature!

signedav:
Thanks guys :+1:

PR #520, opened 2021-08-05
Dataset Selector Combo

signedav:
Dataset Selector Combo to select the dataset to be filled in as default value (actually the basket_id of the selected dataset and the topic of the current layer). The content of the Dataset Selector Combo is loaded according to the layer's **_topic_** and **_datasource_**. The chosen `basket_tid` is stored in the project variables (`<host>_<db>_<schema>_<topic>` or `<filepath>_<topic>`) and accessed in the default expression of the layer's `basket_tid` field.

dataset_selector

- [ ] Maybe tests


QgisModelBaker/gui/panel/dataset_combobox.py
```diff
@@ -0,0 +1,166 @@
...
+            self.setEnabled(True)
+
+    def set_index(self, schema_topic_identificator):
+        current_basket_tid = QgsExpressionContextUtils.projectScope(QgsProject.instance()).variable(schema_topic_identificator)
+        matches = self.filtered_model.match(self.filtered_model.index(0, 0), int(DatasetSourceModel.Roles.BASKET_TID), current_basket_tid, 1, Qt.MatchExactly)
+        if matches:
+            self.setCurrentIndex(matches[0].row())
+
+    def make_schema_identificator(self, source_name, source):
+        if source_name == 'postgres' or source_name == 'mssql':
```
m-kuhn, line 156: Could we move this code to the connector?

QgisModelBaker/gui/panel/dataset_combobox.py
```diff
@@ -0,0 +1,166 @@
...
+        self.basket_model = DatasetSourceModel()
+        self.filtered_model = QSortFilterProxyModel()
+        self.filtered_model.setSourceModel(self.basket_model)
+        self.filtered_model.setFilterRole(int(DatasetSourceModel.Roles.SCHEMA_TOPIC_IDENTIFICATOR))
+        self.setModel(self.filtered_model)
+        self.setEnabled(False)
+
+    def set_current_layer(self, layer):
+        if self.isEnabled():
+            self.currentIndexChanged.disconnect(self.store_basket_tid)
```
m-kuhn, line 103: Will we ever need this combobox for another action than selecting the currently active dataset?
If yes it would be good to separate the code to create a (read only) combobox and the code to update the project variables.

m-kuhn (approved):

QgisModelBaker/gui/panel/dataset_combobox.py
signedav: Hm... we don't create this identification with information we got from the db (and it's connector functions) but from the layers and the data-sources instead. At this point we do not define any database configurations and connectors.

QgisModelBaker/gui/panel/dataset_combobox.py
signedav: You are thinking about the selection of the dataset on data import?
I thought about that, but this combobox diverts from it there. Though we select the datasets here it's actually a selection of the basket from the available baskets for the topic of the current layer. So I don't see much synergies I could use here.

QgisModelBaker/gui/panel/dataset_combobox.py
m-kuhn: I'd like to have all database specific code in one place. We can also put it into a utils file somewhere. If we ever add an oracle (or whatever) backend, we know where to change things.

QgisModelBaker/gui/panel/dataset_combobox.py
m-kuhn: Ok, let's put that away and refactor if we have real requirements.

QgisModelBaker/gui/panel/dataset_combobox.py
signedav: Good point. And our old teacher Mr. Vögelin would be happy about Oräcle ;-)

signedav: utils/db_utils?
I prefer to have it there instead of db_connector.

signedav: then we can put the only function in mssql_utils.py there as well..

signedav: Maybe not merging it with the mssql_utils since otherwise we require always `pyodb`.

QgisModelBaker/libqgsprojectgen/dataobjects/project.py
```diff
@@ -137,7 +137,7 @@ def create(self, path: str, qgis_project: QgsProject):
                     'ShowOpenFormButton': False,
                     'AllowNULL': True,
                     'AllowAddFeatures': False,
-                    'FilterExpression': "\"topic\" = '{}'".format( referencing_layer.model_topic_name ),
+                    'FilterExpression': "\"topic\" = @interlis_topic",
```
signedav, line 140: @m-kuhn This does not work. I think I know why. The expression is done on the referenced layer and the variable is from the current (referencing) layer. Do you have an idea or should we just keep the "hard" entered topic name there?

PR #521, opened 2021-08-11
Dataset Manager

signedav:
With the Dataset Manager the user is able to:
- Create new datasets
- Generate the baskets for each existing topic and this dataset
- Rename datasets
datsetmanager

**To Do:**
- [x] Fix troubles on GPKG "database locked" and layers are not updated
- [x] Tests

Thanks @zigertiger for the icon :-)

On GeoPackages we do not create the project with *transaction mode* anymore.

On starting the dataset manager it sets all the layers in the project to "not editable". If there are unsaved changes it's up to the user to save or discard or open the dataset manager with unsaved and editable layer. In the third case there is a warning box.


signedav:
@m-kuhn When have unsaved changes on a layer based on Geopackage, I cannot make mutations on the db since the "database is locked". After a mutation I cannot make changes on Geopackage since the "database is locked". I have an idea what the problem is but I don't know yet how to handle it. Any experience and ideas about it?

QgisModelBaker/gui/dataset_manager.py
```diff
@@ -0,0 +1,258 @@
...
+
+        self.reload_datasets(self.updated_configuration())
+
+        self.add_button.clicked.connect(self.add_dataset)
+        self.edit_button.clicked.connect(self.edit_dataset)
+        self.create_baskets_button.clicked.connect(self.create_baskets)
+        self.dataset_tableview.selectionModel().selectionChanged.connect( lambda: self.enable_dataset_handling(True))
+
+    def valid_selection(self):
+        return bool(len(self.dataset_tableview.selectedIndexes()))
```
m-kuhn, line 123:
```suggestion
    """
    Returns if at least one dataset is selected
    """
        return bool(self.dataset_tableview.selectedIndexes())
```

Should be enough

QgisModelBaker/gui/dataset_manager.py
```diff
@@ -0,0 +1,258 @@
...
+        def __int__(self):
+            return self.value
+
+    def __init__(self):
+        super().__init__()
+
+    def flags(self, index):
+        return Qt.ItemIsSelectable | Qt.ItemIsEnabled
+
+    def reload_datasets(self, db_connector):
```
m-kuhn, line 63:
```suggestion
    def _reload_datasets(self, db_connector):
```

This is mostly (only?) used internally, let's prefix it with `_` in this case

QgisModelBaker/gui/dataset_manager.py
```diff
@@ -0,0 +1,258 @@
...
+            value.setVisible(is_current_panel_selected)
+            if is_current_panel_selected:
+                value._show_panel()
+        self.reload_datasets(self.updated_configuration())
+
+    def request_for_refresh_datasets(self):
+        # hold refresh back
+        self.refreshTimer.start(500)
+
+    def refresh_datasets(self):
```
m-kuhn, line 149:
```suggestion
    """
    Refreshes the list of datasets.
    Do not call this directly, use the request_for_refresh_datasets() method instead.
    """
    def _refresh_datasets(self):
```

QgisModelBaker/gui/dataset_manager.py
```diff
@@ -0,0 +1,258 @@
...
+        if self.valid_selection():
+            db_connector = self.get_db_connector(self.updated_configuration())
+            if db_connector and db_connector.get_basket_handling:
+                feedbacks = []
+                for record in db_connector.get_topics_info():
+                    dataset_tid = self.dataset_tableview.selectedIndexes()[0].data(int(DatasetSourceModel.Roles.TID))
+                    status, message = db_connector.create_basket( dataset_tid, '.'.join([record['model'], record['topic']]))
+                    feedbacks.append((status, message))
+    
+                warning_box = QMessageBox(self)
```
m-kuhn, line 187: Warning or info?

QgisModelBaker/libqgsprojectgen/dbconnector/gpkg_connector.py
```diff
@@ -494,4 +496,91 @@ def get_basket_handling(self):
...
+            contents = cur.fetchall()
+            return contents
+        return {}
+        
+    def create_dataset(self, datasetname):
+        if self._table_exists(GPKG_DATASET_TABLE):
+            cur = self.conn.cursor()
+            try:
+                cur.execute("""
+                    INSERT INTO {dataset_table} (datasetName) VALUES ('{datasetname}')
```
m-kuhn, line 531: Does that work if I call my dataset `s'hertogenbosch`

QgisModelBaker/ui/dataset_manager.ui
```diff
@@ -0,0 +1,173 @@
...
+           <property name="icon">
+            <iconset>
+             <normaloff>../images/symbologyEdit.svg</normaloff>../images/symbologyEdit.svg</iconset>
+           </property>
+          </widget>
+         </item>
+         <item row="0" column="2">
+          <widget class="QToolButton" name="create_baskets_button">
+           <property name="text">
+            <string>Create baskets for selected dataset</string>
```
m-kuhn, line 146: Can we have a tooltip for this button that explains on a more technical level what happens?

m-kuhn (commented):
Nice.

- [ ] A small ui suggestion (tooltip)
- [ ] A bit of code style
- [ ] Some sql sanitizing (Probably to be integrated with techniques like https://www.psycopg.org/docs/sql.html#psycopg2.sql.SQL.format)

m-kuhn:
Does the "database locked" problem also happen when transaction mode is disabled?

signedav:
> Does the "database locked" problem also happen when transaction mode is disabled?

Hard to reproduce, since it's not always happening.

After a quick test the "database is locked" problem appeared without transaction mode as well.
_Created one feature - created a dataset - created two features - created a dataset - turn edit mode off. -> Problem appears._
It did not appear (yet - after quick tests) when I turn edit mode off before I make any dataset mutations.

m-kuhn:
> It did not appear (yet - after quick tests) when I turn edit mode off before I make any dataset mutations.

In this case we could simply disable the interface if edit mode is on?

signedav:
> In this case we could simply disable the interface if edit mode is on?

Might resolve it when transactionmode is disabled. But I think we cannot simply disable transaction mode?

QgisModelBaker/gui/dataset_manager.py
signedav: Ok. It doesn't even need the `bool` - but I'm not sure what is the best practice here. Considering "explicit is better than implicit" and "simple is better than complex" I like the version you suggested.

QgisModelBaker/libqgsprojectgen/dbconnector/gpkg_connector.py
signedav: No

QgisModelBaker/gui/dataset_manager.py
signedav: I just accepted the change, but I wasn't reading well. I thought it's about the refresh_datasets method. But this here is not called internally. It's called in `DatasetManagerDialog`... So I would not make it private.

m-kuhn:
transaction mode with gpkg has more problems, we might realy disable it for now

signedav:
@m-kuhn I considered your inputs. Thanks for having another look.

m-kuhn (approved):

PR #523, opened 2021-08-16
Add tabstops in pg settings panel

webrian:
Add tabstops in the PostGIS settings panel to ease and/or speed up navigation with the keyboard


m-kuhn (approved):

m-kuhn:
Thanks !

PR #524, opened 2021-08-18
Create multiple relation editor widgets for same relation

### LINKED ISSUE #505: Create multiple relation editor widgets for same relation
Considering this implementation https://github.com/qgis/QGIS/pull/38331 and this issue https://github.com/qgis/QGIS/issues/37675

domi4484:
Create multiple relation editor widgets for same relation close #505

Notes:
- For many to many relations the cardinality is set to the the linked parent table
- If the link table has additional attributes that are not foreign keys (Associations with attributes), an additional relation editor widget pointing to the link table is added

### DIFF
--- a/QgisModelBaker/libqgsprojectgen/dataobjects/form.py
+++ b/QgisModelBaker/libqgsprojectgen/dataobjects/form.py
@@ -110,8 +110,9 @@ class FormFieldWidget(object):
 
 class FormRelationWidget(object):
 
-    def __init__(self, relation):
+    def __init__(self, relation, nmRelation=None):
         self.relation = relation
+        self.nmRelation = nmRelation
 
     def create(self, parent, layer):
         try:
@@ -121,4 +122,6 @@ class FormRelationWidget(object):
             # Feed deprecated API for 3.0.0 and 3.0.1
             widget = QgsAttributeEditorRelation(
                self.relation.id, self.relation.id, parent)
+        if self.nmRelation:
+            widget.setNmRelationId(self.nmRelation.id)
         return widget
--- a/QgisModelBaker/libqgsprojectgen/dataobjects/form.py
+++ b/QgisModelBaker/libqgsprojectgen/dataobjects/form.py
@@ -110,9 +110,9 @@ class FormFieldWidget(object):
 
 class FormRelationWidget(object):
 
-    def __init__(self, relation, nmRelation=None):
+    def __init__(self, relation, nm_relation=None):
         self.relation = relation
-        self.nmRelation = nmRelation
+        self.nm_relation = nm_relation
 
     def create(self, parent, layer):
         try:
@@ -122,6 +122,6 @@ class FormRelationWidget(object):
             # Feed deprecated API for 3.0.0 and 3.0.1
             widget = QgsAttributeEditorRelation(
                self.relation.id, self.relation.id, parent)
-        if self.nmRelation:
-            widget.setNmRelationId(self.nmRelation.id)
+        if self.nm_relation:
+            widget.setNmRelationId(self.nm_relation.id)
         return widget
--- a/QgisModelBaker/libqgsprojectgen/dataobjects/layers.py
+++ b/QgisModelBaker/libqgsprojectgen/dataobjects/layers.py
@@ -17,6 +17,7 @@
  *                                                                         *
  ***************************************************************************/
 """
+import logging
 from .form import (
     Form,
     FormTab,
@@ -24,6 +25,7 @@ from .form import (
     FormFieldWidget
 )
 from qgis.core import (
+    Qgis,
     QgsVectorLayer,
     QgsDataSourceUri,
     QgsWkbTypes,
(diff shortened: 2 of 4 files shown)


QgisModelBaker/libqgsprojectgen/dataobjects/layers.py
```diff
@@ -186,3 +212,27 @@ def real_id(self):
             return self.__layer.id()
         else:
             return None
+
+    def isPureLinkTable(self, project):
```
domi4484, line 216: @signedav I am not 100% sure about that. It looks if all fields (except one, the id) of a layer are referencing fields to decide if the link table has extra attributes or not. Do you think there could be more fields than the Id which are not extra fields? Maybe like composite keys or so?

QgisModelBaker/libqgsprojectgen/dataobjects/form.py
```diff
@@ -110,8 +110,9 @@ def create(self, parent, layer):
 
 class FormRelationWidget(object):
 
-    def __init__(self, relation):
+    def __init__(self, relation, nmRelation=None):
         self.relation = relation
+        self.nmRelation = nmRelation
```
signedav, line 115: Can you avoid camelcase in variable names https://www.python.org/dev/peps/pep-0008/#function-and-variable-names ?

QgisModelBaker/libqgsprojectgen/dataobjects/layers.py
```diff
@@ -186,3 +212,27 @@ def real_id(self):
...
+        '''
+        Returns True if the layer is a pure link table in a n:m relation.
+        With "pure" it is meant the layer has no more fields than foreign keys and its id.
+        '''
+
+        # Check if table is a link table
+        if self.is_nmrel is False:
```
signedav, line 223: I usually prefer `if not self.is_nmrel` - but not sure about any convention about it.

QgisModelBaker/libqgsprojectgen/dataobjects/layers.py
signedav: Maybe yes. Not sure. Maybe some investigations could help. Are there for example not only `t_id` but `t_ili_tid` as well? We might have a look at it together if you want.

signedav (commented):
Haven't tested it but left you some comments.

As far as I understood you, you plan to add some test?

QgisModelBaker/libqgsprojectgen/dataobjects/form.py
domi4484: Yes, it was the habit from c++

QgisModelBaker/libqgsprojectgen/dataobjects/layers.py
domi4484: Again habit from c++ where I don't like writing `!is_nmrel` because the `!` is easily overseen. But the `not` is two more characters so I can accept it ;)

QgisModelBaker/libqgsprojectgen/dataobjects/layers.py
domi4484: I have changed the implementation in a way that `t_id` and `t_ili_tid` are considered.
With the ili2db `--t_id_Name` parameter it is possible to specify a custom id column name. So i will have to look at that to.

QgisModelBaker/libqgsprojectgen/dataobjects/layers.py
```diff
@@ -231,8 +231,11 @@ def isPureLinkTable(self, project):
...
             if relation.referencing_layer != self:
                 continue
-            remaining_fields.remove(relation.referencing_field)
+            remaining_fields.discard(relation.referencing_field)
 
-        # Only Id field left?
-        return len(remaining_fields) <= 1
+        # Remove Id fields
+        remaining_fields.discard('t_id')
+        remaining_fields.discard('t_ili_tid')
```
signedav, line 238: In geopackage this value is "T_Ili_Tid" but you can get this value from the current db_connector.

domi4484: ok thanks

signedav:
This PR breaks the tests checking the tabs on the generated projects. You need to correct them. And it would be cool to have a new test to check this implementation here.

domi4484:
> This PR breaks the tests checking the tabs on the generated projects. You need to correct them. And it would be cool to have a new test to check this implementation here.

Yes I have seen, the tests are coming with the next commit. But to correct the tests it would be very handy to get more outputs from them. It say with parameter `-vv` we can get the full output but I don't know how to pass this parameter

domi4484:
@signedav https://github.com/opengisch/QgisModelBaker/blob/1e3dbd8a844d1fe3f53a55dddd98e1cf947d522e/.docker/run-docker-tests.sh#L37 here we can pass more parameters and select only a part of tests to execute like this `xvfb-run pytest -vv -k test_kbs_postgis`

signedav:
Another thing.

On using baskets every table has a `t_basket` FK pointing to the `t_ili2db_basket` table.
Now when we have a many-to-many association between e.g. `city` and `citicen` there will be created a link table with three FKs (to `city`, `citicen` and `t_ili2db_basket`). I think with the current sollution in `city` will be two relation references (direct to `citicen` and to `t_ili2db_basket`). The relation editor to `t_ili2db_basket` should be avoided.

QgisModelBaker/libqgsprojectgen/dataobjects/layers.py
```diff
@@ -167,21 +164,32 @@ def post_generate(self, project):
 
             self.__form.add_element(tab)
 
-            relations_to_add = set()
+            relations_to_add = []
```
signedav, line 167: Thanks for this elegant solution. Out of curiosity, what has been your thought about not using a set anymore and make a list instead?

signedav (approved):

signedav:
Thanks @domi4484

QgisModelBaker/libqgsprojectgen/dataobjects/layers.py
domi4484: Just to keep ordering constant over multiple generates, because set is unordered and could potentially get to different tab order each time.

PR #525, opened 2021-08-19
fix docker compose names

signedav:
Fixed it in `/scripts/run_tests.sh` to be able to use this to test locally.
I wonder if the ci folder is used at all still. Do you know about it @m-kuhn ? Like the deploy part etc?


domi4484 (approved):
Works for me

m-kuhn:
> Fixed it in `/scripts/run_tests.sh` to be able to use this to test locally.
> I wonder if the ci folder is used at all still. Do you know about it @m-kuhn ? Like the deploy part etc?

Seems that it's only referenced from within the `ci` scripts itself. Good to remove for me. Also anything that references travis.

Issue #533, opened 2021-09-03
Problem with get_topic


signedav:
Structs in no Topis are read as topic - leads to problems with the baskets.

Reproducable with DM_asset model or with this model:
```
INTERLIS 2.3;

MODEL NoTopicStruct (en) 
AT "https://signedav.github.io/usabilitydave/models"
VERSION "2020-06-22" =
    STRUCTURE Address =
        Name: TEXT;
        Street: TEXT;
    END Address;

    TOPIC TopicA =
        CLASS Person =
            Address: Address;
            Remark : TEXT*254;
        END Person;
    END TopicA; !! of TOPIC

END NoTopicStruct. !! of MODEL
```

What is exported (without any basket id handling) like this:
```
<NoTopicStruct.TopicA BID="NoTopicStruct.TopicA">
<NoTopicStruct.TopicA.Person TID="1"><Address><NoTopicStruct.Address><Name>dave</Name><Street>tal</Street></NoTopicStruct.Address></Address><Remark>test</Remark></NoTopicStruct.TopicA.Person>
</NoTopicStruct.TopicA>
```

signedav:
But the get topic problem is only one part of the problem.

The question is how we set the basket id of the address in the example above. Because basket id columns are created for it, but they should contain the basket of it's parent. Or no baskets? Hm...

signedav:
Main problem is that the topics are not parsed correctly and baskets are created that lead to errors. -> **Solvable by better topic parsing function - there should be just no basket created for non-topic-structs**

Second problem is, that it's hard to handle structures without a topic. Because their entries don't belong to an own basket. They to belong to the parent's basket.

**Some tests:**
1. create schema without baskets
2. make data and export
3. create schema with baskets and import (and pass `--dataset Winterthur`) -> works and the `Address` entries have the basket of `TopicA`

1. create schema with baskets
2. remove wrong basket
3. make data using the same basket in `address` like in `parent` (of `TopicA` and export (without the wrong baskets in the databas) -> works

signedav:
This is an issue with the `KbS_Basis_V1_3` `Parzellenidentifikation`e.g.

PR #536, opened 2021-09-08
Fix Topic parsing

signedav:
The parsing was not stable when there are STRUCTURES in the INTERLIS Model that are not in a Topic.

This fixes the basic problem of #533

Not yet solved is being able to choose the correct basket for the STRUCTURE that could be related from classes from several topics.


signedav:
Ps. Not sure if there are more elegant statements in SQLite and MSSQL, but I think they are solid...

domi4484 (approved):

PR #542, opened 2021-09-10
Check validity of emitted db connection parameter from layer

signedav:
And do not try to connect when they are not valid. E.g. when the layer bases on a pg service, this check won't be valid.


QgisModelBaker/utils/db_utils.py
```diff
@@ -37,6 +37,7 @@ def get_schema_identificator(layer_source_name, layer_source):
 
 def get_configuration(layer_source_name, layer_source):
```
m-kuhn, line 38: Can you add a docstring about the return value here?

m-kuhn (approved):

PR #547, opened 2021-09-29
Fix error when trying to connect to an XTF datasource

signedav:
When connecting to a layer based on an XTF file (what returns as datasourcename OGR), the get db connector failed. This is caught by the try.


domi4484 (approved):

signedav:
Thanks @domi4484

PR #549, opened 2021-10-01
Allow source file (transfer or ili) drop

signedav:
On dropping in QGIS -> Wizard Source Selection Page is opened and added to the source model.
On dropping in Wizard Source Selection Page source view -> added to the source model.
drop_it

- [ ] Adjust size of drop dialog

### DIFF
--- a/QgisModelBaker/gui/options.py
+++ b/QgisModelBaker/gui/options.py
@@ -17,6 +17,7 @@
  *                                                                         *
  ***************************************************************************/
 """
+import pathlib
 import webbrowser
 
 from qgis.PyQt.QtCore import QLocale, QModelIndex, QSettings, Qt, pyqtSignal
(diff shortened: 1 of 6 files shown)


QgisModelBaker/gui/options.py
```diff
@@ -203,3 +204,26 @@ def keyPressEvent(self, e):
...
+
+    files_dropped = pyqtSignal(list)
+
+    def __init__(self, parent=None):
+        super(FileDropListView, self).__init__(parent)
+        self.setAcceptDrops(True)
+        self.setDragDropMode(QListView.InternalMove)
+
+    def dragEnterEvent(self, event):
+        event.acceptProposedAction()
```
domi4484, line 219: Should not this be accepted only if entered files are `in ["xtf", "XTF", "itf", "ITF", "ili"]` ?

domi4484: Yes, but if you would reject the event, the mouse cursor changes and let the user now that his files are not droppable in the widget.

QgisModelBaker/gui/options.py
```diff
@@ -203,3 +204,26 @@ def keyPressEvent(self, e):
...
+
+    def dropEvent(self, event):
+        dropped_files = [
+            url.toLocalFile()
+            for url in event.mimeData().urls()
+            if pathlib.Path(url.toLocalFile()).suffix[1:]
+            in ["xtf", "XTF", "itf", "ITF", "ili"]
```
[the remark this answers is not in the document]
domi4484: ok I didn't know about it, then makes sense how it is

signedav:
drop
Thanks @domi4484 Considered your input.

Btw. when at least one file fits the requirements, we accept (and ignore the others)...

domi4484 (approved):

Issue #529, opened 2021-08-31
Optimization for XTF model parser


gacarrillor:
Some time ago at SwissTierras Colombia, we implemented a XTF model parser (see [our code](https://github.com/SwissTierrasColombia/Asistente-LADM-COL/blob/master/asistente_ladm_col/utils/interlis_utils.py)) as the one you guys implemented recently (this is [your code](https://github.com/opengisch/QgisModelBaker/blob/master/QgisModelBaker/gui/workflow_wizard/wizard_tools.py#L306-L322)).

While they both perform well for short XTF file sizes, there is a difference for medium and big file sizes. See this time table:

| XTF File size | STC | Opengis.ch |
| :---------: | :--: | :----------: |
| 4.5MB     | 0.01007986068725586s | 0.1192464828491211s |
| 20MB      | 0.02394556999206543s | 0.4195380210876465s |
| 40MB      | 0.02762746810913086s | 0.899895191192627s |
| 450MB     | 0.043598175048828125s | 10.097153663635254s |
| 1.6GB     | 0.030645132064819336s | 58.79246234893799s |

So, this is just to let you know that you could borrow our method to parse models from an XTF file if you want.

-------
_This is the code I've employed for the comparison: https://gist.github.com/gacarrillor/e873e377a81bd91c31dbaa49a4f2f2a7_

signedav:
Thanks a lot @gacarrillor I'll have a look.

signedav:
Out of curiosity what difference makes the smart check of your function and what does the cElementTree I tested on an xtf file 2.2 MB:
- STC with ElementTree
- STC with cElementTree
- Opengis.ch with ElementTree
- Opengis.ch with cElementTree

**Result:**
- Option Opengis.ch, execution time in seconds: 0.05835247039794922
- C Option Opengis.ch, execution time in seconds: 0.07306528091430664
- Option STC, execution time in seconds: 0.0011992454528808594
- C Option STC, execution time in seconds: 0.000579833984375

Proofs your implementation as better but still, do you have an idea why my parser takes longer with `cElementTree` than `ElementTree`?

PR #555, opened 2021-10-07
Smarter parsing for models in xtf

signedav:
Use of more performant xtf parsing function.

No multicheck of same model.

Dont't consider models parsed from xtf as models from a file.

References #529


gacarrillor (approved):
Approving changes to `_transfer_file_models()`. 

(The other changes in `wizard_tools.py` seem to deal with another (related) use case, but I have no strong opinion on them.)

signedav:
Thanks

Issue #564, opened 2021-10-16
Wizard Dialog don't stays on top in MS Windows


signedav:
On Windows systems we have the problem that the wizard does not stay in front of QGIS, like this is expected with the other Model Baker windows. On the intro page it stays in front, even when clicking on the QGIS window, but as soon as pressed "next" this ability disappears and the wizard dialog is behind the QGIS.

PR #565, opened 2021-10-16
Fix strange behavior of dialog order

### LINKED ISSUE #560: 6.6.2: Workflow Wizard always displayed in front
The window of the workflow wizard is always in front and all other windows are hidden behind. Even windows from other softwares. At least on windows. That's a bit wierd...

workflow_wizard

### LINKED ISSUE #562: t_ili_id generated even if "Import Tid erzeugen" is deselected
When the option  "Import Tid erzeugen" is deselected, there shouldn't be any t_ili_id-column in the generated schema. But it's still there. Seems as the option don't change anything.

image

Is that a bug or am I missing something?

signedav:
resolves #560

I reverted the stay-on-top flag on the windows. This leads to issue #564 but this is minor the #560


m-kuhn:
Would setting the parent of the dialogs to `iface.mainWindow()` work?

signedav:
The `QWizardPages`? I can try it. Still I merge that to have a okay running beta soon.

PR #569, opened 2021-10-19
Fix weird dialog order in MS Windows

### LINKED ISSUE #564: Wizard Dialog don't stays on top in MS Windows

signedav:
Having sub dialog on top.

image

fixes #564

Thanks for the input @m-kuhn


sjib:
@signedav Much better like this! I can click in other windows and come back and it does not disappear anymore when I run the wizard.

signedav:
Cool. Thsnks for the feedback.

Issue #484, opened 2021-05-07
Topping considered selectable catalogue integration


signedav:
There is an idea that we could list all the catalogues defined in the selected metaconfiguration and give the user the possibilty to select or unselect it.

Maybe this is a good idea - maybe it's not used and just leads to confusion and faults.

signedav:
With the linked integrations we have:
- possibility to link catalogues in the metaconfig file
- possibility to link catalogues in the ilidata.xml and are provided in the data import page (add/remove)
- 
IMHO this solves this issue.

PR #583, opened 2021-11-04
Get catalogues and their models over UsabILIty Hub

### LINKED ISSUE #485: Using ilidata.xml to list possible catalogues on data import
An idea is on data import to get all the catalogues (unfiltered) from the ilidata.xml file in a dropdown to import them to a database.
It's siimilar to the model import with ilimodels.xml like we have it on the schema import.

signedav:
1. On entering the _Import Schema Configuration Page_ the ilidata.xml on the UsabILIty Hub repositories are parsed for the modelnames (from the listed models) and the type `referenceData` in the `categories`. There the `modelLink` properties are read and the models are loaded into the list of models to import automatically.

2. On the _Import Data Configuration Page_ we have an autocomplete dropdown containing the `referenceData` items (transfer files) found according all the models in the current database schema (not only the previously imported ones). You can add and remove them by +/-. `xml` files receive the default property that they are catalogues (considering the datasets).

usability_cat

Resolves #485

3. Fixes that a metaconfiguration / referenceData in the ilidata.xml only could have been found by one model. Now there can be multiple models listed in the categories.

**ToDo:**
- [x] Parse and append Linked Models
- [x] Fix appended Linked Model update
- [x] Parse and provide Reference Data
- [x] Remove Button on Data Import Page
- [x] Refresh of data import view
- [x] Go to import data page in case there is the possibilty to get transfer files from repo
- [x] Add info to model list that it's loaded because of dependency
- [x] CLEAN UP! Might make sense to have some utils to use synergies in the completer widget and the download of files etc. (_on_metaconfig_received etc)

*Optional:*
- Only provide transfer files where linked model is already installed
- Being able to start without model import - and having provided all the referenceData concerning the models in the database
- Maybe checkbox to "add all depending transfer files"


signedav:
The tooltips provide the information why this model is now suddenly here:
image

QgisModelBaker/gui/workflow_wizard/import_data_configuration_page.py
```diff
@@ -166,37 +226,177 @@ def order_list(self):
...
+            self._set_basket_defaults()
         else:
             self.file_table_view.setColumnHidden(1, True)
             self.file_table_view.setColumnHidden(2, True)
             self.datasetmanager_button.setHidden(True)
+        self._update_referencedata_completer()
+
+    def _set_basket_defaults(self):
+        for row in range(self.workflow_wizard.source_model.rowCount()):
+            index = self.workflow_wizard.source_model.index(row, 2)
```
domi4484, line 244: I would use an enum instead of literal number `2` for the column

QgisModelBaker/gui/workflow_wizard/import_data_configuration_page.py
```diff
@@ -166,37 +226,177 @@ def order_list(self):
...
+            value = index.data(int(wizard_tools.SourceModel.Roles.DATASET_NAME))
+            if not value:
+                self.workflow_wizard.source_model.setData(
+                    index,
+                    wizard_tools.DEFAULT_DATASETNAME,
+                    int(wizard_tools.SourceModel.Roles.DATASET_NAME),
+                )
+                is_xml = (
+                    self.workflow_wizard.source_model.index(row, 0)
```
domi4484, line 253: same here

QgisModelBaker/gui/workflow_wizard/import_data_configuration_page.py
```diff
@@ -166,37 +226,177 @@ def order_list(self):
...
+                    .data(int(wizard_tools.SourceModel.Roles.TYPE))
+                    .lower()
+                    == "xml"
+                )
+                self.workflow_wizard.source_model.setData(
+                    self.workflow_wizard.source_model.index(row, 1),
```
domi4484, line 259: and here

QgisModelBaker/gui/workflow_wizard/workflow_wizard.py
```diff
@@ -448,7 +466,37 @@ def get_topping_file_model(self, id_list, log_panel):
...
+        self.ilireferencedatacache.refresh()
+
+    def _db_modelnames(self, configuration):
+        db_connector = db_utils.get_db_connector(configuration)
+        modelnames = list()
+        if db_connector:
+            if db_connector.db_or_schema_exists() and db_connector.metadata_exists():
+                db_models = db_connector.get_models()
+                for db_model in db_models:
+                    regex = re.compile(r"(?:\{[^\}]*\}|\s)")
```
domi4484, line 494: In Qt regex are heavy to build, so i would put it before the `for` loop and reuse it. Not sure if it is the same in python tought...

domi4484 (approved):

signedav:
I think this resolves #484 It's not the same, but replaces it.

domi4484 (approved):

signedav:
Minor changes are done since your approve @domi4484 So I will merge when tests pass. Thanks.

PR #584, opened 2021-11-05
Fix MultilingualText / LocalisedText Element for title in the ilicache

signedav:
In http://models.interlis.ch/core/DatasetIdx16.ili the attribute title is of type `MultilingualText` and not `MultilingualMText`.

What would happen when the fixed ModelBaker parses unfixed ilidata.xml or vice versa? - It would take the id as description. So no problem for me to merge that even when people have invalid ilidata.xml.

usabilityhub_notitle


signedav:
https://github.com/opengisch/usabilityhub/pull/10

andreasneumann:
Thanks! I don't need to submit an issue for this if you already fixed it!

domi4484 (approved):

PR #586, opened 2021-11-08
check if item is a dict 

signedav:
to avoid crash when user made mistakes in his yaml tree.

### DIFF
--- a/QgisModelBaker/libqgsprojectgen/generator/generator.py
+++ b/QgisModelBaker/libqgsprojectgen/generator/generator.py
@@ -462,7 +462,7 @@ class Generator(QObject):
 
     def full_node(self, layers, item):
         current_node = None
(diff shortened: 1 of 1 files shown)


QgisModelBaker/libqgsprojectgen/generator/generator.py
```diff
@@ -462,7 +462,7 @@ def relations(self, layers, filter_layer_list=[]):
 
     def full_node(self, layers, item):
         current_node = None
-        if item:
+        if isinstance(item, dict):
```
domi4484, line 465: I think an empty dictionary will return `True`, is this expected?

signedav: You are right. It should never be empty but still better to check it.

domi4484 (approved):

PR #588, opened 2021-11-11
Avoid crash when text of element cannot be found

signedav:
Parsing the `ilimodel.xml` / `ilidata.xml` we sometimes need to check the found element if it's empty before we access it's text. Mostly we trusted that the data has been correctly setup and the elements are there, we are looking for, but in case this is not the case Model Baker crashed. With this helper function `get_element_text` we even can remove some checks for element and make the code slimmer.

### DIFF
--- a/QgisModelBaker/libili2db/ilicache.py
+++ b/QgisModelBaker/libili2db/ilicache.py
@@ -236,7 +236,9 @@ class IliCache(QObject):
                     repo_models.append(model)
 
         self.repositories[netloc] = sorted(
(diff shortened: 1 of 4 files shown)


signedav:
It avoids all the 
```
value = element.find("value").text if element.find("value") else None
```
or 
```
value = None
value_elem = element.find("value")
if value_elem:
  value = value_elem.text
```

m-kuhn (approved):
Looks good to me.
If possible, can you follow up with a test?

PR #596, opened 2021-11-18
Validate

signedav:
Implementation of https://gist.github.com/signedav/c84b1ea0acf9ae16a794b8fd2013abd3
The database connection parameter are emitted from the currently selected layer. Mostly this is representative for the whole project, since mostly a project bases on one single DB schema. In case of multiple used DB Schemas, it's possible to "switch" between the validation results when switching the layers.
valid4

Entries of the type `Error` and `Warning` are listed. With right click menu you can "Zoom to coordinates" (if coordinates are provided) and "Open Feature Form" (if t_ili_tid is provided).

With right click on the error there is opening a menu with:
- Zoom to coordinates (if coordinates are available)
- Open form (if a stable id is available)
- Set to fixed (marking the entry green to have an order in the fixing process)

- [x] selectable model/dataset/basket
- [x] find feature by tilitid (open form or zoom to)
- [x] when no tilitid then geom
- [x] Clean up
- [x] Testing
- [x] set layer editable when form opens (I don't do that. When the layer is active it's better the user sets to editable himself)
- [x] activate layer when form opens
- [x] keep current filter mode
- [x] unselect entry when clicking to the empty space at the end of the list
- [x] status bar turning green when okay (optional)
- [x] entry point zoom to / open form (right click vs. button)
- [x] mark that it's fixed (not sure about checkboxes)
- [x] bug with selecting all models after having no filter after filter change (and with this the list view is invisible) - when setting then to visible (by choosing models/datasets/baskets) the selecting all x works but the list view gets not updated. 

**Follow ups - after merge:**
- the export page of the wizard should use the `FilterDataPanel`
- the `ValidationResultModel` should go to `utils/gui_utils.py` - as well as all the wizard_tools (except the pages) especially the stuff used by the validator. Additionally there can be checked what could go into other utils moduls (like eg. `get_basket_handling`)

**Optional Improvements:***
- show log
- indicator square on geometry coordinates
- evtl. use object tag to find layer of geometry error


gacarrillor:
Nice David. I think we've been working on similar functionality. See for example this commit: https://github.com/SwissTierrasColombia/Asistente-LADM-COL/pull/512/commits/387db2dc0d2ced2a24e08057feae1495df8ed93f

domi4484 (approved):

domi4484 (approved):
Code looks good to me in general, but how can i access this functionality in qgis? I looked in the "view" and "model baker" menus searching the new panel

signedav:
It's view -> panel -> Model Baker Data Validator

domi4484:
> Edit: Might make sence to have it in Database -> Model Baker menue as well...

I would look there in first place

signedav:
Added it to the menu.

Should it be in the toolbar as well? 
image
I'm -1 to have it in the toolbar, since it's not something like the other dialogs we open and close again. It's something we activate and we have it open for quite some time. But I'm open to be convinced by another opinion.

QgisModelBaker/qgismodelbaker.py
```diff
@@ -412,6 +436,41 @@ def show_about_dialog(self):
...
+            settings.value(
+                "QgisModelBaker/validateDockArea", Qt.RightDockWidgetArea, type=int
+            ),
+            self.__validate_dock,
+        )
+        self.__validate_dock.visibilityChanged.connect(
+            self.__validate_action.setChecked
+        )
+        self.__validate_dock.setVisible(
+            settings.value("QgisModelBaker/dockWidgetIsVisible", False, type=bool)
```
domi4484, line 452: All validate dock widget settings could be in the same group. This one in particular would be misleading if in future we get a separate dock widget for another functionality.

signedav: Good point. Will consider that.

signedav: Will name it to `ValidationDockIsVisible`

domi4484: Or:
- `ValidationDock/IsVisible`
- `ValidationDock/Area`
So they are grouped in the settings viewer

signedav: Right.

QgisModelBaker/qgismodelbaker.py
```diff
@@ -458,11 +458,11 @@ def init_validate_dock(self):
     def remove_validate_dock(self):
         settings = QSettings()
         settings.setValue(
-            "QgisModelBaker/validateDockArea",
+            "QgisModelBaker//validate_dock/area",
```
domi4484, line 461:
```suggestion
            "QgisModelBaker/validate_dock/area",
```

signedav: Arg. Thanks :-)

signedav:
Thanks @domi4484

PR #603, opened 2021-12-03
Fix the importTid/exportTid handling in Wizard

### LINKED ISSUE #601: Create Import Tid Configuration
image(3)

The tooltip sais:
> "Creates a new column t_ili_tid in all classes, which is mandatory and should be unique, not only for across the current database, but across the whole system. Normally you should not enable this option directly, but rely on the INTERLIS model and ili2db. The former should have a line enabling OIDs (preferably with UUIDs) and the latter will make sure the database updates the new column every time a new record is created.

So one might think it's about to create the `t_ili_tid` column in the tables. But in the code we have this:

image(2)

The setting "Create Import Tid" decides if `--importTid` is set in `SchemaImportConfiguration`. That's odd since it's not relevant for the Schema Import (not like `--createTidCol` what is set anyway). 

`--importTid` is considering the data import. Like `--exportTid` in the export (but this is not set by the Model Baker.

(There is another variable `with_importtid` and `with_exporttid` controlling the those parameters. I don't know why.)

So it seems that this setting in the configuration is considered in data import only. The user might think that it changes something on schema create but it doesn't.

There are several possibilities to fix that:
a. make the function control `--createTidCol` instead on schema import and `--importTid` at data import (what is not that nice, because then one checkbox would control different parameters in different cases)
b. rename the function to "Import Tid" instead (what is not that nice, when it's in the option dialog that appears for schema import).

Generally the questions are:
1. Should `--createTidCol` be configurable in Model Baker (or always set)?
2. Should `--importTid` / `--exportTid` be configurable in Model Baker (or always set when `--createTidCol` is set)?

I have the feeling that the 1. should be "yes" and 2. should be "no". But this would change the current behavior.

In the wizard I have no settings for data import. So in case we would reply 2 wiht "yes" it might need another setting in data import. And otherwise it needs to  set `--importTid` statically \*

\* same for export with `--exportTid`

signedav:
In the wizard, we get the `TidHandling` from settings table to decide if `--importTid` / `--exportTid` needs to be passed on data import/export. `TidHandling` is in the `t_ili2db_settings` table when `--createTidCol` is passed on schema import. This means always when it's created by Model Baker. 

This fixes #601 and keeps the behavior of the single dialogues how it is (a little confusing but as we are used to).

### DIFF
--- a/QgisModelBaker/libqgsprojectgen/dbconnector/db_connector.py
+++ b/QgisModelBaker/libqgsprojectgen/dbconnector/db_connector.py
@@ -312,6 +312,14 @@ class DBConnector(QObject):
         """
         return False, None
 
+    def get_tid_handling(self):
+        """
+        Returns if there is a tid handling enabled according to the settings table.
+        It's the tid handling that has been enabled "manually" (or by a plugin) and
+        not if t_ili_tids are used according the stable id defined in the model.
+        """
+        return False
+
 
 class DBConnectorError(Exception):
     """This error is raised when DbConnector could not connect to database.
--- a/QgisModelBaker/libqgsprojectgen/dbconnector/db_connector.py
+++ b/QgisModelBaker/libqgsprojectgen/dbconnector/db_connector.py
@@ -251,7 +251,8 @@ class DBConnector(QObject):
 
     def get_basket_handling(self):
         """
-        Returns if there is a basket handling enabled according to the settings table
+        Returns `True` if a basket handling is enabled according to the settings table.
+        Means when the database has been created with `--createBasketCol`.
         """
         return False
 
@@ -314,9 +315,8 @@ class DBConnector(QObject):
 
     def get_tid_handling(self):
         """
-        Returns if there is a tid handling enabled according to the settings table.
-        It's the tid handling that has been enabled "manually" (or by a plugin) and
-        not if t_ili_tids are used according the stable id defined in the model.
+        Returns `True` if a tid handling is enabled according to the settings table (when the database has been created with `--createTidCol`).
+        If t_ili_tids are used only because of a stable id definition in the model (with `OID as` in the topic or the class definition), this parameter is not set and this function will return `False`.
         """
         return False
 
--- a/QgisModelBaker/libqgsprojectgen/dbconnector/mssql_connector.py
+++ b/QgisModelBaker/libqgsprojectgen/dbconnector/mssql_connector.py
@@ -664,10 +664,11 @@ WHERE TABLE_SCHEMA='{schema}'
             cur.execute(
                 """SELECT setting
                             FROM {schema}.{table}
-                            WHERE tag = 'ch.ehi.ili2db.BasketHandling'
+                            WHERE tag = ?
                             """.format(
                     schema=self.schema, table=SETTINGS_TABLE
-                )
+                ),
+                ("ch.ehi.ili2db.BasketHandling",),
             )
             content = cur.fetchone()
             if content:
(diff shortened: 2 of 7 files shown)


QgisModelBaker/libqgsprojectgen/dbconnector/db_connector.py
```diff
@@ -312,6 +312,14 @@ def create_basket(self, dataset_tid, topic):
         """
         return False, None
 
+    def get_tid_handling(self):
+        """
+        Returns if there is a tid handling enabled according to the settings table.
+        It's the tid handling that has been enabled "manually" (or by a plugin) and
+        not if t_ili_tids are used according the stable id defined in the model.
```
m-kuhn, line 319: Can this be rewritten a bit to leave no room for interpretation.
Something like

"Will return ``True`` if ``--createTidCol`` has been used to create the database.
... more information about model definition OID ...
"

please proof read

QgisModelBaker/libqgsprojectgen/dbconnector/mssql_connector.py
```diff
@@ -818,3 +818,19 @@ def create_basket(self, dataset_tid, topic):
...
+
+    def get_tid_handling(self):
+        if self.schema and self._table_exists(SETTINGS_TABLE):
+            cur = self.conn.cursor()
+            cur.execute(
+                """SELECT setting
+                            FROM {schema}.{table}
+                            WHERE tag = 'ch.ehi.ili2db.TidHandling'
+                            """.format(
+                    schema=self.schema, table=SETTINGS_TABLE
```
m-kuhn, line 830: Can we parametrize the query (or at least quote identifiers) https://stackoverflow.com/questions/62396036/is-there-a-difference-between-quote-ident-and-parametized-queries-in-psycopg?

m-kuhn (commented):
Looks good.

QgisModelBaker/libqgsprojectgen/dbconnector/mssql_connector.py
m-kuhn: I think the most important part is where variables are used, i.e. schema and table in this case.

signedav: Agree. I will address this. But only in the current function. I opened an issue #604 I will go over all the other functions but not in this PR since small typos can lead to bigger errors and this PR should be mainly a fix for the upcoming stable release. Ok?

signedav: Still not sure what you mean. You think the using of `{schema}.{table}` is bad?

signedav: like this?

```
           cur.execute(
                sql.SQL( """SELECT setting
                           FROM {}.{}
                           WHERE tag = %s
                        """).format(sql.Identifier(self.schema),sql.Identifier(PG_SETTINGS_TABLE)), ("ch.ehi.ili2db.TidHandling",)
            )

```

m-kuhn: > Still not sure what you mean. You think the using of {schema}.{table} is bad? 

Exactly, e.g. if schema name or table name contains spaces or other special characters the query will fail.

What you proposed looks good.
Or:

```
cur.execute( 
"""SELECT setting
                           FROM {schema}.{table}
                           WHERE tag = %s
                        """.format(schema=ext.quote_ident(schema), table=ext.quote_ident(table))
)
```

signedav: We have sollutions on `psycopg2` - do you know what we can use for `sqlite3` and `pyodbc`?

m-kuhn: A quick search for sqlite3 didn't reveal any core functionality (some helper methods like https://pretagteam.com/question/sqlite3-escapestring-failing-multiple-single-quotes-generated)
For pyodbc, maybe @gacarrillor has an idea?

signedav: Can we move the solution finding here #604 and have this PR "in dry"? :-)

m-kuhn: Yes, that sounds good to me (and was the reason for only commenting on the PG implementation)

PR #607, opened 2021-12-09
Use List  widget for ARRAY mapping

domi4484:
If the metaattribute `!!@ili2db.mapping=ARRAY` is defined in the ili model configure the corresponding widget in QGIS as "List"

### DIFF
--- a/QgisModelBaker/libqgsprojectgen/dbconnector/pg_connector.py
+++ b/QgisModelBaker/libqgsprojectgen/dbconnector/pg_connector.py
@@ -486,6 +486,41 @@ class PGConnector(DBConnector):
 
         return {}
 
+    def get_value_mapping_info(self, ili_name, field_name):
+
+        if not self._table_exists(PG_METAATTRS_TABLE):
+            return None
+
+        if not self.schema:
+            return None
+
+        cur = self.conn.cursor(cursor_factory=psycopg2.extras.DictCursor)
+        cur.execute(
+            """
+                    SELECT
+                      attr_value
+                    FROM {schema}.{metaattrs_table}
+                    WHERE LOWER(ilielement)=LOWER('{ili_name}.{field_name}')
+                    AND attr_name='ili2db.mapping'
+                    LIMIT 1;
+            """.format(
+                schema=self.schema,
+                metaattrs_table=PG_METAATTRS_TABLE,
+                ili_name=ili_name,
+                field_name=field_name
+            )
+        )
+
+        records = cur.fetchall()
+        cur.close()
+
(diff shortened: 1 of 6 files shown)


QgisModelBaker/libqgsprojectgen/dbconnector/pg_connector.py
```diff
@@ -486,6 +486,41 @@ def get_min_max_info(self, table_name):
...
+        if not self.schema:
+            return None
+
+        cur = self.conn.cursor(cursor_factory=psycopg2.extras.DictCursor)
+        cur.execute(
+            """
+                    SELECT
+                      attr_value
+                    FROM {schema}.{metaattrs_table}
+                    WHERE LOWER(ilielement)=LOWER('{ili_name}.{field_name}')
```
domi4484, line 503: @signedav i think there should be a better variable than `field_name`, can you suggest me? (field_name is lowercase but in the model/db is not)

QgisModelBaker/libqgsprojectgen/generator/generator.py
```diff
@@ -303,6 +303,9 @@ def layers(self, filter_layer_list=[]):
                         {val: val} for val in value_map_info[column_name]
                     ]
 
+                if self.get_value_mapping_info(record["ili_name"], column_name ) == "ARRAY":
```
domi4484, line 306: and in fact here is called `column_name` and not even `field_name`...

domi4484 (commented):
@signedav can you take a look at this draft please? Mssql is missing, will add it after the first feedback

QgisModelBaker/libqgsprojectgen/dbconnector/pg_connector.py
signedav: I think `column_name` is fine.

Another approach would be to look up the ili-name of the attribute by the table `t_ili2db_attrname` or even return it by `get_fields_info` accessing this table already - you could do it analogue to `get_value_map` or am I overseeing something?

signedav: Btw. the name `get_value_mapping_info` i find confusing. Wouldn't it be better like `get_list_info` or similar?

domi4484: thanks I removed completely `get_value_mapping_info` and return the mappings information with `get_fields_info`

signedav (approved):
Looks good to me now.

domi4484:
@signedav all tests passed, good to merge for me

signedav:
Cool. Thanks.

gacarrillor:
Thank you @domi4484 and @signedav!

PR #612, opened 2021-12-15
Bump QGIS test versions

### LINKED ISSUE #611: Replace qgis 3.10 with 3.22 tests
Let's stick to supported versions

signedav:
...and let em run...

Resolves #611

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -17,7 +17,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


domi4484 (approved):
In the code (mostly tests) there are checks for `Qgis.QGIS_VERSION_INT` being smaller than 3.16 or even 3.5. Should this be cleaned up? Just asking but i don't see any big disadvantage keeping them.

signedav:
Good point @domi4484 A general clean up is planned where I will consider it. Smaller 3.16 I would keep for a while. Smaller 3.5 I think can be cleaned up then. Another thing is ili2db 3 - I will open an issue for that.

PR #627, opened 2022-01-10
Cast type from calculated value to int 

### LINKED ISSUE #626: unable to install from qgis plugin manager
Couldn't load plugin 'QgisModelBaker' due to an error when calling its initGui() method 

TypeError: setMaximumHeight(self, int): argument 1 has unexpected type 'float' 
  File "/usr/lib/python3.10/site-packages/qgis/utils.py", line 448, in startPlugin
    plugins[packageName].initGui()
  File "/home/sliss/.local/share/QGIS/QGIS3/profiles/default/python/plugins/QgisModelBaker/qgismodelbaker.py", line 245, in initGui
    self.init_validate_dock()
  File "/home/sliss/.local/share/QGIS/QGIS3/profiles/default/python/plugins/QgisModelBaker/qgismodelbaker.py", line 441, in init_validate_dock
    self.__validate_dock = ValidateDock(self.ili2db_configuration, self.iface)
  File "/home/sliss/.local/share/QGIS/QGIS3/profiles/default/python/plugins/QgisModelBaker/gui/validate.py", line 253, in __init__
    self.filter_data_panel.setMaximumHeight(self.height() / 4)
TypeError: setMaximumHeight(self, int): argument 1 has unexpected type 'float'

Python version: 3.10.1 (main, Dec 18 2021, 23:53:45) [GCC 11.1.0] 
QGIS version: 3.22.2-Białowieża Białowieża, exported 

Python Path:
/usr/share/qgis/python
/home/sliss/.local/share/QGIS/QGIS3/profiles/default/python
/home/sliss/.local/share/QGIS/QGIS3/profiles/default/python/plugins
/usr/share/qgis/python/plugins
/usr/lib/python310.zip
/usr/lib/python3.10
/usr/lib/python3.10/lib-dynload
/usr/lib/python3.10/site-packages
/home/sliss/.local/share/QGIS/QGIS3/profiles/default/python
/home/sliss/.local/share/QGIS/QGIS3/profiles/default/python/plugins/HCMGIS/forms
/home/sliss/.local/share/QGIS/QGIS3/profiles/default/python/plugins/qgisgbifapi/vendor
/home/sliss/.local/share/QGIS/QGIS3/profiles/default/python/plugins/qgisgbifapi/vendor
/home/sliss/.local/share/QGIS/QGIS3/profiles/default/python/plugins/postgisQueryBuilder
/home/sliss/.local/share/QGIS/QGIS3/profiles/default/python/plugins/postgisQueryBuilder/extlibs

signedav:
Before passing it to `QWidget.setMaximumHeight`.

Fixes #626

### DIFF
--- a/QgisModelBaker/gui/validate.py
+++ b/QgisModelBaker/gui/validate.py
@@ -250,7 +250,7 @@ class ValidateDock(QDockWidget, DIALOG_UI):
         self.current_filter_mode = SchemaDataFilterMode.NO_FILTER
 
         self.filter_data_panel = FilterDataPanel(self)
-        self.filter_data_panel.setMaximumHeight(150)
+        self.filter_data_panel.setMaximumHeight(self.fontMetrics().lineSpacing() * 10)
         self.filter_layout.addWidget(self.filter_data_panel)
         self._reset_gui()
 
--- a/QgisModelBaker/gui/workflow_wizard/workflow_wizard.py
+++ b/QgisModelBaker/gui/workflow_wizard/workflow_wizard.py
@@ -527,8 +527,8 @@ class WorkflowWizardDialog(QDialog):
         self.workflow_wizard = WorkflowWizard(self.iface, self.base_config, self)
         self.workflow_wizard.setStartId(PageIds.Intro)
         self.workflow_wizard.setWindowFlags(Qt.Widget)
-        self.workflow_wizard.setFixedHeight(600)
-        self.workflow_wizard.setMinimumWidth(800)
+        self.workflow_wizard.setFixedHeight(self.fontMetrics().lineSpacing() * 36)
+        self.workflow_wizard.setMinimumWidth(self.fontMetrics().lineSpacing() * 48)
         self.workflow_wizard.show()
 
         self.workflow_wizard.finished.connect(self.done)


signedav:
I don't really know why this issue happened and I cannot reproduce it neither. But this fixes it hopefully.

m-kuhn:
If the dock height is not a multpile of 4, the result will be a float. The dock height might be influenced by the geometryrestorer too?

In either case, is this the right place to insert this? The maximumHeight is calculated with the initial height of the widget and is then unchanged.
The following screenshot is taken when it's initialized with a very small height, it doesn't take the available screen space for the widgets even if resized after.
image

Should this be a/ recalculated on widget size change or b/ be set to take as much space as needed but not more instead?

I agree that this patch will very likely fix the issue.

signedav:
> If the dock height is not a multpile of 4, the result will be a float. The dock height might be influenced by the geometryrestorer too?

Yes. I got that. But I did not experienced that a float is not implicit converted to an int. But I know it now:
```
<stdin>:1: DeprecationWarning: an integer is required (got type float).  Implicit conversion to integers using __int__ is deprecated, and may be removed in a future version of Python.
```
(with python 3.8.10)

> In either case, is this the right place to insert this? The maximumHeight is calculated with the initial height of the widget and is then unchanged.
Should this be a/ recalculated on widget size change or b/ be set to take as much space as needed but not more instead?

Well I don't want to have the filtering part of the GUI too big and keep at least 3/4 for the result of the validation. A minimum size of the filtering part would make sense but I don't see another way as calculating it out from it's content, and the content has no fixed size (since the list widget of models/baskets) is stretched. I noticed your case, but then I thought that it's rarely the case that someone opens it with such a small QGIS main window size.

m-kuhn:
The situation was that the panel started at the bottom right corner, next to where the python console is often located.
I wonder if a different sizepolicy or some other flags would be better for this. Maybe @domi4484 has an idea?
Otherwise it's good to merge also

m-kuhn (approved):

domi4484:
Another approach to set this kind of proportions is to use layout stretch: [https://doc.qt.io/qt-5/layout.html#stretch-factors](https://doc.qt.io/qt-5/layout.html#stretch-factors)

signedav:
Stretch factor is a good input, but don't we have the problem with the stretch factor that this part of the form (filter panel) is then always stretched to the factor e.g. 25% even if the child elements need less space?

I want to have it (filter panel) only the space it needs and give the rest to the other widgets / layouts. But if it would take more than 1/4 of the complete panel (because the list view has lots of entries) it should only stretch to the maximum size of 1/4 of the main panel and the list view should turn scrollable. That's what this implementation does.

But yes. The current solution is not that nice. On initialization of the plugin it sets this `maximumHeight` and it does not evaluate how big the panel will become when it's stretched - it takes the height that is defined here https://github.com/opengisch/QgisModelBaker/blob/master/QgisModelBaker/ui/validator.ui#L10 so it's always 175. 

To make the `maximumHeight` depending on the current panel size I guess we need to implement kind of a sizechanged connection of the main panel and reset the the `maximumHeight. But this might be a little an overkill, since we are quite happy with the `maximumHeight` of unintended 175.

I'm a little confused. When we define a size of 175 will it always have the space for title, dropdown, listview with 4 items? Main question is, is a label on every system the same pixel size?

m-kuhn:
> I'm a little confused. When we define a size of 175 will it always have the space for title, dropdown, listview with 4 items? Main question is, is a label on every system the same pixel size?

No, it depends on screen resolution, system font size ...

What could be done is using QFontMetrics height to calculate this

domi4484:
> Stretch factor is a good input, but don't we have the problem with the stretch factor that this part of the form (filter panel) is then always stretched to the factor e.g. 25% even if the child elements need less space?

I think this will still depend by the size policy of the widget. If it is `expanding` it will probably always try to get the maximum allowed by the stretch proportions. 

> 
> I want to have it (filter panel) only the space it needs and give the rest to the other widgets / layouts. But if it would take more than 1/4 of the complete panel (because the list view has lots of entries) it should only stretch to the maximum size of 1/4 of the main panel and the list view should turn scrollable. That's what this implementation does.

With layout stretch should be possible to achieve this behaviour

> 
> But yes. The current solution is not that nice. On initialization of the plugin it sets this `maximumHeight` and it does not evaluate how big the panel will become when it's stretched - it takes the height that is defined here https://github.com/opengisch/QgisModelBaker/blob/master/QgisModelBaker/ui/validator.ui#L10 so it's always 175.
> 
> To make the `maximumHeight` depending on the current panel size I guess we need to implement kind of a sizechanged connection of the main panel and reset the the `maximumHeight`. But this might be a little an overkill, since we are quite happy with the `maximumHeight` of unintended 175.

Connect to size changed for me it also sound doable, i don't think it is a big overkill

> 
> I'm a little confused. When we define a size of 175 will it always have the space for title, dropdown, listview with 4 items? Main question is, is a label on every system the same pixel size?

Label size will probably depend on font type and size, so will not be always the same

signedav:
> What could be done is using QFontMetrics height to calculate this.

Okay yes. I think this is the bit I needed. Why this is not always done / why I've never seen it? Maybe because it's mostly not a big difference between the systems and the GUIs looks mostly okay even if there is sometimes too much / too little space...

> With layout stretch should be possible to achieve this behaviour
@domi4484 Sure? Isn't the stretch factor applied to the `height` (and not the `maximumHeight`) - because this widget here has parts turning invisible - it should apply the size to it.
filter_panel

Thinking about it. Is there a big reason not just to use 200 px as the maximum size for this widget part? 
It does not need to be bigger. Even when the main window is way bigger (more than 800). I agree that the signal to change maximumHeight won't be big deal, but still I think it's not needed here.

signedav:
I did it now statically to 150 maximum height. I know that it's the lazy way but I think it's not worth the effort to make it nicer since it does not need a specific height, it just needs to be not too big in the end. And I think that usually everything fits in this height that needs to be there.

I decided against the stretch approach since this is only possible when I add the GUI parts one by one programmatically afaik.

m-kuhn:
For reference, that's equivalent to 3 line edits on my hidpi screen, does that sound reasonable?

signedav:
It sounds not reasonable that you have a way better screen than me :-)

Seriously 3 line edits is the minimum. But don't you have any other trouble with QGIS Plugin GUIs? I've seen fixed size quite for some time. And how does the Model Baker Wizard look at your resolution?

signedav:
And anyway - before it should have been less. What does it look with the current version?

m-kuhn:
image

signedav:
Makes sense. Now it would be a bit bigger but not that much. Let me see.

If you can give me a screenshot of the wizard would be great.

m-kuhn:
Tried the patch (150 fixed) and it's roughly the same

signedav:
Yeah. It's only a little larger. I think I need to solve it differently.

m-kuhn:
Peek 2022-01-14 15-40

signedav:
Yeah. That's what I expected.

signedav:
Fixed with `QFontMetrics` in the validator and in the wizard.

**Why not stretch in the validator?**
On a quick test with `setRowStretch` I experienced that it stretches always to the maximum, because it has this expandable list view (even when it's invisible)

```
        self.filter_data_panel = FilterDataPanel(self)
        self.filter_data_panel.setSizePolicy(QSizePolicy.Preferred, QSizePolicy.Preferred)
        self.validation_layout.addWidget(self.filter_data_panel)
        self.validation_layout.setRowStretch(0,1)
        self.validation_layout.setRowStretch(1,2)
        self.validation_layout.setRowStretch(2,1)
```

Leads to:
| no filter     |     baskets  |
|-----------|--------------------|
| stretch_nofilter | stretch_baskets|

So I decided to go for the FontMetric approach. This seems straight forward and stable for me.

**Why we need a fix size in the wizard?**
Because otherwise it changes the size on every page according to the content. We have expandable content, fix content that is way smaller and even generically added multiple widgets (like in the session pages on having multiple ili2db sessions).

m-kuhn (approved):

PR #641, opened 2022-02-25
Plugin implementation of basket handling wiht library

signedav:
After making library independent #640 It depends on that PR, so don't merge before.
We tell the Generator to act according the basket handling created by ili2db. And we set the default parameter.


QgisModelBaker/libqgsprojectgen/utils/qgis_utils.py
```diff
@@ -106,3 +106,12 @@ def get_layer_type(layer):
         return "raster"
     else:
         return "unknown"
+
+
+def get_group_non_recursive(group, group_name):
+    groups = group.findGroups(False)
```
signedav, line 112:
```suggestion
    groups = (
        group.findGroups(False)
        if Qgis.QGIS_VERSION_INT >= 31800
        else group.findGroups()
    )
```

signedav:
@domi4484 It might be hard to follow. It's mainly about releasing the library part from custom GUI / UI stuff I did in the other PR and now fix the GUI. I would appreciate just a code review for any small coding mistakes. You don't have to dig too much. Thanks.

domi4484 (approved):

signedav:
Thanks

PR #643, opened 2022-03-02
Integration of Linking Relation Editor

domi4484:


signedav (approved):

signedav:
Still let them test first...

signedav:
And maybe it would make sense model baker uses it already?

domi4484:
> And maybe it would make sense model baker uses it already?

Yes, I am just having some troubles to replace the widget type with the new one. I am doing this [here](https://github.com/opengisch/QgisModelBaker/blob/8c007cb1c3fbb525399deeb76afe0e6c66759182/QgisModelBaker/libqgsprojectgen/dataobjects/form.py#L119) with `widget.setRelationWidgetTypeId("linking_relation_editor")` but when generating the project the information get lost.
I have do dig into it

signedav:
Is it possible to make the plugin dependency depending on the QGIS version? Since I suppose linking relation editor does not work for 3.16 (and maybe nor for 3.22)?

domi4484:
> Is it possible to make the plugin dependency depending on the QGIS version?

From what I have seen it is not possible


> Since I suppose linking relation editor does not work for 3.16 (and maybe nor for 3.22)?

It works from version 3.18.

domi4484:
@signedav Installing ModelBaker on QGIS 3.16 would show the following window, but the plugin can still be installed:
image

signedav:
If you press "cancel"? Well, fair enough.

signedav:
Can you confirm that when the project is created with >3.16 and the Linking Relation Editor is chosen as widget, I get a proper Relation Editor Widget when I open it on 3.16 (with or without having Model Baker installed)?

domi4484:
> If you press "cancel"? Well, fair enough.

It makes no difference, the plugin get also installed successfully 🤔 Should this be considered a qgis bug?




> Can you confirm that when the project is created with >3.16 and the Linking Relation Editor is chosen as widget, I get a proper Relation Editor Widget when I open it on 3.16 (with or without having Model Baker installed)?

I can't tell you at the moment because I don't even get it to use the new widget for >3.16. Still have to find out why, this is how I was trying:


> Yes, I am just having some troubles to replace the widget type with the new one. I am doing this [here](https://github.com/opengisch/QgisModelBaker/blob/8c007cb1c3fbb525399deeb76afe0e6c66759182/QgisModelBaker/libqgsprojectgen/dataobjects/form.py#L119) with `widget.setRelationWidgetTypeId("linking_relation_editor")` but when generating the project the information get lost.
> I have do dig into it

signedav:
> Should this be considered a qgis bug?

Hm... little confusing that okay and cancel does the same. Maybe a "minor" bug. But what should be the behavior? Cancel cancels the installation? Or cancel cancels the dependency?

> I can't tell you at the moment because I don't even get it to use the new widget for >3.16. Still have to find out why, this is how I was trying

Okay, just check it before we publish.

signedav:
After your fix, do you know about that @domi4484 ?
> Can you confirm that when the project is created with >3.16 and the Linking Relation Editor is chosen as widget, I get a proper Relation Editor Widget when I open it on 3.16 (with or without having Model Baker installed)?

domi4484:
> After your fix, do you know about that @domi4484 ?
> 
> > Can you confirm that when the project is created with >3.16 and the Linking Relation Editor is chosen as widget, I get a proper Relation Editor Widget when I open it on 3.16 (with or without having Model Baker installed)?

@signedav yes I can confirm it.

signedav:
Is it still wip?

domi4484:
No sorry, i removed the draft label

signedav:
To merge after succesful tests (besides autoflake)

signedav:
Good to merge?

signedav:
Welcome Linking Relation Editor :wave:

Thanks @domi4484

PR #647, opened 2022-03-23
Pg service test + last bits of sslmode implementation

domi4484:


QgisModelBaker/gui/panel/pg_config_panel.py
```diff
@@ -20,7 +20,7 @@
 from enum import IntEnum
 
 try:
-    import QgisModelBaker.libs.pgserviceparser
+    import QgisModelBaker.libs.pgserviceparser as pgserviceparser
```
domi4484, line 23: @signedav not sure why the tests were failing locally without this... does it make sense for you?

signedav: locally are not using this import right?

signedav (approved):

QgisModelBaker/gui/panel/pg_config_panel.py
domi4484: Thanks for the explanation :)

signedav: This should work as well

```
from QgisModelBaker.libs import pgserviceparser
```

PR #652, opened 2022-03-24
do not add duplicate modelnames because otherwise they are listed mul…

signedav:
…tiple times in validator/exporter etc. when one model imports another

### DIFF
--- a/QgisModelBaker/utils/gui_utils.py
+++ b/QgisModelBaker/utils/gui_utils.py
@@ -770,8 +770,13 @@ class SchemaModelsModel(CheckEntriesModel):
                 regex = re.compile(r"(?:\{[^\}]*\}|\s)")
                 for db_model in db_models:
                     for modelname in regex.split(db_model["modelname"]):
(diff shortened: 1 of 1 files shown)


QgisModelBaker/utils/gui_utils.py
```diff
@@ -770,8 +770,13 @@ def refresh_model(self, db_connector=None):
                 regex = re.compile(r"(?:\{[^\}]*\}|\s)")
                 for db_model in db_models:
                     for modelname in regex.split(db_model["modelname"]):
-                        if modelname and modelname not in TRANSFERFILE_MODELS_BLACKLIST:
-                            modelnames.append(modelname.strip())
+                        name = modelname.strip()
+                        if (
+                            name
+                            and name not in TRANSFERFILE_MODELS_BLACKLIST
+                            and name not in modelnames
```
domi4484, line 777: You could use a set. But maybe all other functions need a list?

domi4484 (approved):

PR #658, opened 2022-03-25
Model Baker Logo and Branding

signedav:
Having a supernice postmodern logo now.
Long-Logo_Green_Modelbaker_RGB_QGIS

Put all the sources (including fonts) on the repo.

Used icon in Plugin Repository and in Documentation.


signedav:
@romefi I tended to use the long logo, since in most desktop usage, it's better from the size. But still both are official.

Btw. the fonts will be available in the branding folder as well. Maybe that's nice to use in presentations.

domi4484 (approved):
very nice the new branding!

signedav:
Yes. It's from https://www.danina.org/

PR #669, opened 2022-04-13
set write rights to lib folder 

signedav:
because qgis-plugin-ci needs it for release


signedav:
@domi4484 merging that to go on with the release and since it won't influence the solution itself.

PR #687, opened 2022-04-27
Parse ITF files for models to be listen in the model import dialog

signedav:
And escape parsing on XTF when it goes over 100 lines without any results

### DIFF
--- a/QgisModelBaker/utils/gui_utils.py
+++ b/QgisModelBaker/utils/gui_utils.py
@@ -1,3 +1,4 @@
+import mmap
 import os
 import pathlib
 import re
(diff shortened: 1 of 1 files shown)


QgisModelBaker/utils/gui_utils.py
```diff
@@ -490,9 +491,10 @@ def refresh_model(self, source_model, db_connector=None, silent=False):
 
     def _transfer_file_models(self, xtf_file_path):
```
domi4484, line 492:
```suggestion
    def _transfer_file_models(self, transfer_file_path):
```
The parameter should be renamed if it can be both an xtf or an itf file

domi4484 (approved):
Looks good to me. Where can I find an itf file? i can't remember how it looks like and wanted to take a look...
Could a test be added for the itf parsing?

PR #694, opened 2022-05-13
Fix defaultbasketvars

signedav:


domi4484 (approved):
I'am not 100% sure what the 1st point meands. But the code looks good to me

PR #719, opened 2022-09-01
Validator --skipGeometryCheck and --validConfig options

domi4484:
Depends on https://github.com/opengisch/QgisModelBakerLibrary/pull/20

image

### DIFF
--- a/QgisModelBaker/gui/validate.py
+++ b/QgisModelBaker/gui/validate.py
@@ -29,7 +29,7 @@ from qgis.core import (
 )
 from qgis.gui import QgsGui
 from qgis.PyQt.QtCore import QStandardPaths, Qt, QTimer
-from qgis.PyQt.QtWidgets import QAction, QDockWidget, QHeaderView, QMenu
+from qgis.PyQt.QtWidgets import QAction, QDockWidget, QFileDialog, QHeaderView, QMenu
 
 import QgisModelBaker.libs.modelbaker.utils.db_utils as db_utils
 from QgisModelBaker.gui.panel.filter_data_panel import FilterDataPanel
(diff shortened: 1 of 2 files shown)


signedav:
Nice. Can you make a file browser option for the config file? See existing implementation in ili2db_options.py

domi4484:
@signedav do you mean having the possibility to set it into this form? Or just so that the path is taken/saved to settings?

QgisModelBaker/gui/validate.py
```diff
@@ -464,3 +475,10 @@ def _get_feature_in_project(self, t_ili_tid):
                     if feature.attributes()[idx] == t_ili_tid:
                         return layer, feature
         return None, None
+
+    def _select_validator_config_file(self):
+        filename, _ = QFileDialog.getOpenFileName(
```
signedav, line 480: You can use here make_file_selector from qt_utils.py doing not much but having less lines of code here.

Anyway, would be nice to pass a filter on "ini" and "INI".

signedav (approved):

signedav:
> @signedav do you mean having the possibility to set it into this form? Or just so that the path is taken/saved to settings?

Ah sorry. I was confused by the tooltip in your screenshot. :-)

signedav:
Do you plan to append the configuration part here or in another PR?

QgisModelBaker/gui/validate.py
domi4484: Thanks I changed that as suggested

domi4484:
What do you mean with configuration part? this one? https://github.com/opengisch/QgisModelBakerLibrary/pull/20

signedav:
No the passing of the parameter. But I've overseen it. Fine. So this is good here.

signedav:
Merge when https://github.com/opengisch/QgisModelBakerLibrary/pull/20 is done.

signedav:
🎉

PR #720, opened 2022-09-01
Flash pan and zoom

domi4484:
Peek 2022-09-01 15-14


domi4484:
@signedav sorry is not based on master but #719 because I was getting conflicts with the .ui file :see_no_evil:

signedav (approved):

PR #733, opened 2022-09-19
Use functions with source provider instead of converted QgsDataSource

signedav:
Because QgsDataSource::uri does return unproper filepath when having spaces in the path.

Needs modelbaker 1.3.2


signedav:
needs new lib before test will work...

domi4484 (approved):

PR #740, opened 2022-10-24
only pan when feature has geometry (and zoom to feature)

signedav:
do flash coordinates and features geometry


sidhochwind:
Thanks

signedav:
Thanks

PR #741, opened 2022-10-26
Validation Improvements and Documentation

signedav:
- Zoom to feature if feature is found and otherwise to the coordinates with automatic zoom
- Show object tag (classname) in tooltip
- When finding the feature according the t_ili_tid, only consider the ones in the schema that is validated
- Update the documentation
- Info that diableAreaValidation is set on skip geometry error
[ ] bump to modelbaker 1.3.3 (depending on https://github.com/opengisch/QgisModelBakerLibrary/pull/32)


signedav:
No worries. We did not add the estimate that we update the documentation. So it's alright. We had some time left on this and the customer requested some explanation, so I did it here.

PR #742, opened 2022-11-02
unzip all whl files downloaded (means including dependancies) and bump modelbaker library

signedav:


signedav:
I merge it because it's only packaging stuff

PR #745, opened 2022-11-03
Decrease scale on zoom  to coordinates and action to open selection in attribute table

signedav:
1. Decreased the scale from 50 (map units) to 10 means the extend is then 20 map units on "Zoom to coordinates" - reason is a customer feedback that mostly a more detailed scale is needed.

2. Select in Attribute Table
grafik
To open an attribute table with the current feature selected:
grafik


signedav:
I think last changes in this implementation scope. Could you have a look @domi4484 ?

QgisModelBaker/gui/validate.py
```diff
@@ -521,6 +530,15 @@ def _open_form(self, t_ili_tid):
...
             self.iface.openFeatureForm(layer, feature, True)
 
+    def _select_feature_in_attributetable(self, t_ili_tid):
+        layer, feature = self._get_feature_in_project(t_ili_tid)
+        if layer and feature:
+            self.iface.layerTreeView().setCurrentLayer(layer)
+            layer.removeSelection()
+            layer.select(feature.id())
+            attribute_table = self.iface.showAttributeTable(layer)
+            attribute_table.findChild(QAction, "mActionSelectedFilter").trigger()
```
domi4484, line 540: It would be good to check if the returned widget is valid to avoid crashes if something changes upstream. For example mActionSelectedFilter renamed or so

domi4484 (approved):

QgisModelBaker/gui/validate.py
signedav: good input

PR #761, opened 2022-11-21
Session panel improvements (cancel java, re-run process)

### LINKED ISSUE #714: Cancel java process when closing Model Baker Wizard
When having a java process running (ili2db) and one closes the Model Baker Wizard window. The process is still running in the background.

It should cancel the java process or even ask the user "Do you want to cancel the process?"

### LINKED ISSUE #748: Export: Don't disable succeeded export session
grafik

On going back and forward again it's still disabled. Only when something changes it's reset. But maybe one wants to "Edit Command".

signedav:
This PR resolves three topics:
- [x] Cancel button appears while running and cancels the current process. Resolves #714
~- [x] Canceling all session when closing the wizard dialog. Resolves #714~
- [x] When a session is done or skipped it's possible to re-run them (they don't get disabled). Resolves #748

grafik

Requires https://github.com/opengisch/QgisModelBakerLibrary/pull/39

### DIFF
--- a/QgisModelBaker/gui/panel/session_panel.py
+++ b/QgisModelBaker/gui/panel/session_panel.py
@@ -49,7 +49,6 @@ class SessionPanel(QWidget, WIDGET_UI):
     on_process_finished = pyqtSignal(int, int)
     on_done_or_skipped = pyqtSignal(object, bool)
     cancel_session = pyqtSignal()
-    run_finished = pyqtSignal()
 
     def __init__(
         self,
@@ -136,6 +135,15 @@ class SessionPanel(QWidget, WIDGET_UI):
             self.configuration.baskets = self.baskets
 
         self.is_skipped_or_done = False
+
+    def hideEvent(self, event):
+        """
+        When widget gets hidden, we emit the cancel_session to terminate running java processes.
+        This is called on every hide (including the one right before getting destroyed because of dialog close.
+        """
+        self.print_info.emit(self.tr("Cancel session..."), LogColor.COLOR_INFO)
+        self.cancel_session.emit()
+        super().hideEvent(event)
 
     @property
     def id(self):
@@ -259,6 +267,7 @@ class SessionPanel(QWidget, WIDGET_UI):
             self.print_info.emit(self.tr("Cancel session..."), LogColor.COLOR_INFO)
             self.set_button_to_last_create_state()
             self.cancel_session.emit()
+            self.is_running = False
             return
         else:
             self.on_done_or_skipped.emit(self.id, False)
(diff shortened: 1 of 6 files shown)


QgisModelBaker/gui/panel/session_panel.py
```diff
@@ -136,6 +136,14 @@ def __init__(
 
         self.is_skipped_or_done = False
 
+    def hideEvent(self, event):
```
signedav, line 139: @m-kuhn Maybe you have an input here. We have a QWidget in a QWizardPage in a QWizard and this in a QDialog. When the dialog is closed we notice this only when this object here is destroyed (and then it's too late to do anything) or when it gets hidden. If it's hidden for any other reason (like clicking "back" in wizard") it makes sense as well that the process is terminated.

But is this save?

QgisModelBaker/gui/workflow_wizard/execution_page.py
```diff
@@ -39,6 +39,9 @@
 
 
 class ExecutionPage(QWizardPage, PAGE_UI):
+
+    cancel_current_session = pyqtSignal()
```
domi4484, line 43: This signal is never emitted

QgisModelBaker/gui/panel/session_panel.py
domi4484: Would it not be safer to handle this logic at QDialog closeEvent level? Calling there cancel_session on the relevant instance?
But i did not look very deeply so sorry if not mega relevant. But more generally speaking i prefer to take actions at the source of the event than relying on signal/slot connections.

signedav: I thought about that, but we do not receive a closeEvent on the QWidget and not on the QWizardPage and the QWizard. Only the QDialog. There the approach would be to override it and then go to the wizard / wizard pages / widgets to kill the process. Would be possible for sure, but not sure if it's better.

signedav: Hm... Maybe I need to follow another approach anyway. Since on terminating the process the function returns to the dialog that is no longer existent. I think the way to go is blocking the closing until everything is cleaned up.

QgisModelBaker/gui/workflow_wizard/execution_page.py
```diff
@@ -75,6 +75,21 @@ def __init__(self, parent, title, db_action_type):
         self.is_complete = False
         self.pending_sessions = []
 
+    def cancel_sessions(self):
```
signedav, line 78: @domi4484 Hey Mr. QTimer ;-) Any idea what I am doing wrong here?
Because I don't receive the signal `run_finished` before the timer expires. It's kind of blocked as long as the loop is executing...


And in general. What do you think about this approach with blocking `reject()` until the QProcesses are terminated?

domi4484: It is blocked because (I suppose) cancel_session is called by the main thread and the call to `loop.exec()` will block the main thread where the `timer` was created. Why are you unsing an event loop here? Just to block until canceling or timeout?
There are also other ways to wait without blocking the main thread

I think it is fine to block in the reject function as long as you display a hourglass to tell the user you are not hanging but doing a long operation.

signedav: Thanks. Yes I assume something like that. But the main thread should be blocked in fact. But be released on receiving `run_finished`. We use that on other places as well, but I think there we download a file with QNetworkRequest where it possibly creates a second thread...

signedav:
Canceling all session when closing the wizard dialog is not part of this PR. See https://github.com/opengisch/QgisModelBaker/issues/714#issuecomment-1357828478

Issue #747, opened 2022-11-03
Validator: Pan to coordinates


signedav:
On automatic pan it pans to the feature geometry centroid, if the OID is provided by ili2db, and if not **then no pan is performed (no pan on coordinates yet)**. On automatic zoom it zooms to the features geometry, if the OID is provided by ili2db, and if not then to the coordinates with the extend of 10 map units. (1)

The request is not only that it should pan to the coordinates when no OID available, but to do this as the preferred way. (2) Means pan to coordinates, if no coordinates then to feature geometry (centroid).

signedav:
1. Coord and OID
2. Only OID
3. Only Coord

**Status Quo:**
On automatic pan:
1. Pan to features centroid
2. Pan to features centroid
3. Nothing

On automatic zoom:
1. Zoom to features geometry
1. Zoom to features geometry
2. Zoom to coordinates with an extend of 10 map units


**Status Futura:**
On automatic pan:
1. Pan to coordinates
2. Pan to features centroid
3. Nothing

On automatic zoom:
1. **Zoom to features geometry or zoom to coordinates?**
1. Zoom to features geometry
2. Zoom to coordinates with an extend of 10 map units

kingemir:
While validating, centroids are actually neither of interest nor helpful. We anyway are using in practice extremely high zoom factors while validating, and a pan to centroid leads thus to complete desorientation since edges tend to fall out of the extension of the current map. 
If the error is related to the edges or vertices, we need to see them. In our daily business, these cases represent over 95% (most cases are duplicate coord).
If the error is not related to edges or vertices as in your example, it doesn‘t matter wether panning happens to any vertice or to the centroid. For helping to identify the object, one can keep the flash-up function activated.

signedav:
I see. Duplicate coord is in fact a common error. And the disadvantage of not pan to the coordinates in these cases is higher than the advantage of avoiding confusion when it pans to the "first" coordinate on a non-geometric error. I agree on that.

signedav:
What about the zoom? Should it prefer the coordinates there as well?

kingemir:
yes, also for zoom

signedav:
Okay. Maybe in future ili2db provides us kind of types we can take into account. But for the moment it will be done like this: #770

Thanks for your feedback @kingemir

PR #770, opened 2022-12-15
Validator: Pan and zoom to coordinate (whenever possible)

signedav:
With this change we pan to the coordinates. When no coordinate available, then we pan to the feature geometry (according to the provided OID). Same for zoom. On zooming to a feature geometry, we take it's extent.

1. Coord and OID available
2. Only OID available
3. Only Coord available

### Pan
**Status before:**
1. Pan to features centroid
2. Pan to features centroid
3. Nothing

**Implemented here*
1. Pan to coordinates
2. Pan to features
3. Pan to coordinates

Because we do receive error messages concerning text value but providing coordinates as well as geometry errors  providing OID, sometimes it can be confusing why it pans to the coordinates. See the discussion here #747

### Zoom
**Status before:**
1. Zoom to features
2. Zoom to features
3. Zoom to coordinates

**Implemented here*
1. Zoom to coordinates
2. Zoom to features
3. Zoom to coordinates

- [x] Pan to coord function
- [x] Priorize what should be taken first (coord, then OID feature geom)
- [x] Docu

### DIFF
--- a/docs/user_guide/validation.md
+++ b/docs/user_guide/validation.md
@@ -39,10 +39,10 @@ With *right click* on the error a menu is opened with the following options:
 - Copy (to copy the message text)
 
 Automatic pan, zoom and highlight features or coordinates are performed by clicking on the result tables entry.
+On automatic pan and zoom, the coordinates are taken into account if they are provided by ili2db, if not, then the geometry of the feature (according to the OID provided by ili2db). On automatic zoom on the features geometry, it's extent is taken and on coordinates an extend of 10 map units instead.
 
 !!! Note
-    On automatic pan it pans to the coordinates, if provided by ili2db, if not then to the features geometry (according to the OID provided by ili2db). On automatic zoom it zooms to the features geometry, if the OID is provided by ili2db, and if not then to the coordinates with the extend of 10 map units.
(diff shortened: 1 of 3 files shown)


docs/user_guide/validation.md
```diff
@@ -39,9 +39,10 @@ With *right click* on the error a menu is opened with the following options:
 - Copy (to copy the message text)
 
 Automatic pan, zoom and highlight features or coordinates are performed by clicking on the result tables entry.
+On automatic pan and zoom, the coordinates are taken into account if they are provided by ili2db, if not, then the geometry of the feature (according to the OID provided by ili2db). On automatic zoom on the features geometry, it's extent is taken and on coordinates an extend of 10 map units instead.
 
 !!! Note
-    On automatic pan it pans to the feature geometry centroid, if the OID is provided by ili2db, and if not then no pan is performed (no pan to coordinates yet). On automatic zoom it zooms to the features geometry, if the OID is provided by ili2db, and if not then to the coordinates with the extend of 10 map units.
+    Since ili2db sometimes on non-geometry errors provides the coordinates as well, there could be a confusion when it zooms or pans to the coordinates there. Still it's preverable to not zoom or pan to the coordinates on geometry errors when they provide an OID. Currently the validator cannot make a difference between those error-types.
```
domi4484, line 45:
```suggestion
    Since ili2db sometimes on non-geometry errors provides the coordinates as well, there could be a confusion when it zooms or pans to the coordinates there. Still it's preferable to not zoom or pan to the coordinates on geometry errors when they provide an OID. Currently the validator cannot make a difference between those error-types.
```

domi4484 (approved):

PR #780, opened 2023-02-01
pyupgrade

m-kuhn:
Modernize code


QgisModelBaker/utils/gui_utils.py
```diff
@@ -3,7 +3,7 @@
 import pathlib
 import re
 import warnings
-import xml.etree.cElementTree as CET
+import xml.etree.ElementTree as CET
```
signedav, line 6: Only thing I wonder is why we imported here something as CET (but it's not the cElementTree but the ElementTree but I guess it's a typo and since those libraries behave the same (maybe the speed is different) we did not notice..

signedav (approved):
Thanks. It's nice that we already had such modern code before :-)

Can you do it on https://github.com/opengisch/QgisModelBakerLibrary as well?

QgisModelBaker/utils/gui_utils.py
m-kuhn: Here is some context
https://github.com/opengisch/QgisModelBaker/issues/529#issuecomment-937519862

m-kuhn: And according to https://stackoverflow.com/a/50509403/2319028 the two are identical nowadays

signedav: I remembered my conclusion in this test with German. But fine - when they are identical nowadays, then I trust pyupgrade, that the cElementTree is what we should use.

.pre-commit-config.yaml
```diff
@@ -35,3 +35,8 @@ repos:
     rev: 22.3.0
     hooks:
       - id: black
+
+  - repo: https://github.com/asottile/pyupgrade
```
signedav, line 39:
```suggestion
  # Automatically upgrade syntax for newer versions
  - repo: https://github.com/asottile/pyupgrade
```

signedav:
Thanks @m-kuhn

signedav:
Thanks @m-kuhn

PR #800, opened 2023-04-12
Fix multiple evaluation of next-id in wizard and broken export models check-list

signedav:
### Improved log output on searching for reference data according to the model
Because this can take a while (when ilidata / ilisites is missing on a repo or local repo) and blocks the wizard. In future a solution to not block the wizard or make it more transparent could help, but this is not considered because of (maybe) upcoming redesign of the workflow.

### Fix the broken export models check-list
When going back on the export session page, the export config's export-model list has been broken because of multiple connected slots. So it checked (and immediately unchecked) the box etc.

### Fix that the nextId has been evaluated multiple times
`next_id` is called on the nextId overrides of the pages - so after the next-button is pressed. It finalizes the edits on the current page and returns the evaluated id of the next page.

But it's called multiple times. As well it's called on pressing "Back". It's not really clear why.
That's why it checks if the current page is the one the last current_id has been stored for, to be sure it only proceeds on pressing "Next".

### DIFF
--- a/QgisModelBaker/gui/workflow_wizard/workflow_wizard.py
+++ b/QgisModelBaker/gui/workflow_wizard/workflow_wizard.py
@@ -384,22 +384,24 @@ class WorkflowWizard(QWizard):
     def _current_page_title(self, id):
         if id == PageIds.ImportSourceSelection:
             return self.tr("Source Selection")
-        elif id == PageIds.ImportDatabaseSelection:
-            return self.tr("Database Configuration")
-        elif id == PageIds.GenerateDatabaseSelection:
+        elif (
+            id == PageIds.ImportDatabaseSelection
+            or id == PageIds.ExportDatabaseSelection
+            or id == PageIds.GenerateDatabaseSelection
+        ):
             return self.tr("Database Configuration")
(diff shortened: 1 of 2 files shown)


QgisModelBaker/gui/workflow_wizard/workflow_wizard.py
```diff
@@ -206,79 +206,109 @@ def sizeHint(self):
...
-                self._update_configurations(self.import_database_selection_page)
-                if self.refresh_import_models(True):
-                    # when there are models to import, we go to the configuration page for schema import
-                    return PageIds.ImportSchemaConfiguration
-                if self.import_data_file_model.rowCount():
-                    # when there are transfer files found, we go to the configuration page for data import
-                    return PageIds.ImportDataConfiguration
-                return PageIds.ProjectCreation
+        But it's called multiple times. As well it's called on pressing "Back". It's not really clear why.
+        That's why it checks if the current page is the one the last current_id has been stored for, to be sure it only proceeds on pressing "Next".
```
m-kuhn, line 214: That's more a line comment for line 217 and not api docs, right?

m-kuhn (approved):

PR #801, opened 2023-04-14
Fix broken export models gui on validator.

signedav:
This here https://github.com/opengisch/QgisModelBaker/pull/800/commits/21604c6159ccf1d572d233b7a0cf74562c09fb0e introduced the error that the model is not loaded according to the layer in the validator.

It needs to set the model new and it needs to disconnect and reconnect the sloths. Now it does re-set the model on setup dialog to be stable on validation gui and on export data page.


signedav:
@m-kuhn Last fix introduced a new bug - solved it differently now to be stable for the wizard page (on export) and on the panel (validator)...

m-kuhn (approved):

Issue #471, opened 2021-02-09
Sugg: sort "select multiple interlis models"


Pierre-de-la-Verre:
The list of the found models seem to be sorted by its path, not its name.
I think it could be nice when the list is (optional) sorted by name of the models.
![grafik](https://user-images.githubusercontent.com/13277944/107394922-a5268500-6afc-11eb-9cc6-e0850e5bbcf4.png)

signedav:
Should they be sorted by the name only? 
Or by the name and the repo (grouped by repo)?
And should the repo be sorted as well (according to their name or according to the steps going through ilisite.xml)?

Pierre-de-la-Verre:
I think that by name would be the minimum. 
A max solution would be a multiple-sortable table like Excel, but that's an overload ..

PR #804, opened 2023-05-05
Sort models / toppings in autocompleter box 

signedav:
Before it has been sorted like it has been added to the model. Means mostly it had kind of unstable sorting according the repository. This repository sorting is gone, but I think it's fine, mostly you look after a model name you know.

[Screencast from 05.05.2023 08:36:48.webm](https://user-images.githubusercontent.com/28384354/236391540-b8a0cff4-916f-46fe-8f7d-baa2e9587f98.webm)

First model on a punch is now AGNES, what is kind of nice, since it's a female name and the title of quite a nice novel by Peter Stamm.

This requires https://github.com/opengisch/QgisModelBakerLibrary/pull/58 and resolves https://github.com/opengisch/QgisModelBaker/issues/471

### DIFF
--- a/QgisModelBaker/gui/workflow_wizard/workflow_wizard.py
+++ b/QgisModelBaker/gui/workflow_wizard/workflow_wizard.py
@@ -505,12 +505,15 @@ class WorkflowWizard(QWizard):
             for downloaded_file_id in topping_file_cache.downloaded_files:
                 if downloaded_file_id in missing_file_ids:
                     missing_file_ids.remove(downloaded_file_id)
-            self.log_panel.print_info(
-                self.tr(
-                    "- - Some topping files where not successfully downloaded: {}"
-                ).format(" ".join(missing_file_ids)),
-                LogColor.COLOR_TOPPING,
-            )
+            try:
+                self.log_panel.print_info(
+                    self.tr(
+                        "- - Some topping files where not successfully downloaded: {}"
+                    ).format(" ".join(missing_file_ids)),
+                    LogColor.COLOR_TOPPING,
+                )
+            except Exception:
+                pass
 
         return topping_file_cache.model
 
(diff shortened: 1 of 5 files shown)


QgisModelBaker/gui/workflow_wizard/workflow_wizard.py
```diff
@@ -505,12 +505,15 @@ def get_topping_file_model(self, id_list):
...
-            )
+            try:
+                self.log_panel.print_info(
+                    self.tr(
+                        "- - Some topping files where not successfully downloaded: {}"
+                    ).format(" ".join(missing_file_ids)),
+                    LogColor.COLOR_TOPPING,
+                )
+            except Exception:
+                pass
```
domi4484, line 516: instead of `pass` could be a QgsMessageLog with the same text that would have been shown in the log panel?
Or what error could have been raised here?

domi4484 (approved):
Looks good to me!

QgisModelBaker/gui/workflow_wizard/workflow_wizard.py
signedav: Uh oh. You found my dirty hack.

When downloading a file we do that in another thread. If the user closes the dialog this thread is still running and on a failure of the download the dialog does not exist anymore. This is similar to the situation with the running java thread (maybe you remember). This would be a bigger effort to fix. Here we just pass. Because when the user closes the dialog, she/he is not interested in the feedback. But yes, a QgsMessageLog would be appropriate.

signedav: Hmm.. But this does not work. It's the whole object that is deleted on closing the dialog. So no option to write the LogMessage

signedav: I leave it like this. There is an issue open to solve all this threading problem. Thing is, that the user encounters it very rarely. I encounter it more often while testing.

PR #808, opened 2023-06-07
Enable automatic transaction groups per default

signedav:


signedav:
Nah, I better don't do it. Let's do the buffered when it works properly.

PR #810, opened 2023-06-21
Deactivate next button while searching for referenceData 

### LINKED ISSUE #809: Incomplete or long loading data resources
### What it is about
When a model has been imported, we check for potetial reference data - means we search the repositories for ilidata.xml entries that reference this model. Because according to this result we go after schema import to the data import page or - in case no reference data found - to the project generation page.

In `IliDataCache`:
```
self.model.modelReset.connect(lambda: self.model_refreshed.emit(self.model.rowCount())
```

### Bug
It waits until the model has been reset (by signal `modelReset`) but this behavior is unreliable, since it calls the `modelReset` after the ilidata.xml parsing of the first repo - means, we go on even when not the complete. ***Pretty sure, no-one noticed, since it's quicker in parsing all the repos than switching the page (not even a second), but it could has been possible, that no referenceData is found and then no data import page would have been appeared.***

Another thing is, that it waits until timeout when no ilidata is available, so when we only work with a local repo (directory) it never resets the model. That's pretty long (10 seconds). The user thinks it's stuck and might be tempted to press "next" again and again and again...

Screenshot from 2023-06-21 13-04-55

### Sollution approaches
- [ ] Concerning unreliability we need to know if all repos are parsed and the model updated, to go on. Means having kind of a counter "processing repos" and  then count up on every repo found on ilisites and count down when the request timeouted etc. But maybe this request is obsolete after potential redesign has been made. See https://gist.github.com/signedav/7c9e2c8e895003bd1f7a9b4dd80797c0 
- [ ] Because it's bad designed anyway to block the wizard on such requests (would be preferable to have somewhere a button "Get reference Data" - maybe that could be a workaround solution) but still it would need the first approach
- [x] Decrease the timeout and lock "Next" Button during this process improve the bad solution at least (quick fix)

signedav:
As well decrease the timeout, since it's unreliable anyway and usually gets a result in less than one second. 
This does not fix #809 in a nice way but makes it ways more enjoyable

### DIFF
--- a/QgisModelBaker/gui/topping_wizard/referencedata_page.py
+++ b/QgisModelBaker/gui/topping_wizard/referencedata_page.py
@@ -145,7 +145,7 @@ class ReferencedataPage(QWizardPage, PAGE_UI):
         timer = QTimer()
         timer.setSingleShot(True)
         timer.timeout.connect(lambda: loop.quit())
-        timer.start(10000)
+        timer.start(3000)
         self.refresh_referencedata_cache(filter_models, type)
         loop.exec()
         return self.ilireferencedatacache.model
--- a/QgisModelBaker/gui/workflow_wizard/workflow_wizard.py
+++ b/QgisModelBaker/gui/workflow_wizard/workflow_wizard.py
@@ -257,6 +257,7 @@ class WorkflowWizard(QWizard):
                         "Checking for potential referenced data on the repositories (might take a while)..."
                     )
                 )
+                self.schema_configuration_page.setComplete(False)
                 if (
                     self.import_data_file_model.rowCount()
                     or self.update_referecedata_cache_model(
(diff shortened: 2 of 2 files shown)


QgisModelBaker/gui/topping_wizard/referencedata_page.py
```diff
@@ -145,7 +145,7 @@ def update_referecedata_cache_model(self, filter_models, type):
         timer = QTimer()
         timer.setSingleShot(True)
         timer.timeout.connect(lambda: loop.quit())
-        timer.start(10000)
+        timer.start(3000)
```
signedav, line 148: Here we have 3 seconds to be supersure and there the next-page does not rely on this timeout (compared to the one in the workflow wizard)...

domi4484 (approved):

PR #811, opened 2023-07-11
Clear ilicache button

### LINKED ISSUE #776: Button to clear ilicache
Lots of users struggle with the .ilicache. Either they are not aware or they don't know that there is a .ilicache which could cause some problems.

It would be fancy to have a "Clear ILI-Cache" Button in Model Baker, so users (a) know that there is a cache and (b) can empty the cache very easily without having to know how ili2db handles the cache.

E.g. like this:

image

domi4484:
Fix #776

Depends on https://github.com/opengisch/QgisModelBakerLibrary/pull/60

image

### DIFF
--- a/QgisModelBaker/gui/workflow_wizard/import_source_selection_page.py
+++ b/QgisModelBaker/gui/workflow_wizard/import_source_selection_page.py
@@ -20,7 +20,7 @@
 
 from qgis.core import QgsApplication
 from qgis.PyQt.QtCore import Qt
-from qgis.PyQt.QtWidgets import QCompleter, QWizardPage
+from qgis.PyQt.QtWidgets import QCompleter, QPushButton, QWizardPage
 
 import QgisModelBaker.utils.gui_utils as gui_utils
 from QgisModelBaker.libs.modelbaker.iliwrapper.ilicache import (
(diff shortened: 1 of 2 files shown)


QgisModelBaker/gui/workflow_wizard/import_source_selection_page.py
```diff
@@ -186,3 +189,6 @@ def _remove_selected_rows(self):
         indices = self.source_list_view.selectionModel().selectedIndexes()
         self.source_list_view.model().remove_sources(indices)
         self.remove_button.setEnabled(self._valid_selection())
+
+    def _clear_ilicache_button_clicked(self):
+        print("_clear_ilicache_button_clicked")
```
signedav, line 194: And what happens then?

domi4484: ouch I have lost a commit somewhere

signedav (approved):
The code looks good. Have you tested it out with the whole workflow?

domi4484:
> The code looks good. Have you tested it out with the whole workflow?

I have checked that the cache directories get deleted and then created again  when running some ili operations again.

Is there something else I should pay attention?

PR #813, opened 2023-07-12
Add a visible error dialog in case Java is not installed. Fix #785

### LINKED ISSUE #785: Nicer outputs on Java / PG errors
On having Java not installed the user gets a JavaNotFoundError with a Python stack.
The Scram issue reported here #779 is a line at the end of the log.

Those errors could be handled nicer. With popup "Do you have Java installed? If not follow this instruction" etc..

domi4484:
image

### DIFF
--- a/QgisModelBaker/gui/export.py
+++ b/QgisModelBaker/gui/export.py
@@ -470,6 +470,12 @@ class ExportDialog(QDialog, DIALOG_UI):
                 self.txtStdout.setText(e.error_string)
                 self.enable()
                 self.progress_bar.hide()
+
+                QApplication.restoreOverrideCursor()
+                QMessageBox.critical(
+                    self, self.tr("Java not found error"), e.error_string
+                )
+
                 return
 
             self.buttonBox.clear()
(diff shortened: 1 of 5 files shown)


signedav (commented):
Nice. Where do the links lead to?

signedav:
The second link should lead to https://opengisch.github.io/QgisModelBaker/user_guide/plugin_configuration/#interlis-settings

The first one not to Oracle Java (preferable OpenJDK, so https://adoptium.net/), but since it can be easier installed on linux, maybe we can add another part in the docu for it as well (so it's easy to append more information if we need them). Maybe you can add another page here: https://github.com/opengisch/QgisModelBaker/tree/master/docs/background_info (add it then here as well https://github.com/opengisch/QgisModelBaker/blob/master/mkdocs.yml#L53

There can be then the adoptium link and for linux info like:
```
sudo apt-get install openjdk-18-jre 
```

domi4484:
I have fixed the links in this pr: https://github.com/opengisch/QgisModelBakerLibrary/pull/62

I think the problem if we start differentiating the instructions per Os is that for every flavour will be different and we will have to mantain it. The version from adoptium can also be used very easily on linux by setting JAVA_HOME

signedav:
Fair enough.

signedav (approved):

PR #814, opened 2023-07-12
Add persistent logging functionality

domi4484:


domi4484:
@signedav during implementation I have noticed that there is a LogPanel widget. But in some (legacy?) places like for example `QgisModelBaker/gui/export.py` it is not used handing to some duplicate code. Would it make sense to refactor those files or are they meant to disappear?

signedav:
Dont concern these export.py and import.py They should be removed in future...

QgisModelBaker/gui/export.py
```diff
@@ -17,6 +17,7 @@
  ***************************************************************************/
 """
 
+import logging
```
domi4484, line 20:
```suggestion
```

QgisModelBaker/gui/export.py
```diff
@@ -480,11 +481,13 @@ def accepted(self, edited_command=None):
     def print_info(self, text):
         self.txtStdout.setTextColor(QColor(LogColor.COLOR_INFO))
         self.txtStdout.append(text)
+        logging.info(text)
```
domi4484, line 484:
```suggestion
```

QgisModelBaker/gui/export.py
```diff
@@ -480,11 +481,13 @@ def accepted(self, edited_command=None):
...
         QCoreApplication.processEvents()
 
     def on_stderr(self, text):
         color_log_text(text, self.txtStdout)
         self.advance_progress_bar_by_text(text)
+        logging.error(text)
```
domi4484, line 490:
```suggestion
```

QgisModelBaker/gui/generate_project.py
```diff
@@ -17,6 +17,7 @@
  ***************************************************************************/
 """
 import configparser
+import logging
```
domi4484, line 20:
```suggestion
```

QgisModelBaker/gui/generate_project.py
```diff
@@ -693,11 +694,13 @@ def accepted(self, edited_command=None):
     def print_info(self, text, text_color=LogColor.COLOR_INFO):
         self.txtStdout.setTextColor(QColor(text_color))
         self.txtStdout.append(text)
+        logging.info(text)
```
domi4484, line 697:
```suggestion
```

QgisModelBaker/gui/generate_project.py
```diff
@@ -693,11 +694,13 @@ def accepted(self, edited_command=None):
...
         QCoreApplication.processEvents()
 
     def on_stderr(self, text):
         color_log_text(text, self.txtStdout)
         self.advance_progress_bar_by_text(text)
+        logging.error(text)
```
domi4484, line 703:
```suggestion
```

signedav:
Will the implementation of the logpanel in the validator be in a follow up @domi4484 ?

QgisModelBaker/gui/panel/log_panel.py
```diff
@@ -52,6 +54,17 @@ def print_info(self, text, text_color=LogColor.COLOR_INFO):
         self.txtStdout.setTextColor(QColor(text_color))
         self.txtStdout.append(text)
 
+        if text_color == LogColor.COLOR_INFO:
```
signedav, line 57: I'm a little bit unhappy about this because it looks a little strange to use the color as indicator of the type of message. It would be cooler - maybe - to have a type info / warning / fail / success... On the other hand, it's nice to be able to override the the color in case we have to visualize something specific (like we do it on topping).

So after all I would keep it like it is (no change requested). But maybe you have an idea to make it nicer...

signedav (approved):
Code looks good. I haven't  ested it though.

QgisModelBaker/gui/panel/log_panel.py
domi4484: It was also my thought but did not came to a better idea in the moment

signedav: Let's keep it dynamic :balloon:

domi4484:
> Will the implementation of the logpanel in the validator be in a follow up @domi4484 ?

Thanks I forgot that part. Will add it.

QgisModelBaker/gui/panel/log_panel.py
domi4484: I took another look at it and I think it's a bit a "puff" with those colors, there is also another function in ili2dbutils called `color_log_text`. But maybe this is not used anymore?


> Dont concern these export.py and import_data.py and generate_project.py They should be removed in future...

Is that not possible at the moment? Or it just need to be done by someone?

signedav: Needs to be done by someone. You can check it out if you have time.

domi4484:
@signedav now stdout/err of the validation are added to the logs.

Still I didn't implemented the logs tab that we discussed in the panel for now.

PR #815, opened 2023-07-12
Drag&Drop support for extra meta attribute files (toml/ini)

### LINKED ISSUE #534: Consider TOML in Wizard Source Selector
Maybe it would be good to give a lot of use cases the possibility to just add all files there without any other selections needed.
SQL is not that easy, since the user needs to define if it's a pre- or post-script.

domi4484:
Fix #534

### DIFF
--- a/QgisModelBaker/gui/workflow_wizard/workflow_wizard.py
+++ b/QgisModelBaker/gui/workflow_wizard/workflow_wizard.py
@@ -16,6 +16,7 @@
  *                                                                         *
  ***************************************************************************/
 """
+import logging
 import os
 import pathlib
 import re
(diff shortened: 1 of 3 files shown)


QgisModelBaker/gui/workflow_wizard/workflow_wizard.py
```diff
@@ -564,13 +564,22 @@ def add_source(self, source, origin_info):
...
+    def append_dropped_files(self, dropped_files, dropped_ini_files):
         if dropped_files:
             for dropped_file in dropped_files:
                 self.add_source(
                     dropped_file, self.tr("Added by user with drag'n'drop.")
                 )
 
+        if dropped_ini_files:
+            if len(dropped_ini_files) > 1:
+                # TODO logging.warning("Only one INI/TOML file is supported by drag&drop: {}".format(dropped_ini_files))
```
signedav, line 576: Can you do it or just ignore it. But I think it's an easy one...

signedav (changes requested):

QgisModelBaker/gui/workflow_wizard/workflow_wizard.py
domi4484: As soon as #814 is merged it can be uncommented

signedav (approved):

PR #824, opened 2023-09-07
Support drop of INI Files as well in Source Selector

signedav:
and fix break of signal


signedav:
Good to merge @domi4484 ?

domi4484 (approved):
Looks good, could you just replace the literal list with the const here: https://github.com/opengisch/QgisModelBaker/blob/master/QgisModelBaker/qgismodelbaker.py#L542 ?

PR #828, opened 2023-09-25
store validator config file in project variable -

### LINKED ISSUE #820: Store path "Validator config file" in Data Validation in QGIS project
We outsource our checks for a model to an additional model, which should only be used while checking. Therefore we'd love to save the path to "Validator config file" in the QGIS project.

image

So we cann assure, that the user is always testing with the additional model and don't have to (a) know where the ini-file is and (b) add the ini-file each time he wants to check the data.

signedav:
- [ ] Resolve path to relative
- [ ] on layer change to another schema and back - what should happen? reset (to none or the project var) or keep it?

- Resolves #820


QgisModelBaker/gui/validate.py
```diff
@@ -363,9 +380,10 @@ def _run(self, edited_command=None):
         validator.configuration.skip_geometry_errors = (
             self.skip_geometry_errors_check_box.isChecked()
         )
-        validator.configuration.valid_config = (
-            self.validator_config_file_line_edit.text()
+        validator.configuration.valid_config = self._absolute_path(
+            self.config_file_line_edit.text()
         )
+        print(validator.configuration.valid_config)
```
domi4484, line 386: forgotten print? or is it just to check if i look at the changes? 😄

QgisModelBaker/gui/validate.py
```diff
@@ -632,12 +650,50 @@ def _auto_zoom_button_clicked(self):
...
-    def _select_validator_config_file(self):
+    def _select_config_file(self):
         filename, _ = QFileDialog.getOpenFileName(
             self, self.tr("Select the validator config file")
         )
         if filename:
-            self.validator_config_file_line_edit.setText(filename)
+            self.config_file_line_edit.setText(self._relative_path(filename))
+
+    def _save_config_file(self):
```
domi4484, line 660: I would name it `_save_config_file_path` to make it clear that the path to the file is stored, and not the entire file itself.

Maybe in the gui as well.

domi4484 (approved):
Why do we need the load "Load from project" button? Isn't it always loaded when the widget becomes visible? Or is it just in case someone edits it and want to get back the stored value?

signedav:
> Why do we need the load "Load from project" button? Isn't it always loaded when the widget becomes visible? Or is it just in case someone edits it and want to get back the stored value?
Yes, that was my thought. Otherwise you would need to go to the Project Variables and copy it around.

QgisModelBaker/gui/validate.py
signedav: yes :-p

PR #830, opened 2023-10-07
Fix absolute path conversion

signedav:
It didn't converted the path to absolute, because the file was not recognized


signedav:
Merging without review because of it's minority and size.

PR #840, opened 2023-11-09
Schemas combobox

### LINKED ISSUE #456: Combobox for schemas when exporting
When exporting a project, it could suggest some schemas which are available on the system (combobox, autocompletion, ...).
Either by retrieving them from the database or by looking them up in the layers of the current project.

It should still be possible to write custom values.

domi4484:
Postgres available schema list is read directly from the database and presented in a combo box. Still it is possible to manually entry a new schema name.

Fix #456 

Depends on https://github.com/opengisch/QgisModelBakerLibrary/pull/70

image

### DIFF
--- a/QgisModelBaker/gui/panel/pg_config_panel.py
+++ b/QgisModelBaker/gui/panel/pg_config_panel.py
@@ -19,6 +19,9 @@
 import logging
 from enum import IntEnum
 
+import psycopg2
+import psycopg2.extras
+from psycopg2 import OperationalError
 from qgis.PyQt.QtCore import Qt, pyqtSignal
 
 import QgisModelBaker.libs.modelbaker.libs.pgserviceparser as pgserviceparser
@@ -81,7 +84,9 @@ class PgConfigPanel(DbConfigPanel, WIDGET_UI):
 
         self.pg_host_line_edit.setValidator(nonEmptyValidator)
         self.pg_database_line_edit.setValidator(nonEmptyValidator)
-        self.pg_schema_line_edit.setValidator(nonEmptyValidator)
+        self.pg_schema_combo_box.setValidator(nonEmptyValidator)
+
+        self.pg_schema_combo_box.setEditable(True)
 
         self.pg_host_line_edit.textChanged.connect(self.validators.validate_line_edits)
         self.pg_host_line_edit.textChanged.emit(self.pg_host_line_edit.text())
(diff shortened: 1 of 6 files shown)


QgisModelBaker/gui/panel/pg_config_panel.py
```diff
@@ -440,3 +447,46 @@ def _keep_custom_settings(self):
...
+            return
+
+        sql = """SELECT schema_name
+        FROM information_schema.schemata; """
+
+        cursor = connection.cursor()
+        cursor.execute(sql)
+
+        schemas = cursor.fetchall()
+        print(schemas)
```
domi4484, line 473:
```suggestion
```

QgisModelBaker/gui/panel/pg_config_panel.py
```diff
@@ -440,3 +447,45 @@ def _keep_custom_settings(self):
...
+            logging.warning(f"Pg connection error: {exception}")
+            return
+
+        sql = """SELECT schema_name
+        FROM information_schema.schemata; """
+
+        cursor = connection.cursor()
+        cursor.execute(sql)
+
+        schemas = cursor.fetchall()
```
signedav, line 472: Hm... Maybe this would be nicer to go over the db_connector like most requests we perform on the database. Or are there any limitations?

```
db_connector = db_utils.get_db_connector(configuration)
        if db_connector and db_connector.get_schema:     
```

QgisModelBaker/gui/panel/pg_config_panel.py
```diff
@@ -440,3 +447,45 @@ def _keep_custom_settings(self):
             self.pg_ssl_mode_combo_box.currentData(),
             PgConfigPanel._SERVICE_COMBOBOX_ROLE.SSLMODE,
         )
+
+    def _fill_schema_combo_box(self):
```
signedav, line 451: When is this called? Is this called when I change the e.g. db name?

QgisModelBaker/gui/panel/pg_config_panel.py
domi4484: I thought about that but then I have seen that the pg_connector need a schema as parameter for the constructor.
But by re-looking at that I can just leave the schema empty... Will change this

QgisModelBaker/gui/panel/pg_config_panel.py
domi4484: Yes, this is the idea (but I have completely forgotten it in the end 🙈)
I will add a timer so that it isn't triggered continuosly during editation. Like it is done here https://github.com/opengisch/QgisModelBaker/blob/203428af241f25b956fe5418e1c95269d832087a/QgisModelBaker/gui/dataset_manager.py#L78

QgisModelBaker/gui/panel/pg_config_panel.py
```diff
@@ -395,6 +410,9 @@ def _pg_service_combo_box_changed(self):
             == self.pg_ssl_mode_combo_box.findData(None)
         )
 
+        logging.info("_pg_service_combo_box_changed")
```
domi4484, line 413:
```suggestion
```

domi4484:
I don't get why are the tests failing for 3.16... do you have an hint @signedav ?

signedav:
No idea. Maybe it's the 3.16 build. I would need to have a deeper look to find out...

domi4484:
> No idea. Maybe it's the 3.16 build. I would need to have a deeper look to find out...

Ok thanks I'll try to find out...

signedav:
Maybe related to this https://github.com/qgis/QGIS/pull/40567 But I wonder why it runs into it on this PR an not on the others...

domi4484:
@signedav in the last commit I replaced qgis 3.16 with 3.28 like in the library. Now all tests are passing.
I did one more try and ran the plugin on QGIS 3.16, from a quick test everything seems to work normally....

signedav:
Cool.

PR #842, opened 2023-11-12
Project creation option to expose t_ili_tid

### LINKED ISSUE #713: Expose t_ili_id for manual input
Sometimes it's necessery to give specific OID to objects.  Those have to be given manual (so https://github.com/opengisch/QgisModelBaker/issues/599 won't help).

It would be nice, if `t_ili_tid` would be exposed in the formular. To be discussed if it may be an option at generating project, if a user wants to expose `t_ili_tid` or not (default not exposed). Maybe there are better ideas around.

domi4484:
Fix #713

Depends on https://github.com/opengisch/QgisModelBakerLibrary/pull/72

image

### DIFF
--- a/QgisModelBaker/gui/workflow_wizard/project_creation_page.py
+++ b/QgisModelBaker/gui/workflow_wizard/project_creation_page.py
@@ -85,6 +85,7 @@ class ProjectCreationPage(QWizardPage, PAGE_UI):
                 mgmt_uri=mgmt_uri,
                 consider_basket_handling=True,
                 optimize_strategy=self.optimize_combo.currentData(),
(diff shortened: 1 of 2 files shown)


domi4484:
@signedav did you had a better location in mind where to put this option?

signedav:
Hm... I dind't made toughts about the setting. After thinking TID Generator this would make more sense to put it there I guess. See https://github.com/opengisch/QgisModelBakerLibrary/pull/69 we could discuss it again. Are you in Bern at Wednesday?

domi4484:
Yes I will be in Bern.. we can look together

PR #844, opened 2023-11-17
TID (OID) Configurator

signedav:
#### There are some panels:
- The `LayerTIDsPanel` providing the view of all the single layers and the OID settings (and function to save)
- The `SetSequencePanel` providing the current sequence (and function to save)
- First two are summarized in `TIDConfiguratorPanel` having a "Reset Values" Button
- Wizard Page `TIDConfigurationPage` containing  `TIDConfiguratorPanel` and button to set (save).
- TID (OID) manager openable by menu and saving settings with "OK".

#### Wizard Page
Screenshot from 2023-11-20 13-38-58

#### TID Manager
Open with *Database > Model Baker > TID (OID) Manager*

Screenshot from 2023-11-20 13-40-54

### Depends on
https://github.com/opengisch/QgisModelBakerLibrary/pull/69 https://github.com/opengisch/QgisModelBakerLibrary/pull/68
https://github.com/opengisch/QgisModelBakerLibrary/pull/71


docs/background_info/oid_tid_generator.md
```diff
@@ -0,0 +1,145 @@
...
+... while the `t_id` column contains just a ***schema-internal*** sequence used for foreign keys etc. Those `t_id`s are ***not*** the cross-system unique identificator.
+
+![oid physical data](../assets/oid_physical_data.png)
+
+!!! Note
+    There are situations when that you can find the `t_id` as `TID` in your data. It's when there is no `t_ili_tid` available ili2db could use. But be aware those `TID`s are not stable then.
+
+### OID Domains
+
+There are different types of OID domains. In Model Baker we try to set default values that fits into the definition.
```
m-kuhn, line 38:
```suggestion
There are different types of OID domains. In Model Baker we try to set default values that fit into the definition.
```

docs/background_info/oid_tid_generator.md
```diff
@@ -0,0 +1,145 @@
...
+
+```
+t_id
+```
+
+#### `STANDARDOID`
+
+It's defined as `OID TEXT*16` and follows some specific requirements.
+
+It requireds an 8 char prefix and 8 char postfix:
```
m-kuhn, line 71:
```suggestion
It requires an 8 char prefix and 8 char postfix:
```

docs/background_info/oid_tid_generator.md
```diff
@@ -0,0 +1,145 @@
...
+
+- **Prefix (2 + 6 chars):** Country identifier + a *global identification part* assigned once by the [official authority](https://www.interlis.ch/dienste/oid-bestellen)
```
m-kuhn, line 73:
```suggestion
- **Prefix (2 + 6 chars):** Country identifier + a *global identification part*. The global identification part can be ordered from the [official authority](https://www.interlis.ch/dienste/oid-bestellen)
```

docs/background_info/oid_tid_generator.md
```diff
@@ -0,0 +1,145 @@
...
+Model Baker does not know what your *global identification part* is, why it's important that you edit this part of the expression.
+
+As *local identification part* Model Baker suggests the couter provided by the `t_id` sequence.
+
+In the QGIS Project the default value expression for `t_ili_tid` is therefore this:
+
+```
+'ch100000' || lpad( T_Id, 8, 0 )
+```
+
```
m-kuhn, line 86:
```suggestion

assuming `ch` as the country code and `100000` as the registered OID prefix.
```

docs/background_info/oid_tid_generator.md
```diff
@@ -0,0 +1,145 @@
...
+#### `ANYOID`
+
+The `ANYOID` does not define a format of the OID but just that an OID needs to be defined in all the extended models. This domain is only used on topics that needs to be extended. There this sollution has some [limitations](#limitations).
```
m-kuhn, line 89:
```suggestion
The `ANYOID` does not define a format of the OID but just that an OID needs to be defined in all the extended models. This domain is only used on topics that need to be extended. There this sollution has some [limitations](#limitations).
```

docs/background_info/oid_tid_generator.md
```diff
@@ -0,0 +1,145 @@
...
+
+Find the **TID (OID) Manager** via the *Database > Model Baker* menu.
+
+![tid manager](../assets/oid_tid_manager.png)
+
+Here you can use the QGIS Expression Dialog to edit the default value expression for the `t_ili_tid` field of each layer.
+
+If you want to have the `t_ili_tid` exposed to the form, you can select ***Show***.
+
+If you need a counter in the expressions, you can use the `t_id` field, that has a schema-wide sequence counting up. This sequence can be reset as well by the user, but be carefull not to set it lower than already existing `t_id`s in your project. See below the [limitations](#limitations).
```
m-kuhn, line 127:
```suggestion
If you need a counter in the expressions, you can use the `t_id` field, that has a schema-wide sequence counting up. This sequence can be reset as well by the user, but be careful not to set it lower than already existing `t_id`s in your project. See below the [limitations](#limitations).
```

docs/background_info/oid_tid_generator.md
```diff
@@ -0,0 +1,145 @@
...
+
+### Limitations
+
+This sollution covers a lot, but not everything.
```
m-kuhn, line 131:
```suggestion
This solution covers a lot, but not everything.
```

docs/user_guide/import_workflow.md
```diff
@@ -138,7 +138,19 @@ Then, with a simple click you generate your project. All the toppings received f
 
 ![baked project](../assets/generated_project.png)
 
-Bon appetit!
+### TID (OID) Values
+
+Often the models definition requires cross-system unique identificators. So called OIDs or TIDs, what are represented in the physical database as the `t_ili_tid` column. Find a clear definition and more details about them in the [corresponding chapter](../../background_info/oid_tid_generator).
+
+On creating a QGIS Project with Model Baker, there are default default value expression generated for the `t_ili_tid` field on the attribute form. But often those default value expressions need to be edited by the user (like e.g. the prefix in the `STANDARDOID`).
```
m-kuhn, line 145: I guess the double default in "default default value expression" is on purpose, would it make sense to replace the first one with "preset" or reformulate to "a default value expression is added by Model Baker"?

docs/user_guide/import_workflow.md
```diff
@@ -138,7 +138,19 @@ Then, with a simple click you generate your project. All the toppings received f
...
+
+![tid generator](../assets/workflow_wizard_tid_generator_page.png)
+
+Here you can use the QGIS Expression Dialog to edit the default value expression for the `t_ili_tid` field of each layer.
+
+If you need a counter in the expressions, you can use the `t_id` field, that has a schema-wide sequence counting up. This sequence can be reset as well by the user, but be carefull not to set it lower than already existing `t_id`s in your project.
```
m-kuhn, line 151:
```suggestion
If you need a counter in the expressions, you can use the `t_id` field, that has a schema-wide sequence counting up. This sequence can be reset as well by the user, but be careful not to set it lower than already existing `t_id`s in your project.
```

QgisModelBaker/qgismodelbaker.py
```diff
@@ -131,6 +134,16 @@ def initGui(self):
             self.tr("Dataset Manager"),
             None,
         )
+        self.__tidmanager_action = QAction(
+            QIcon(
+                os.path.join(
+                    os.path.dirname(__file__),
+                    "images/QgisModelBaker-tidmanager-icon.svg",
```
m-kuhn, line 141: is this file missing or am I missing something??

QgisModelBaker/gui/workflow_wizard/project_creation_page.py
```diff
@@ -63,6 +62,16 @@ def __init__(self, parent, title):
 
         self.create_project_button.clicked.connect(self._create_project)
 
+        self.is_complete = False
+
+    def isComplete(self):
```
m-kuhn, line 67: Random suggestion (because I had to lookup the `QWizardPage` documentation to find that this overrides a virtual method and this would have helped me see this directly from here): https://stackoverflow.com/a/8313042/2319028

m-kuhn (approved):
Great improvement, thanks a lot

QgisModelBaker/qgismodelbaker.py
signedav: Oh yes. My artistic streak was not active yet :-)

docs/background_info/oid_tid_generator.md
m-kuhn: or is `ch100000` the OID prefix and `100000` the "global identification part"?

docs/background_info/oid_tid_generator.md
```diff
@@ -0,0 +1,145 @@
...
+
+It requireds an 8 char prefix and 8 char postfix:
+
+- **Prefix (2 + 6 chars):** Country identifier + a *global identification part* assigned once by the [official authority](https://www.interlis.ch/dienste/oid-bestellen)
+
+- **Postfix (8 chars):** Sequence (numeric or alphanumeric) of your system as *local identification part*
+
+Model Baker does not know what your *global identification part* is, why it's important that you edit this part of the expression.
+
+As *local identification part* Model Baker suggests the couter provided by the `t_id` sequence.
```
signedav, line 79:
```suggestion
As *local identification part* Model Baker suggests the counter provided by the `t_id` sequence.
```

docs/user_guide/import_workflow.md
signedav, line 145:
```suggestion
On creating a QGIS Project with Model Baker, there are preset default value expression generated for the `t_ili_tid` field on the attribute form. But often those default value expressions need to be edited by the user (like e.g. the prefix in the `STANDARDOID`).
```

signedav: yes on pupose and yes confusing. changed it...

docs/background_info/oid_tid_generator.md
```diff
@@ -0,0 +1,145 @@
...
+
+It's defined as `OID TEXT*16` and follows some specific requirements.
+
+It requireds an 8 char prefix and 8 char postfix:
+
+- **Prefix (2 + 6 chars):** Country identifier + a *global identification part* assigned once by the [official authority](https://www.interlis.ch/dienste/oid-bestellen)
+
+- **Postfix (8 chars):** Sequence (numeric or alphanumeric) of your system as *local identification part*
+
+Model Baker does not know what your *global identification part* is, why it's important that you edit this part of the expression.
```
signedav, line 77:
```suggestion
Model Baker does not know what your *global identification part* is and uses a test-prefix `ch100000`. It's important that replace this part with your own prefix. 
```

signedav: @m-kuhn hope it's clearer now.

QgisModelBaker/gui/workflow_wizard/project_creation_page.py
signedav: I like it. 

Is this something we do as best-practice?

docs/background_info/oid_tid_generator.md
m-kuhn: @signedav  test-prefix -> sample-prefix?

beistehen: How about "registering" a real prefix with interlis.ch as a MB sample prefix? I might have the possibility to tweak the randomness of the prefix value... 😉

m-kuhn: can we have `changeme`? :-)

m-kuhn:
> I like it.
> 
> Is this something we do as best-practice?

@signedav this comment is somehow shown out of context, I assume it's about `@override`.
I have only found this right now, we can discuss if it should be added to best practice, apparently starting from python 3.12 (which is super new, so we cannot rely on it so far ...) there's even native support for such a construct.

docs/background_info/oid_tid_generator.md
signedav: fine

signedav: Oh I misunderstood.

`changeme` would be cool because it's descriptive, but only if we really can register it - otherwise it could lead to conflicts with the real owners of this prefix.

To register a Model Baker prefix is very undescriptive. It does not imply that you need to change it. And suddenly we have huge conflicts because people use the Model Baker prefix in their data.

Same with `ch100000` but at least this is defined as the test-prefix...

signedav: See 
image

beistehen: 😆 
`chANGEME` would at least emphasize the swissness of the default value...

seriously (and somehow out of scope of this PR, but anyway): a default/sample prefix will rarely be exchanged by the end user, because it will generate a completely valid transfer file. But if validating an exported transfer file raises OID errors because of a missing prefix, the user needs to address it (by inserting a prefix or even order one first at interlis.ch). So the awareness to get prefixes is raised when not having a valid fallback.

This contradicts my first comment, I know.

docs/background_info/oid_tid_generator.md
```diff
@@ -0,0 +1,145 @@
...
+
+It's defined as `OID TEXT*16` and follows some specific requirements.
+
+It requires an 8 char prefix and 8 char postfix:
+
+- **Prefix (2 + 6 chars):** Country identifier + a *global identification part*. The global identification part can be ordered from the [official authority](https://www.interlis.ch/dienste/oid-bestellen)
+
+- **Postfix (8 chars):** Sequence (numeric or alphanumeric) of your system as *local identification part*
+
+Model Baker does not know what your *global identification part* is, why it's important that you edit this part of the expression.
```
signedav, line 77:
```suggestion
Model Baker does not know what your *global identification part* is and uses a sample-prefix `ch100000`. It's important that replace this part with your own prefix.
```

docs/background_info/oid_tid_generator.md
signedav: Yes that is true. But I don't know if the goal should be that a new user creating the database and project and collects data already runs into validation issues after using the "suggested" default values.

m-kuhn: Ideally they should be made aware before adding objects with OIDs.

signedav: Not sure. Ideally yes, but often people just check out INTERLIS with the Model Baker as a first step / a learning and then I think they should not be confronted with STANDARDOID registration as a first step.

Then maybe it's better to create an invalid OID as preset value. Then they encounter it at the validation / export (and are still able to export the data invalid)...

beistehen: Oh, I forgot about the recommendation in the reference manual!

I just got triggered by your sample/test prefix discussion and I am absolutely not into technical details of this PR. But in other commercial systems I know, the prefix is stored in a metadata table. When one changes it, it is updated for all data objects in the project. So the change can also be made after having objects already inserted into the db.

How about using a sample prefix that does not validate when used in an OID? Like `%bmpref%` or `%change%`?

signedav: > But in other commercial systems I know, the prefix is stored in a metadata table. When one changes it, it is updated for all data objects in the project. So the change can also be made after having objects already inserted into the db.

Do you have more information about that?

> How about using a sample prefix that does not validate when used in an OID? Like %bmpref% or %change%?

Yes that's what I suggested. `_` or `%` as first character breaks the validity. `%change%` would be fine for me.

beistehen: > Yes that's what I suggested. `_` or `%` as first character breaks the validity.

Sorry, I didn't see your comment about this because my browser didn't refresh the page. It's a good sign to have the same idea though :wink:

> `%change%` would be fine for me.

For me too! 👌 

> Do you have more information about that?

Well, this system isn't configured using ili2db so there are no constraints to stick to the provided INTERLIS schema.  But basically a table which contains objects uses the following column setup:

`prefix`, `sequence (project-wide)`, `OID (=concat(prefix+lpad(sequence)))`, `attribute1`, `attribute2` and so on.

The prefix source is located in a metadata table. Using SQL update, the prefix can be changed for the records and therefore the OID will be updated as well.

signedav: Ah okay. That's a nice way to have it implemented in the back end. Not scope of this implementation, but good to know.

Concerning default value I'll go for %change% - in QGIS it's easy to replace all the %change% of a layer with a correct prefix.

If it will be often the case we might consider as well a follow up in the TID (OID) manager: "Replace prefix of STANDARDOID in all table with ..."

PR #845, opened 2023-11-21
Don't handle xml files 2 times

### LINKED ISSUE #837: Handle Import Files lists multiple times the same file
Maybe introduced around #815 ?

image

domi4484:
They are already handled by FileDropListView.ValidExtenstions

Fix #837

### DIFF
--- a/QgisModelBaker/qgismodelbaker.py
+++ b/QgisModelBaker/qgismodelbaker.py
@@ -530,11 +530,6 @@ class DropFileFilter(QObject):
                 if pathlib.Path(url.toLocalFile()).suffix[1:]
                 in FileDropListView.ValidExtenstions
             ]
-            additional_xml_files = [
-                url.toLocalFile()
-                for url in event.mimeData().urls()
-                if pathlib.Path(url.toLocalFile()).suffix[1:] in ["xml", "XML"]
-            ]
             additional_ini_files = [
                 url.toLocalFile()
                 for url in event.mimeData().urls()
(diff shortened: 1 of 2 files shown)


signedav:
Now I get it, what is happening here. Thanks for this PR.

We only want to open our dialog on the extensions ["xtf", "XTF", "itf", "ITF", "ili"] and not on XML, xml - that's why this list `additional_xml_files` has been introduced. But I broke the behavior by checking `FileDropListView.ValidExtenstions`.

So the problem is not that `additional_xml_files` is existing but that  `FileDropListView.ValidExtenstions` contains also xml, XML

So the solution would be:
- remove xml, XML from `ValidExtensions`
- keep `additional_xml_files` checking for something like `ValidXmlExtensions`
- Here https://github.com/opengisch/QgisModelBaker/blob/master/QgisModelBaker/utils/gui_utils.py#L262 `+ ValidXmlExtensions`
- Here https://github.com/opengisch/QgisModelBaker/blob/master/QgisModelBaker/utils/gui_utils.py#L272 `+ ValidXmlExtensions`

Can you do it or - if you are busy - I can do it, since I know now what it is about.

domi4484:
I could only test drops in the main QGIS windows, as the wizard dialog disappears for me when clicking outside in the file manager (maybe a KDE thisg?).

In the main window behaviour is following:
- Drop accepted for xtf, itf, ili
- Drop discarded for xml, ini, toml

In the wizard (untested):
- Drop accepted for xtf, itf, ili, xml, ini, toml

signedav:
But when you drop  xtf, itf, ili, xml, ini, toml in the main window all are accepted, right?

signedav:
> (maybe a KDE thisg?).

Maybe yes...

domi4484:
> But when you drop xtf, itf, ili, xml, ini, toml in the main window all are accepted, right? (because at least one INTERLIS file)

Exactly, if at least one xtf, itf, ili is in the list, then all are accepted

PR #854, opened 2023-11-30
Update schemas in separate task to avoid freezes

### LINKED ISSUE #848: Problem on updating schema config
1. when I change something in the form (e.g. database) it does not update the schema dropdown
[Screencast from 24.11.2023 11:30:49.webm](https://github.com/opengisch/QgisModelBaker/assets/28384354/fafee81a-71f4-4975-ba19-417a9d949cc0)

2. when I don't have the correct credentials but everything else is correct, the GUI freezes (maybe because of long responsetime of the remote server)
image

domi4484:
Fix #848

### DIFF
--- a/QgisModelBaker/gui/panel/pg_config_panel.py
+++ b/QgisModelBaker/gui/panel/pg_config_panel.py
@@ -19,7 +19,7 @@
 import logging
 from enum import IntEnum
 
-from qgis.PyQt.QtCore import Qt, QTimer
+from qgis.PyQt.QtCore import Qt, QThread, QTimer
 
 import QgisModelBaker.libs.modelbaker.libs.pgserviceparser as pgserviceparser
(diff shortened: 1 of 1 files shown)


domi4484:
The same super long freeze is present in older MB versions when entering db.sob.opengis.ch, wrong port, and pressing the `Next` button. In each case when executed in a thread it should not block... I'll try to find out where it locks

signedav:
> The same super long freeze is present in older MB versions when entering db.sob.opengis.ch, wrong port, and pressing the Next button. In each case when executed in a thread it should not block... I'll try to find out where it locks

Hm... Or when it takes that long maybe a shorter time out would be cool. Or don't we have it in our hands and it's a server thing?

signedav:
Why it checks ili2db version? 🤔

domi4484:
Its not necessary, I copy pasted too much code 😅

signedav (approved):
Cool. That looks good.

PR #855, opened 2023-11-30
UsabILItyHub: Filter Toppings according to the Database Source

### LINKED ISSUE #759: Toppings: Possible incompatibilities (gpkg/postgis)
Toppings which are exported from a project which is based on a PostGIS schema will not work properly when you try to use them to generate a project and import a model into a gpkg. 

Example: The table name for dataset table in PostGIS is "t_ili2db_dataset" in gpkg it is "T_ILI2DB_DATASET". This has consequences on the "yaml" file.

Problem: A user, who did not create a topping and has no interest in browsing and reading config files, does not know if a Topping will work for him. 

Possible solution:The user is informed for which data format the topping was created

signedav:
In the `ilidata` it's the `category` with the key `http://codes.modelbaker.ch/preferredDataSource` describing what db the file is made for. If there is no such category it's always displayed.

### Means on import
Filters UsabILIty Hub Metaconfig Files according to the current database.

### Means on topping export
It detects what db sources are used and suggests you to optimize. But you can choose none - means the topping will be displayed on every db source...
Screenshot from 2023-12-01 09-09-18

Resolves #759

Depends on https://github.com/opengisch/QgisModelBakerLibrary/pull/79

### DIFF
--- a/QgisModelBaker/gui/topping_wizard/models_page.py
+++ b/QgisModelBaker/gui/topping_wizard/models_page.py
@@ -70,18 +70,21 @@ class ModelsPage(QWizardPage, PAGE_UI):
         return super().validatePage()
 
     def _refresh(self):
-        self._load_available_models()
+        self._load_available_models_and_sources()
         self.models_model.check_entries(self.topping_wizard.topping.models)
 
-    def _load_available_models(self):
+    def _load_available_models_and_sources(self):
         """
         Collects all the available sources in the project and makes the models_model to refresh accordingly.
         """
         checked_identificators = []
         db_connectors = []
+        sources = set()
+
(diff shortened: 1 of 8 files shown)


signedav:
I know this topic is a little bit special, but code review would be much appreciated @domi4484

sjib:
Screenshot from 2023-12-01 09-09-18
> 
Suggested change in explanation on wizard window - rephrasing:
"These are the models for which metaconfigurations / toppings are available on the Usability Hub. Select the models wanted!" 

and
"Filtering by type of topping. Choose 'none' to allow all."

domi4484 (approved):

PR #856, opened 2023-12-01
set transaction mode as enum

signedav:
Requires https://github.com/opengisch/QgisModelBakerLibrary/pull/80


domi4484 (approved):
Could we define "AutomaticGroups", "Disable", and so on as const in the library instead of using the literals?

signedav:
Yeah. or what do you think about `Qgis.TransactionMode.AutomaticGroups.name`?

domi4484:
> Yeah. or what do you think about `Qgis.TransactionMode.AutomaticGroups.name`?

yes even better!

PR #858, opened 2023-12-04
Topping on create Project

### LINKED ISSUE #479: Toppings on Project Create
Damit auch aus einer bestehenden DB die Usabilityhub Toppings verwendet werden können.

Es muss irgendwie sichergestellt werden, dass die DB auch korrekt erstellt wurde (korrekte TOMLs etc und benötigte Kataloge importiert wurden.).

Dann muss irgendwie gelesen werden, welche Models dass importiert wurden um die Metaconfigurations zu listen.

Es gibt verschiedene Risiken. Sollen Kataloge noch einmal importiert werden oder nicht?

### LINKED ISSUE #768: Enable Topping import for genrating projects from existing databases
Proposal:
- The import of Toppings should also be possible when generating projects from existing databases.

Pros:
- When a project is corrupt, one could easily restore them without regenerating the database and reimporting the data
- One could set up the project for different users and do not have to share projects which were probably changed by the user after they were generated

Cons:
- Compatibility-issues for users using the proposed option with a generated database which is not compatible with the database they have generated (see. #759)

This is an idea which my organization would benefit from. It must be discussed whether other users would also benefit from it, or if the dangers of incompatibilities/complexity is too big.

signedav:
### Get Topping of Metaconfig stored in the DB
Since ili2db 4.11 the id of the metaconfiguration file used to create the physical database is stored in the meta-tables. This means, on **Project Creation** we consider this id and provide the linked **project topping**.

Screenshot from 2023-12-04 14-41-47

### Get Topping by selection
Still it's possible to **disable it** and choose another **project topping file**.

image

You can choose a **project topping file** (identified by model, db-type) from the **UsabILITy Hub** repositories or as well a YAML file from the **local system**.

### Why, *project topping* file
... and not "metaconfiguration* file? 

Because on this step only project topping files are relevant. Metaconfiguration files are only relevant, when the schema is created. We thought about to provide them, to choose the same like on schema import, but this part is already provided by getting the topping from the DB.

### To do
- [x] preferredDataSource
- [x] Finetune GUI
- [x] Synchronize with other open PRs

Resolves #479 and resolves #768

## Busy Bar

Not directly related to this topic but it popped up again. Sometimes the wizard has to load - mostly checking the repos to suggest data / reload the comboboxes. And there users are confused, what's happening. For this reason it does now:
- Disable the Wizard Page (not cancel button)
- Showing a busy bar... 

[Screencast from 04.12.2023 16:24:24.webm](https://github.com/opengisch/QgisModelBaker/assets/28384354/95a0f31c-8d70-4ba5-b728-4c711d862270)

### DIFF
--- a/QgisModelBaker/gui/workflow_wizard/workflow_wizard.py
+++ b/QgisModelBaker/gui/workflow_wizard/workflow_wizard.py
@@ -694,11 +694,10 @@ class WorkflowWizard(QWizard):
 
     def busy(self, page, busy, text="Busy..."):
         page.setEnabled(not busy)
+        self.log_panel.busy_bar.setVisible(busy)
         if busy:
             self.log_panel.busy_bar.setFormat(text)
-            self.log_panel.busy_bar.setVisible(True)
         else:
-            self.log_panel.busy_bar.setVisible(False)
             self.log_panel.scrollbar.setValue(self.log_panel.scrollbar.maximum())
 
 
--- a/QgisModelBaker/gui/workflow_wizard/workflow_wizard.py
+++ b/QgisModelBaker/gui/workflow_wizard/workflow_wizard.py
@@ -272,8 +272,10 @@ class WorkflowWizard(QWizard):
                         "Checking for potential referenced data on the repositories (might take a while)..."
                     )
                 )
-                self.schema_configuration_page.busy(
-                    True, self.tr("Checking for potential referenced data...")
+                self.busy(
+                    self.schema_configuration_page,
+                    True,
+                    self.tr("Checking for potential referenced data..."),
                 )
                 self.schema_configuration_page.setComplete(False)
                 if (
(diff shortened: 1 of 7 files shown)


signedav:
Any inputs on that @geckoGIS ?

QgisModelBaker/gui/workflow_wizard/workflow_wizard.py
```diff
@@ -670,10 +692,20 @@ def append_dropped_files(self, dropped_files, dropped_ini_files):
                 dropped_ini_files[0]
             )
 
+    def busy(self, page, busy, text="Busy..."):
+        page.setEnabled(not busy)
+        if busy:
+            self.log_panel.busy_bar.setFormat(text)
+            self.log_panel.busy_bar.setVisible(True)
+        else:
+            self.log_panel.busy_bar.setVisible(False)
```
domi4484, line 701:
```suggestion
        page.setEnabled(not busy)
        self.log_panel.busy_bar.setVisible(busy)
        if busy:
            self.log_panel.busy_bar.setFormat(text)
        else:
```

domi4484 (approved):
Code looks good to me. Just one question which code makes the progressbar moving? I can't see it :)

signedav:
> Just one question which code makes the progressbar moving? I can't see it :)

That's pure magic 🧙 It's the `QProgressBar.setRange(0,0)` - Haven't knew it untill today neither...

domi4484 (approved):

PR #860, opened 2023-12-06
fix wrong negation on transaction mode

signedav:


signedav:
too much negations :-)

domi4484 (approved):

PR #866, opened 2023-12-08
Checkbox to use superuser in TID Manager

signedav:
In case of PostgreSQL provider, there is a checkbox to perform set (set in DB) and reset (reset in GUI by current value from DB) as super user (configured in the general Model Baker settings).

image

If the credentials have not the rights (no matter if superuser or not) an error occurs, and the value in the GUI is reset to the last correctly read value (in case it could never been read from the DB it's "-").

Special behavior on resetting the values (in the gui) with wrong credentials -> it will set the value to "-" (because we cannot know what is the current value). On the other hand, when we want to set the value with wrong credentials, it's reset to the last loaded value (in the current session). This means:
- we choose "Reset Values" with correct credentials -> Set to 15 in GUI
- we change it to 16 -> Set to 16 in GUI
- we store with wrong credentials -> Set to 15 in GUI
But when:
- we choose "Reset Values" with wrong credentials -> Set to "-" in GUI
- we change it to 16 -> Set to 16 in GUI 
- we store with wrong credentials -> Set to "-" in GUI


QgisModelBaker/gui/panel/set_sequence_panel.py
```diff
@@ -29,18 +31,59 @@ class SetSequencePanel(QWidget, WIDGET_UI):
...
+                result, message = False, None
+                db_connector = db_utils.get_db_connector(self.configuration)
+                if db_connector:
+                    result, message = db_connector.set_ili2db_sequence_value(
+                        self.sequence_value_edit.value()
+                    )
+                if not result:
+                    # reset value
+                    if self.last_loaded_sequence_value is None:
+                        self.sequence_value_edit.setSpecialValueText("-")
```
domi4484, line 79:
```suggestion
                        self.sequence_value_edit.setValue(0)
```
specialvalue text is specified in the constructor

domi4484: or even better `self.sequence_value_edit.setValue(self.sequence_value_edit.minimum())`

On my crusade against literals ⚔️ 😅

PR #868, opened 2023-12-13
Name OID/TID strictly OID to avoid confusions

signedav:
- [x] Screenshots


signedav:
@domi4484 Thanks. Hope it's the last one for a moment.

docs/background_info/oid_tid_generator.md
```diff
@@ -110,13 +110,13 @@ In the QGIS Project the default value expression for `t_ili_tid` therefore is:
 | City_V1.TypeID | <code>'_' &#124;&#124; uuid('WithoutBraces')</code> |
 | not defined | <code>'_' &#124;&#124; uuid('WithoutBraces')</code> |
 
-### TID (OID) Manager
+### OID Manager
 
 Since the user has to be able to edit those values, they are provided in the GUI.
 
-Additionally to the [page implemented in the wizard](../../user_guide/import_workflow/#tid-oid-values) the configuration of the TIDs can be made on existing QGIS Projects.
+Additionally to the [page implemented in the wizard](../../user_guide/import_workflow/#tid-oid-values) the configuration of the OIDs can be made on existing QGIS Projects.
```
domi4484, line 117: How does those links work? They search for a chapter with that name using the `-` as a wildchar?

In that case the link should become `../../user_guide/import_workflow/#oid-values`

In other places as well... maybe there is a pipeline action that can check those internal links?

domi4484 (approved):

docs/background_info/oid_tid_generator.md
signedav: Yes. You are right, it should be #oid-values The - is representing a space.

Good idea to check the links in a pipeline. I wonder if there is something existing for that :thinking:

domi4484: I think Adrien did something in the talking repo. Don't know if it can be used here

PR #870, opened 2023-12-15
accept xml files in wizard

signedav:
just not on dropping in qgis directly


domi4484 (approved):
Ah yes, this clearly needs to be there

PR #873, opened 2023-12-21
Keep Settings in UsabILIty Exporter on Navigation

### LINKED ISSUE #746: UsabILItyHub Topping Exporter Improvements
### UI/UX
- [ ] On the categories it would be cool to have a "select/unselect all"
- [x] The categories window could be enlarged a little
- [x] On going back to models, the selection of the export sources (like qml, qlr etc.) are reset, while the categories are not

### Features
- [x] Transaction mode is not handled by the exporter (properties)
- [x] iliname or tablename is not handled by the exporter, so renaming is not yet integrated properly -> This is implemented with tablename and geometry column. I guess this is sufficient.

signedav:
On going forward and back navigation in the wizard the checked states in...
- model page
- layer tree page
- additive settings (variables, print layouts etc)

... persists.

And a "select all" on categories...

Resolves #746

### DIFF
--- a/QgisModelBaker/gui/topping_wizard/additives_page.py
+++ b/QgisModelBaker/gui/topping_wizard/additives_page.py
@@ -18,7 +18,6 @@
 """
 
 from qgis.core import QgsProject
-from qgis.PyQt.QtCore import Qt
 from qgis.PyQt.QtWidgets import QWizardPage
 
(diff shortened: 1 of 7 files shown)


domi4484 (approved):
Looks good to me

domi4484 (approved):
Looks good to me

PR #877, opened 2024-02-09
Wait for model refresh on local repository topping receive

signedav:
We had the functionality where we wait in a loop for the topping files to be downloaded. When there is a local repository, we did not wait for download (because the files where already there). But the model (where the mapping from id to file location is done) was sometimes not refreshed yet. That's why we wait for the signal in every case now.

And I renamed the signal in the library to avoid confusion.


domi4484 (approved):

signedav:
Waiting for new Library release after https://github.com/opengisch/QgisModelBakerLibrary/pull/85

PR #904, opened 2024-04-29
Add tag modelbaker

domi4484:
For people like me that always search for "modelbaker" without space in the plugin manager


signedav:
No idea if it would work. If yes, then go for it and add 'QgisModelBaker' as well.

signedav (approved):

signedav:
Thanks

PR #914, opened 2024-05-30
Smart 1 project optimization

signedav:
Most is done in the backend, but in the strategy selection combobox only HIDE and NONE should be provided (no GROUP).

As well the texts should describe the use case with smart1.

image

With this it reads the smart-inheritance mapping that has been chosen for the current schema and updates the combobox accordingly.

Requires https://github.com/opengisch/QgisModelBakerLibrary/pull/93


gacarrillor (approved):

docs/docs/background_info/extended_models_optimization.md
```diff
@@ -36,7 +36,7 @@ What to do then? Well, you have always the option for the [NONE-strategy](../../
 
 ## Strategies
 
-Let's check out the following examle model and it's implementation according to the strategies.
+Let's check out the following examle model and it's implementation according to the strategies on a database created with ***smart2intehritance***.
```
gacarrillor, line 39: *example

PR #919, opened 2024-06-04
Implement override decorator to make it clear what methods override a superclass method

gacarrillor:


signedav:
Cool. Btw. do you plan to have a look at https://github.com/opengisch/QgisModelBakerLibrary and check if it would make sense there as well (and if so, then where). I know that it might not be possible to implement it there everywhere though.

gacarrillor:
> Cool. Btw. do you plan to have a look at https://github.com/opengisch/QgisModelBakerLibrary and check if it would make sense there as well (and if so, then where)? I know that it might not be possible to implement it there everywhere though.
> 
> Edit: Maybe other stuff makes more sense in the backend library (are higher priority) like e.g. return type definitions etc.

Yeap, I was going to continue with the lib. Thanks for the pointers, perhaps we can add both override and type definitions, let's see if time allows it.

signedav (approved):

PR #918, opened 2024-06-04
[gui] Remember previous settings when using a PG service

gacarrillor:
**What is changed**

When opening the dialog, if there is a service stored in QSettings, we set it in the first place (before it was set in the last place), and after that we check for QSettings that complement the PG service settings.

**Other details**

Note the current PG service definition will take precedence over corresponding stored QSettings.

Also note that SSLMode is not being stored in QSettings (see https://github.com/opengisch/QgisModelBakerLibrary/blob/412e60ccf4a0cf96432470c49d3eb725489626de/modelbaker/db_factory/pg_command_config_manager.py#L149-L173). 

[PR #96](https://github.com/opengisch/QgisModelBakerLibrary/pull/96) to the underlying lib fixes that.

-------------

Fix #709

PR #922, opened 2024-06-06
[tests] Test that PG service settings are remembered

gacarrillor:
As well as additional settings given by users like schema, ssl-mode and auth config.

Note:
Added a prefix on a test utility function to avoid the `SKIPPED (This is a utility function, not a test function)` notice when running tests.

Follow-up #918

### DIFF
--- a/QgisModelBaker/tests/utils.py
+++ b/QgisModelBaker/tests/utils.py
@@ -4,6 +4,6 @@ import pytest
 
 
 @pytest.mark.skip("This is a utility function, not a test function")
-def _testdata_path(path):
+def path_testdata(path):
     basepath = os.path.dirname(os.path.abspath(__file__))
     return os.path.join(basepath, "testdata", path)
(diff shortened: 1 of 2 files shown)


QgisModelBaker/tests/utils.py
```diff
@@ -4,6 +4,6 @@
 
 
 @pytest.mark.skip("This is a utility function, not a test function")
-def testdata_path(path):
+def _testdata_path(path):
```
signedav, line 7: Could it be something else than an `_`, since for me it's an indicator for "private" functions.  What about `path_testdata` ? As far as I understand it just must not start with "test", right?

gacarrillor: Correct, addressed in https://github.com/opengisch/QgisModelBaker/pull/922/commits/085c0293d40099533464742b78a81caa492f3b11

signedav (approved):

PR #923, opened 2024-06-10
[export] Include the basket's attachmentkey to the basket selection panel

### LINKED ISSUE #600: Identifying basket at export
When a dataset has baskets, it's difiicult to choose the right one at export, as topic can be the same and t_ili_id is often an UUID. Would be nice to have attachmentkey listed as well, so it's more easy to choose the right basket. Assume same applies to import (didn't test yet).

image

gacarrillor:
…, so that it's easier to identify them.

Requires: https://github.com/opengisch/QgisModelBakerLibrary/pull/100

Fix #600

### DIFF
--- a/QgisModelBaker/gui/dataset_manager.py
+++ b/QgisModelBaker/gui/dataset_manager.py
@@ -38,7 +38,9 @@ DIALOG_UI = gui_utils.get_ui_class("dataset_manager.ui")
 
 
 class DatasetManagerDialog(QDialog, DIALOG_UI):
-    def __init__(self, iface, parent=None, wizard_embedded=False):
+    def __init__(
+        self, iface, parent=None, wizard_embedded=False, basket_integration=True
+    ):
 
         QDialog.__init__(self, parent)
         self.iface = iface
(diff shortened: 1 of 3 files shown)


signedav (approved):

gacarrillor:
@signedav, before merging this, please have a look at the last commit and let me know if it's good for you. See: https://github.com/opengisch/QgisModelBaker/issues/600#issuecomment-2163382402

signedav:
I think I would have preferred to pass an argument to the dialog (not a context maybe, but just a bool like `basket_integration` or similar) but I guess it's fine like this as well. Good work!

gacarrillor:
> I think I would have preferred to pass an argument to the dialog (not a context maybe, but just a bool like `basket_integration` or similar) but I guess it's fine like this as well. Good work!

Yeap, that would be better indeed. Applied in https://github.com/opengisch/QgisModelBaker/pull/923/commits/62a644a9f7665064bdb3d1a7232a5479d315d291, thanks!

signedav:
Oups. Just noticed there is in fact already such a parameter: `wizard_embedded`

gacarrillor:
> Oups. Just noticed there is in fact already such a parameter: `wizard_embedded`

Ouch, should I open another PR to use `wizard_embedded` instead?

signedav:
yes. sorry, I haven't seen it before.

gacarrillor:
> yes. sorry, I haven't seen it before.

No problem, I also overlooked it.
Done in https://github.com/opengisch/QgisModelBaker/pull/934

PR #936, opened 2024-06-21
Consequent use of database settings in Export / Dataset Manager

### LINKED ISSUE #847: Database Settings in Export / Dataset Manager / Validator
The current implementation is like this. When opening...
- ... Wizard > Export: We see the last used db connection
- ... Dataset Manager: We see the last used db connection
- ... Validator: We have (not editable) the db connection of the currently selected layer

image

This behavior is a little confusing. Both ways have pros and cons. Pro on the last used db connection and having it editable is that you don't need to have this schema in the current project. The con is, that you have to set it manually if it does not correspond to your current project. 

### My thoughts
I like, that we cannot edit the db-connection in the Validator. Because the results points then to the data in the current project so it would be very misleading. So I would not change there anything.

I find it confusing that I open the Dataset Manager or the Wizard > Export and it does not correspond to the schema I have in my current project (what is mostly only one), but it's fine to be able to change them.

### Question

What would be the best behavior on Dataset Manager and Wizard Export?

1. Connection should be taken over *automatically* from the currently selected layer's datasource
3. Connection should be taken over *on request* (means by button) from the currently selected layer's datasource
4. It's perfect like it is now...
5. other

Any opinions?

signedav:
## Dataset Manager

Because you barely want to edit datasets on another datasource than the one you are currently using, Model Baker detects now automatically the source of your project (according to the selected layer). You cannot edit the datasource in the Dataset Manager anymore. We removed it because it leaded to confusion.

And what if my current layer is not a valid layer (e.g. a group or a WMS)?
> Then it goes through all the layers and takes the first valid one.

And what if I have no project opened at all? 
> Then it takes the last used connection (used anywhere - not only in Dataset Manager)

Screenshot from 2024-06-21 13-04-34

### Without an opened project

Screenshot from 2024-06-21 13-05-16

### Embedded in the Wizard (takes the setting from the import)

Screenshot from 2024-06-21 12-59-45

### Detecting a data source but without basket handling
Screenshot from 2024-06-21 13-07-25

### Some thoughts I made here

I like that with Model Baker the QGIS Window is still responsive. But we had to decide, what functions in Model Baker Window (Wizard or Dataset Manager) should be triggered on interactions in the QGIS Window. In Validaton it's clear, there you change the db-settings by changing the layer. For the Dataset Manager we get now the db-settings from the current layer as well, but I avoided that it triggers a layer change (because this could lead to faults and confusions). In case you opened with a wrongly activated layer, you have to change the db-settings by selecting the correct one and close and open the Dataset Manager again.

## Export

On export it does the same. It checks for layers in the current project and takes their data source. If no project or valid layer present, it takes the last used connection.

## And on import / generate

After some thoughts I decided to adapt the behavior here. Although on schema import you usually create a new schema / gpkg at the moment it just takes the last used connection, what is not better than take the one from the active project. And additionally I assume, that mostly the other parameters (like the db / host etc. or on GeoPackage the directory) are the same like the one of the opened project.

[Screencast from 21.06.2024 14:44:33.webm](https://github.com/opengisch/QgisModelBaker/assets/28384354/3bd09f36-0c73-4f99-aa09-e4879d0714b4)

And on Generate you have no project opened usually. But better have it everywhere in the same way...

Resolves #847

### DIFF
--- a/QgisModelBaker/gui/dataset_manager.py
+++ b/QgisModelBaker/gui/dataset_manager.py
@@ -17,21 +17,25 @@
 """
 
 from qgis.core import QgsApplication, QgsMapLayer, QgsProject
-from qgis.PyQt.QtCore import QSettings, Qt, QTimer
-from qgis.PyQt.QtWidgets import QDialog, QHeaderView, QMessageBox, QTableView
+from qgis.gui import QgsMessageBar
+from qgis.PyQt.QtCore import QSettings, Qt
+from qgis.PyQt.QtWidgets import (
+    QDialog,
+    QHeaderView,
+    QMessageBox,
+    QSizePolicy,
+    QTableView,
+)
 
 import QgisModelBaker.libs.modelbaker.utils.db_utils as db_utils
 from QgisModelBaker.gui.basket_manager import BasketManagerDialog
 from QgisModelBaker.gui.edit_dataset_name import EditDatasetDialog
-from QgisModelBaker.gui.panel import db_panel_utils
 from QgisModelBaker.libs.modelbaker.db_factory.db_simple_factory import DbSimpleFactory
 from QgisModelBaker.libs.modelbaker.iliwrapper.globals import DbIliMode
 from QgisModelBaker.libs.modelbaker.iliwrapper.ili2dbconfig import (
     Ili2DbCommandConfiguration,
 )
-from QgisModelBaker.libs.modelbaker.utils.globals import DbActionType
 from QgisModelBaker.utils import gui_utils
-from QgisModelBaker.utils.globals import displayDbIliMode
 from QgisModelBaker.utils.gui_utils import DatasetModel
 
 DIALOG_UI = gui_utils.get_ui_class("dataset_manager.ui")
@@ -42,28 +46,17 @@ class DatasetManagerDialog(QDialog, DIALOG_UI):
 
         QDialog.__init__(self, parent)
         self.iface = iface
+        self.embedded = wizard_embedded
         self._close_editing()
 
         self.setupUi(self)
         self.buttonBox.accepted.connect(self._accepted)
         self.buttonBox.rejected.connect(self._rejected)
+        self.bar = QgsMessageBar()
+        self.bar.setSizePolicy(QSizePolicy.Minimum, QSizePolicy.Expanding)
+        self.info_layout.addWidget(self.bar, 0, Qt.AlignTop)
 
-        self.type_combo_box.clear()
-        self._lst_panel = dict()
         self.db_simple_factory = DbSimpleFactory()
-
-        for db_id in self.db_simple_factory.get_db_list(False):
-            self.type_combo_box.addItem(displayDbIliMode[db_id], db_id)
-            item_panel = db_panel_utils.get_config_panel(
-                db_id, self, DbActionType.EXPORT
-            )
-            self._lst_panel[db_id] = item_panel
-            self.db_layout.addWidget(item_panel)
-
-        self.type_combo_box.currentIndexChanged.connect(self._type_changed)
-
-        # when opened by the wizard it uses the current db connection settings and should not be changeable
-        self.db_frame.setHidden(wizard_embedded)
 
         self.dataset_model = DatasetModel()
         self.dataset_tableview.horizontalHeader().setSectionResizeMode(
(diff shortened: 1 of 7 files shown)


QgisModelBaker/gui/dataset_manager.py
```diff
@@ -74,37 +67,26 @@ def __init__(self, iface, parent=None, wizard_embedded=False):
...
         self.add_button.clicked.connect(self._add_dataset)
         self.edit_button.clicked.connect(self._edit_dataset)
         self.dataset_tableview.selectionModel().selectionChanged.connect(
             lambda: self._enable_dataset_handling(True)
         )
 
-        if wizard_embedded:
-            self.basket_manager_button.clicked.connect(self._open_basket_manager)
-        else:
+        if self.embedded:
```
gacarrillor, line 76: Oops, my bad here. Thanks for the fix.

QgisModelBaker/gui/dataset_manager.py
```diff
@@ -213,45 +183,94 @@ def _jump_to_entry(self, datasetname):
...
-
-    def _updated_configuration(self):
-        configuration = Ili2DbCommandConfiguration()
-
-        mode = self.type_combo_box.currentData()
-        self._lst_panel[mode].get_fields(configuration)
-
         configuration.tool = mode
-        configuration.db_ili_version = db_utils.db_ili_version(configuration)
+        print(configuration.dbschema)
```
gacarrillor, line 261: Leftover?

gacarrillor (approved):
Nice move!

In terms of UX, I think it would be handy for the users to have a message box warning them of a db-settings change when selecting another layer in QGIS layer tree. For instance, "Would you like to change your connection settings to the newly selected layer (database: MyDB, schema: MySchema)? Yes/No". This would make them aware of the impact of QGIS interactions while having MB dialogs/dock widgets opened.

After they had seen this message box a couple of times, they will get used to potential db-settings changes, so the message box could offer a checkbox to avoid being shown in the future (e.g., "Don't show me this dialog anymore.")

Don't know if it's too much, though,

signedav:
> In terms of UX, I think it would be handy for the users to have a message box warning them of a db-settings change when selecting another layer in QGIS layer tree. For instance, "Would you like to change your connection settings to the newly selected layer (database: MyDB, schema: MySchema)? Yes/No". This would make them aware of the impact of QGIS interactions while having MB dialogs/dock widgets opened.
> 
> After they had seen this message box a couple of times, they will get used to potential db-settings changes, so the message box could offer a checkbox to avoid being shown in the future (e.g., "Don't show me this dialog anymore.")
> 
> Don't know if it's too much, though,

Well I first had it like this - that the source gets changed when changing the layer - but then I decided against it. Now it's like this, that when the Dataset Manager opens, it takes the current layer and does not change anymore (on any changes in the layer-tree). My thoughts where:
- User opens the Dataset Manager and sees the recognized source. Mostly they don't care at all where it comes from, but it is - mostly - the correct one. I see the use case where people want to click through the layer-tree and change the datasource accordingly as a quite rare one.
- Mostly they have only one data source anyway in the project. But in case there are multiple and they click around in the layer tree, it could suddenly recognize another source and the user does not get it and might break something accidentally. There it would need your warning-popup or - as it is now - not a change of the datasource at all.

QgisModelBaker/gui/dataset_manager.py
signedav: yes

gacarrillor:
> > In terms of UX, I think it would be handy for the users to have a message box warning them of a db-settings change when selecting another layer in QGIS layer tree. For instance, "Would you like to change your connection settings to the newly selected layer (database: MyDB, schema: MySchema)? Yes/No". This would make them aware of the impact of QGIS interactions while having MB dialogs/dock widgets opened.
> > After they had seen this message box a couple of times, they will get used to potential db-settings changes, so the message box could offer a checkbox to avoid being shown in the future (e.g., "Don't show me this dialog anymore.")
> > Don't know if it's too much, though,
> 
> Well I first had it like this - that the source gets changed when changing the layer - but then I decided against it. Now it's like this, that when the Dataset Manager opens, it takes the current layer and does not change anymore (on any changes in the layer-tree). My thoughts where:
> 
>     * User opens the Dataset Manager and sees the recognized source. Mostly they don't care at all where it comes from, but it is - mostly - the correct one. I see the use case where people want to click through the layer-tree and change the datasource accordingly as a quite rare one.
> 
>     * Mostly they have only one data source anyway in the project. But in case there are multiple and they click around in the layer tree, it could suddenly recognize another source and the user does not get it and might break something accidentally. There it would need your warning-popup or - as it is now - not a change of the datasource at all.

Yeap, for the Dataset Manager I see it's really clear now with the messages you've added. That gives a lot of context for users.
The comment was more general, pointing to any immediate change in an open MD dialog/dockwidget after a QGIS interaction (i.e., active layer changed). If I understand well, with this PR, an immediate db-settings change will only happen in the Validator's dock widget, since for other dialogs the automatic db-settings change will be blocked (thus, no need for a message box).

For the validation, I guess we can live with the current situation, and only in case you consider it'd be confusing for users, use a warning message box.

signedav:
> If I understand well, with this PR, an immediate db-settings change will only happen in the Validator's dock widget, since for other dialogs the automatic db-settings change will be blocked

Yes.

> For the validation, I guess we can live with the current situation, and only in case you consider it'd be confusing for users, use a warning message box.

Agree. I think there it's not a big issue but good to keep an eye on that. Thanks for your thoughts.

PR #928, opened 2024-06-13
[ux] Improve night/dark theme mode

gacarrillor:
+ [x] Improve color definition for log texts, taking into account the current theme (palette) and adjusting some colors to work well on both environments light and dark.
 + [x] DB panels' input widgets: 
    + [x] Background color according to validator state.
 + [x] Topping wizard's structure preview now visible in night mode.

Requires https://github.com/opengisch/QgisModelBakerLibrary/pull/101

Screenshots of new look and feel (default theme added for reference):

Screenshot from 2024-06-13 10-58-19

Screenshot from 2024-06-13 11-04-31

Screenshot from 2024-06-13 11-06-12

Screenshot from 2024-06-13 10-51-32

Screenshot from 2024-06-13 11-05-43

Screenshot from 2024-06-13 17-20-52

Screenshot from 2024-06-13 17-17-02

Screenshot from 2024-06-13 18-03-09

Screenshot from 2024-06-13 18-03-21

Fix #568

PR #937, opened 2024-06-21
[ux] Night Mapping theme improvements follow-up

gacarrillor:
+ [x] Make sure text on ProgressBars is visible also for Night Mapping theme.
 + [x] Model (.ili) icon

Some screenshots:

Screenshot from 2024-06-24 10-25-51

image

image

image

Before (Model icon):

image

--------------

Follow-up #928


gacarrillor:
Regarding the `LayerModel` items, this is how they look like in non-default QGIS themes:

image

image

The `LayerModel.data()`'s returned value for the `Qt.BackgroundRole` role is being overridden by a global stylesheet set by the non-default QGIS themes (like `Night Mapping` theme):

```
QTreeView::item, QTreeView::branch {
	background: transparent;
	color: @textlight;
}
```

One might expect that even having a global stylesheet, one could set the item background via Qt.BackgroundRole, but the current behavior is well known in Qt and considered as a feature, rather than a bug. For reference, some related links:

https://forum.qt.io/topic/95940/model-backgroundrole-overridden-by-style-sheet
https://bugreports.qt.io/browse/QTBUG-70100

## Suggestion
Instead of using colors, add a text clue for non-INTERLIS layers (e.g., `My Layer*`) and in the legend below the `QTreeView`, replace color symbols by something like `*: Non-INTERLIS layer`.

gacarrillor:
What's missing couldn't be solved due to a Qt bug and was posted as an issue: https://github.com/opengisch/QgisModelBaker/issues/939

QgisModelBaker/images/file_types/model.png
signedav, line 1: Which one is the used one?

QgisModelBaker/utils/gui_utils.py
```diff
@@ -115,45 +115,57 @@ class DropMode(Enum):
...
 
 # style
-ORANGE = "#D1C28E"
+ORANGE = "#D1C28E"  # Not used
 GREEN = "#A1DE9B"
 BLUE = "#9BCADE"
-PURPLE = "#B18BC9"
-RED = "#EBB3A4"
+PURPLE = "#B18BC9"  # Not used
+RED = "#ff0000" if QgsApplication.themeName() == "Night Mapping" else "#EBB3A4"
```
signedav, line 122: Can we have softer colors here. Similar to the bright theme. While there we had more white in it (pastel colors) maybe with the dark more grayish colors could work...

QgisModelBaker/images/file_types/model.png
gacarrillor: Sorry, I uploaded the wrong icon version.
Just updated it and added a before screenshot for comparison.

QgisModelBaker/utils/gui_utils.py
gacarrillor: The orange color is also used by the palette, and actually used in all toolbars for intermediate steps.

Would this red work for you?

image

What about this green color? (Compare it with the selected dataset color)

image

signedav: Yeah this is nice. The other made me leave through the emergency escape everytime when the validation failed ;-)

Orange and green is fine. Thanks.

signedav (approved):

signedav:
Here a nice screeny maybe for the documenation:
Screenshot from 2024-06-27 12-18-24

Looks really nice!

gacarrillor:
> Here a nice screeny maybe for the documenation

Nice! Yeah, it would be cool to advertise multi-theme support.

PR #966, opened 2024-10-07
Topping Exporter: Consider the layer type for style categories

signedav:
And with newer QGIS use the exposed QGIS model and delegate to make it beautiful: 

image

Requires https://github.com/qgis/QGIS/pull/58729


gacarrillor (approved):
Nice!
I'm not 100% sure of my review here, since I couldn't test it in GUI, but in general, it looks good to me.

Issue #531, opened 2021-09-03
Trouble creating relations in GPKG to multigeometry tables


signedav:
While in PG three layers with the same tables are created, in GPKG it creates three tables. The Relation to it are brocken then.

See example:
```
INTERLIS 2.3;

MODEL GeometriesStruct (en) 
AT "https://signedav.github.io/usabilitydave/models"
VERSION "2020-06-22" =
  
    IMPORTS GeometryCHLV95_V1, Units;
  
    DOMAIN
        Line = POLYLINE WITH (STRAIGHTS) VERTEX GeometryCHLV95_V1.Coord2;
        Surface = SURFACE WITH (STRAIGHTS) VERTEX GeometryCHLV95_V1.Coord2 WITHOUT OVERLAPS > 0.005;
        
        STRUCTURE Geometry =
            Points: GeometryCHLV95_V1.Coord2;
            Lines: Line;
            Surfaces: Surface;
            Remark : TEXT*20;
        END Geometry; 

    TOPIC Assets =

        CLASS Asset =
            Name: TEXT;
            Geometries: BAG {0..*} OF Geometry;
        END Asset;
        
    END Assets;

END GeometriesStruct.
```

signedav:
In fact we get a relation to the point layer (containing remarks). And in the point layer a relation to the other geometry layers.

But e.g. in the KbS_LV95_V1_4 we have "Belasteter_Standort" -> proper polygon layer and "Belasteter_Standort (Geo_Lage_Punkt)" - How is this managed on working on those data? Is there a sufficient solution for multi-geom-classes with GPKG at all? 

@romefi @m-kuhn any idea?

m-kuhn:
There was a discussion (with @claeis among others) to create multi-geometry gpkgs with ili2gpkg and to be able to edit these with qgis/ogr.

This would be clearly an extension to the gpkg standard and other software would not be able to properly interact with gpkgs of this format. Still it might be interesting for workflows which are only withing qgis/ogr/ili2gpkg.

claeis:
It would require a alternative/extension to the gpkg metatable gpkg_contents (a new column column_name) and a change in the constraints of metatable gpkg_geometry_columns (so that they include column column_name)

m-kuhn:
> It would require a alternative/extension to the gpkg metatable gpkg_contents (a new column column_name) and a change in the constraints of metatable gpkg_geometry_columns (so that they include column column_name)

[...]

PR #970, opened 2024-10-21
Concern multiple geometries per GeoPackage Table Frontend

signedav:
## The setting for `gpkgMultiGeomPerTable` on Schema Import

Screenshot from 2024-10-30 22-06-22

Otherwise the setting is invisible and the parameter is `false`

This does only work with a sufficient GDAL version \* Otherwise it's invisible and false.

## Generating QGIS Project

Reading such a project does only work with a sufficient GDAL version \* If the schema has been created with the `gpkgMultiGeomPerTable` *and* it contains tables with multiple geometry columns, but the current **GDAL version is not sufficient**, we block the creation of a QGIS Project.

Screenshot from 2024-10-30 22-08-58

When we have a **sufficient** GDAL version \* we can create a QGIS project, but need to warn, that others reading this project might have trouble.

image

## GDAL Version 3.8

This functionality is not standardized and does only work with GDAL Version >= 3.8

Windows Builds have it since 3.36 but Ubuntu from the current repositories not yet.

When a schema is created with this parameter  `gpkgMultiGeomPerTable`  and the Model contains multiple geometries per table,  it can neither create a project on QGIS with GDAL<3.8 nor read a project that has been created on a QGIS with GDAL >= 3.8. 

If no multiple geometries per table, it still can create a project from it by GDAL<3.8 or open such a project.

## Night mode:

image

Requires https://github.com/opengisch/QgisModelBakerLibrary/pull/108

Resolves https://github.com/opengisch/QgisModelBaker/issues/531

### DIFF
--- a/QgisModelBaker/gui/workflow_wizard/project_creation_page.py
+++ b/QgisModelBaker/gui/workflow_wizard/project_creation_page.py
@@ -22,7 +22,6 @@ import os
 import re
 
 import yaml
-
 from osgeo import gdal
 from qgis.core import Qgis, QgsApplication, QgsProject
 from qgis.PyQt.QtCore import Qt
--- a/QgisModelBaker/gui/workflow_wizard/project_creation_page.py
+++ b/QgisModelBaker/gui/workflow_wizard/project_creation_page.py
@@ -22,6 +22,8 @@ import os
 import re
 
 import yaml
+
+from osgeo import gdal
 from qgis.core import Qgis, QgsApplication, QgsProject
 from qgis.PyQt.QtCore import Qt
 from qgis.PyQt.QtWidgets import QCompleter, QWizardPage
@@ -95,6 +97,7 @@ class ProjectCreationPage(QWizardPage, PAGE_UI):
         self.fileValidator = gui_utils.FileValidator(
             pattern=["*." + ext for ext in self.ValidExtensions], allow_empty=False
         )
+        self.gpkg_multigeometry_info.setVisible(False)
 
     def isComplete(self):
         return self.is_complete
@@ -129,6 +132,8 @@ class ProjectCreationPage(QWizardPage, PAGE_UI):
             self.existing_topping_checkbox.setChecked(True)
         else:
             self._use_existing(False)
+
+        self.gpkg_multigeometry_info.setVisible(self._multigeom_gpkg())
 
         self.workflow_wizard.busy(self, False)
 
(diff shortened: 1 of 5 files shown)


signedav:
@gacarrillor I guess this will lead to conflicts with the mulitlanguage GUI, but i guess it's an easy one to resolve...

QgisModelBaker/gui/workflow_wizard/project_creation_page.py
```diff
@@ -712,6 +716,50 @@ def _inheritance(self):
...
+                </body></html>
+                """.format(
+                    qgis_version=Qgis.QGIS_VERSION,
+                    gdal_version=gdal.VersionInfo("RELEASE_NAME"),
+                )
+            )
+        return True
+
+    def _multiple_geometry_gpkg_table(self):
+        if self.db_connector.multiple_geometry_tables() > 0:
```
gacarrillor, line 758: I think this should be 

`if len(self.db_connector.multiple_geometry_tables()) > 0:`

gacarrillor: And could be called in line 725?
(Apparently it's not being used anywhere)

QgisModelBaker/gui/workflow_wizard/project_creation_page.py
```diff
@@ -712,6 +716,50 @@ def _inheritance(self):
...
+            return True
+
+        return false
```
gacarrillor, line 761: `NameError: name 'false' is not defined` :)

gacarrillor (approved):
Concerning the orange frames (and the tooltip):

image

I'd suggest that we remove the QGIS version from these messages or at least that we move it after we've shown the installed GDAL version (perhaps even removing the bold style for the QGIS version). In that way, it'd be much easier to compare which GDAL version is needed vs. which GDAL version is installed.

QgisModelBaker/gui/workflow_wizard/project_creation_page.py
signedav: oh it's a leftover

QgisModelBaker/gui/workflow_wizard/project_creation_page.py
signedav: yes, same leftover like above

signedav:
Thanks @gacarrillor I considered that. I shortened and simplified it generally.

PR #971, opened 2024-10-25
Add support for translated models

### LINKED ISSUE #402: Support for translated models
Interlis has a concept with `TRANSLATION OF` models. A very good example are the [ARE models](https://models.geo.admin.ch/ARE/) where there are the models `Nutzungsplanung_V1_1` (German base) with translation models `PianiDiUtilizzazione_V1_1` and `PlansDAffectation_V1_1`.

The translated models are listed in the search interface of model baker, but if selected, the base language version is created. I.e. if "PlansDAffectation_V1_1" is typed in the model text edit, it is found (in French), but the generated db structure is in German.

The [`--nameLang` parameter can be used to control the language](https://github.com/claeis/ili2db/issues/347) of the generated db structure.

I see several possibilities:

 1. use the language from the model (if PlansDAffectation is selected -> use French)
 2. use the language of the system (QGIS language)
 3.a. add a combobox to allow selecting a language from a list of languages
 3.b. add a combobox only listing languages supported by the model

I think 1. would best match user expectations.
And could potentially be implemented as an ili2db switch (`--useModelLanguage`CC @claeis)?

gacarrillor:
Requires https://github.com/opengisch/QgisModelBakerLibrary/pull/113

image

Fix #402

### DIFF
--- a/QgisModelBaker/gui/workflow_wizard/project_creation_page.py
+++ b/QgisModelBaker/gui/workflow_wizard/project_creation_page.py
@@ -41,7 +41,7 @@ from QgisModelBaker.libs.modelbaker.iliwrapper.ilicache import (
 from QgisModelBaker.libs.modelbaker.utils.globals import OptimizeStrategy
 from QgisModelBaker.libs.modelbaker.utils.qt_utils import make_file_selector
 from QgisModelBaker.utils import gui_utils
-from QgisModelBaker.utils.globals import CATALOGUE_DATASETNAME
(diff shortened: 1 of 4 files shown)


signedav:
Maybe a tool tip: 

Several translations of the INTERLIS model are implemented in the database schema (`TRANSLATION OF`). Select the language in which the QGIS project is to be set up.

signedav:
image

Would it be possible to have the same size for the comboboxes?

signedav (approved):

gacarrillor:
Both suggestions (tooltip and combobox size) have been addressed.

PR #978, opened 2024-11-05
refresh validator panel when added new dataset

signedav:
i did not solve it with signals because we handle updates after closing dataset manager anyway in the main class.


domi4484 (approved):

domi4484:
@signedav checked out and tested successfully

PR #983, opened 2024-11-13
log output from ili2db in the log files

signedav:


QgisModelBaker/utils/gui_utils.py
```diff
@@ -204,18 +204,15 @@ def get_text_color_object(level: LogLevel = LogLevel.INFO) -> QColor:
     return QColor(get_text_color(level))
 
 
-def get_parsed_log_text_color(text, txt_edit) -> None:
+def get_parsed_log_text_level(text) -> LogLevel:
     textlines = text.splitlines()
```
gacarrillor, line 208: It seems a bit strange to deal with several lines here but only return the first line case.

If we don't have a multifaceted case where we get a `textlines` variable with several levels (i.e., warning + error, info + warning), please add a comment to explain that we'll get the level from the first line content.

QgisModelBaker/utils/gui_utils.py
```diff
@@ -204,18 +204,15 @@ def get_text_color_object(level: LogLevel = LogLevel.INFO) -> QColor:
...
-            txt_edit.append(textline)
+            return LogLevel.WARNING
         elif "error" in textline.lower() or "failed" in textline.lower():
-            txt_edit.setTextColor(get_text_color_object(LogLevel.FAIL))
-            txt_edit.append(textline)
+            return LogLevel.FAIL
         else:
-            txt_edit.setTextColor(get_text_color_object(LogLevel.INFO))
-            txt_edit.append(textline)
+            return LogLevel.INFO
```
gacarrillor, line 216: I'm not 100% sure that this method can be called with an empty `text` value, but should we have a fallback return value for this method (i.e., out of the for loop)?

gacarrillor (approved):
Looks good.
(I haven't tested it, though)

PR #984, opened 2024-11-19
improve log outputs

signedav:


signedav:
Merge it because there are no risks. Just log output.

PR #989, opened 2024-11-27
make difference between action type SCHEMA_IMPORT and GENERATE 

### LINKED ISSUE #987: Generate a QGIS Project: Superuser Login required when using an Authentication Configuration
It is basically exactly the same issue as in #905 but for the "Generate a QGIS Project from an existing database" workflow. Since the database isn't changed in this workflow there is no need for more privileges than 'SELECT' privileges on the database tables.

Bildschirmfoto vom 2024-11-26 21-20-27

signedav:
what fixes #987

depends on https://github.com/opengisch/QgisModelBakerLibrary/pull/123

### DIFF
--- a/scripts/package_pip_packages.sh
+++ b/scripts/package_pip_packages.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 LIBS_DIR="QgisModelBaker/libs"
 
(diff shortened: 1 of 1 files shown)


gacarrillor (approved):
Looks good to me.
(I haven't tested it, though.)

Issue #829, opened 2023-10-06
Validation Config File from the Repository


signedav:
Since this integration #828 we can store config files per project. In UsabILIty Hub we can store the project variable, means the path. When the path is already an `ilidata:key` then this could work as long as it's stored per model.

At the moment this ilidata:key needs to be entered manually (no file from the repo selectable in the line edit of the validator - not sure if it should be supported, because it would be the only functionality where we use the repos after project generation (during editing)).

But if it's a path to a file this file will not be "exported" into a topic. It's a conceptional thing. Usually we do not store files permanently on a Topping import. So where should this file be stored?

As attachment file next to the project file (but on generation there is not a project file stored - so complicated)

Write it into the project (but this is very intransparent)

Being able to export it and add it as topping-file (like the catalogue file) - after that the key is written to this specific project variable (preferred possibility)

signedav:
Would mean it's a topping exporter thing. 
1. reading of variable 
2. create a topping file of it
3. create a link to the topping file (maybe change the project variable)

Makes me think if a custom property would have been nicer than a project variable and if we can change this or if it should remain the same :thinking:

PR #998, opened 2024-12-03
Export ValidationConfig File as Topping

signedav:
Since quite some time we are able to store the path to the validation configuration file as project variable, so we have it always in the validator:

image

This variable could be exported so far to the project topping **but** if the file is not stored on a accessible drive the project using this topping could not access to it.

## Topping Export

That's why with this PR you have the option in the **Topping Exporter** to *create* a topping file from the variable with the name "validator_config" that is stored at the topping files.

image

Means we have now the validation-config.ini file in the toppings and link over the variable to the `ilidata:` key.

## Topping Import
But what happens on project create with this topping?

Nothing :-) It keeps the `ilidata:` key in the variable. This means in the Validator we see the `ilidata:` key as well. And this can be used then on validation (passing this key as `--validConfig` to ili2db).

image

## Why the file is not downloaded
This because we are not able to download and store the file, since we do not know where to store it. Next to the project? Maybe, but on generating a project there will not be a project stored. So it needs to be an interaction with the user.
### Future idea
If it is required to get the file again locally, then an implementation might be done in the validator, to download the file and store it next to the project and replace the variable with the relative path

image
(mockup)

Resolves https://github.com/opengisch/QgisModelBaker/issues/829
Requires https://github.com/opengisch/toppingmaker/pull/14 and with this a new library

### DIFF
--- a/QgisModelBaker/gui/topping_wizard/additives_page.py
+++ b/QgisModelBaker/gui/topping_wizard/additives_page.py
@@ -27,6 +27,8 @@ PAGE_UI = gui_utils.get_ui_class("topping_wizard/additives.ui")
 
 VARIABLE_PREFIX_BLACKLIST = ["default_basket", "optimize_strategy"]
 
+VALIDATION_VARIABLE = "validator_config"
+
 
 class AdditivesPage(QWizardPage, PAGE_UI):
     def __init__(self, parent, title):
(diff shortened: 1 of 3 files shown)


signedav:
https://github.com/opengisch/QgisModelBaker/issues/829

signedav:
https://github.com/opengisch/toppingmaker/pull/14

QgisModelBaker/gui/topping_wizard/additives_page.py
```diff
@@ -71,11 +73,28 @@ def initializePage(self) -> None:
...
+
+        if not (
+            self.mapthemes_model.rowCount()
+            or self.variables_model.rowCount()
+            or self.layouts_model.rowCount()
+        ):
+            self.topping_wizard.log_panel.print_info(
+                self.tr("No additive settings - go on...")
+            )
+            self.topping_wizard.next()
```
gacarrillor, line 97: Nice one

gacarrillor (approved):

PR #1026, opened 2025-05-29
Fix issue #1010 (mimeData not found in event obj)

### LINKED ISSUE #1010: Exception when dropping a non-INTERLIS GPKG into QGIS (MB installed)
Image

The drop event was done in the map canvas.

The event is a `KeyPress` (i.e., `event.type()` gives `6`) and `event.key()` gives `16777220`:

`16777220 decimal is 0x1000004, which is Qt::Key_Return` ([source](https://bugreports.qt.io/browse/QTBUG-66838))

I wasn't (as far as I know) pressing a key, just using a mouse.
A bit cryptic because I see MB is filtering out events that are not `DropEvent`.

Ubuntu 24.04
QGIS v3.41.0-Master
QgisModelBaker v7.10.6

gacarrillor:
Fix #1010

### DIFF
--- a/QgisModelBaker/qgismodelbaker.py
+++ b/QgisModelBaker/qgismodelbaker.py
@@ -571,7 +571,7 @@ class DropFileFilter(QObject):
         """
         When files are dropped, then ask to use it in the model baker.
         """
-        if event.type() == QEvent.Drop:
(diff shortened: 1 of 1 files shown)


gacarrillor:
> On what kind of files does this happen? Model Baker related files or others?

Has happened to me with non-INTERLIS GPKGs (from time to time, not always).

Issue #1011, opened 2025-02-05
Watch XTF even more quick and dirty


romefi:
You can already drag and drop XTF into QGIS and view it with Model Baker. Since INTERLIS-Connector is no longer maintained in gdal, it would be cool to be able to view an XTF with Model Baker via drag-and-drop, even if e.g. the data is incorrect or catalogs are missing. There should be an import option at the beginning where settings are preset and the wizard no longer opens. Not sure what ili2db parameters are necessary for this to work. If the data is too junky, of course it won't work.

Even cooler would be if Model Baker is not yet installed and an XTF is dragged into QGIS, a message would appear that Model Baker needs to be installed.

sjib:
Why is INTERLIS connector in gdal not maintained anymore?

romefi:
@sjib https://interlis.discourse.group/t/gdal-ohne-ili-und-jetzt/349 explains it quite well.

PR #1035, opened 2025-06-16
XTF Quick (and ofc dirty) import function

signedav:
This is introduced due to no support of INTERLIS 2.4 of GDAL. Means we cannot display the data by simply drag'n'drop how we used to before.

## QuickVisualizer

This class provides the functionality for a quick import of dropped xtf files.

It generates a GeoPackage in the temporary directory withouth any constraints and imports the data without any validation. 

> If no configuration is done for a custom model repositories, then it gathers the models from the usual http://model.interlis.ch and the xtf-directory (the directory where the dropped files come from). If there is a custom model directory set it checks this and in case it does not find the model, it adds automatically the xtf-directory.

The project layers are then generated.

The class stored in `utils/tools.py`. Not sure if it should be exposed in the modelbaker library in future, but I decided to put it here first until we know that it is what we want.

### Drop the Files

Besides the dialog asking if you want to handle such files with Model Baker in general, the following dialog appears:

image

Here you can decide to use the "usual" Wizard to continue or to visualize your data quickly. If you decide to visualize quickly, the files are imported (without constraints and validations) and the layers are created.

You get the status in the Message Bar:
image

And finally the feedback:
image

Or if there are troubles (like e.g. no model available), you get a negative feedback with a button to check the log files:
image

### Quick Visualization from the Wizard

If you choose to use the wizard (or you opened the wizard manually (without dropping files)) you still can start the Quick Visualizer with the newly added button.

image

> This means, if you mostly use the wizard and get annoyed by the first dialog, you can select there "always open wizard" and you will come directly to the wizard on file drop and you still have the posility to quickly visualize.

Btw. you can deactivate the decision you made on always opening the wizard in the options:

image

## To do

- [x] Prototype to evaluate functionality
- [x] Missing things like:
  - Model Diretories should be concerned
  - Fallback if ili not in repo
- [x] ili2db command as slim as possible, but hard to say what parameters should be cropped. In the end I think I would not crop them. Just use the defaults with `smart2`
- [x] Project Generation as slim as possible
- [x] Where should the integration be? Kind of a utils function in the lib (would make sense)
- [x] Finetune entry point - decide what we want and then clean up the code and ...
- [x] ... make consistent naming in the functions and the logs
- [x] Retrieve the xtf-files from the `SourceModel` (if the button stays at the source page)
- [x] Messages (like Status bar and Message bar)
- [x] Logs
- [x] Tests

Concerns #1011

### DIFF
--- a/QgisModelBaker/gui/drop_message.py
+++ b/QgisModelBaker/gui/drop_message.py
@@ -84,9 +84,11 @@ class DropMessageQuickDialog(QDialog, gui_utils.get_ui_class("drop_quick_message
         )
         self.info_label.setText(
             self.tr(
-                "Would you like to quickly visualize these files, or would you prefer to create a structured database using the Model Baker Wizard?\n- {} {}"
+                "Would you like to quickly visualize {}, or would you prefer to create a structured database using the Model Baker Wizard?\n- {} {}"
             ).format(
-                self.tr("files") if len(dropped_files) > 1 else self.tr("file"),
+                self.tr("these files")
+                if len(dropped_files) > 1
+                else self.tr("this file"),
                 file_list,
                 self.tr("\nand {} more...").format(len(dropped_files) - 10)
                 if len(dropped_files) > 10
--- a/QgisModelBaker/utils/tools.py
+++ b/QgisModelBaker/utils/tools.py
@@ -112,7 +112,7 @@ class QuickVisualizer(QObject):
         if len(data_files) == 0:
             self.push_message_bar(
                 self.tr(
-                    "Nothing to import with Quick XTF Backer (use the Model Baker wizard instead)."
+                    "Nothing to import with Quick Visualizer (use the Model Baker wizard instead)."
                 ),
                 False,
                 Qgis.MessageLevel.Warning,
@@ -122,7 +122,7 @@ class QuickVisualizer(QObject):
 
         if len(model_files):
             self.log(
-                f"Dropped model files are ignored {model_files} (but if not available in repo, they are used because they are in the same directory as the tranfer files)."
+                f"Dropped model files are ignored {model_files} (but if not available in repo, they are used because they are in the same directory as the transfer files)."
             )
 
         self.push_message_bar(
@@ -144,6 +144,8 @@ class QuickVisualizer(QObject):
         if (
             len(failed_files) > 0
             and self.parent.ili2db_configuration.custom_model_directories_enabled
+            and "%XTF_DIR"
+            not in self.parent.ili2db_configuration.custom_model_directories
         ):
             self.log(
                 "Retry failed XTF imports with models in the same directory as the datafile"


signedav:
@romefi According to the video, do you think this is what we want? 
Do you agree on the thoughts regarding the entry point?

sjib:
Ist dies eine Lösung für https://interlis.discourse.group/t/gdal-ohne-ili-und-jetzt/349 ? Das begrüsse ich sehr!

signedav:
Ja, genau @sjib

webrian:
Nice!
I think it is important and crucial to set the --modeldir parameter sensibly, e.g. I think %ITF_DIR should always be in there, because it happens again and again that you get data with the model as a file.

signedav:
> Nice! I think it is important and crucial to set the --modeldir parameter sensibly, e.g. I think %ITF_DIR should always be in there, because it happens again and again that you get data with the model as a file.

Yes. This is done here https://github.com/opengisch/QgisModelBaker/pull/1035/files#diff-8333eb09bd38e5a60dc1bd289456a06aa424435207353bcdfd3c4ad05c01998eR152 If the model cannot be found in the repo and no xtfdir is set in the modeldir (what basicly only happens when you configured a custom model dir) it retries with the parameter.

romefi:
@signedav looks very good to me, thank you.

I'm just not sure about the UI. The yellow "Import quick and dirty..." button is a bit confusing. Especially when your'e not used to INTERLIS. I would rather suggest a pop-up window with two buttons

- "Import quick and dirty..." -> does the quick and dirty import
- e.g. "Open wizzard to import properly" -> opens wizzard

What dou you think?

signedav:
Dialog pops up: "Import Quick and dirty or open Wizard"
Options: "Quick and Dirty" "Wizard"
Checkbox: Always open wizard

Option next to "Clear ilicache" in source selection: "Quick and Dirty"

In settings deactivate "always open wizard"

signedav:
Here we go. Updated the entry point. Thanks for your feedbacks @sjib @webrian and @romefi 

@gacarrillor Could you review this? If any questions contact me.

QgisModelBaker/gui/drop_message.py
```diff
@@ -62,3 +61,49 @@ def handle_dropped_file_configuration(self, handle_dropped):
...
+        file_list = "\n- ".join(
+            os.path.basename(path)
+            for path in dropped_files
+            if dropped_files.index(path) < 10
+        )
+        self.info_label.setText(
+            self.tr(
+                "Would you like to quickly visualize these files, or would you prefer to create a structured database using the Model Baker Wizard?\n- {} {}"
+            ).format(
+                self.tr("files") if len(dropped_files) > 1 else self.tr("file"),
```
gacarrillor, line 89: Minor suggestion: When multiple files are dropped, it seems not so readable to have the `files` word accompanying the first file base name. Perhaps we could have something like this instead?

```
Files:
- my_file1.xtf
- my_file2.xtf

```

In this way, single file drops would look like this:

```
File:
- my_file1.xtf

```

Which still looks acceptable.

QgisModelBaker/utils/tools.py
```diff
@@ -0,0 +1,271 @@
...
+
+        # one GeoPackage per file
+        status_map = {}
+        suc_files = set()
+        failed_files = set()
+
+        if len(data_files) == 0:
+            self.push_message_bar(
+                self.tr(
+                    "Nothing to import with Quick XTF Backer (use the Model Baker wizard instead)."
```
gacarrillor, line 115: *XTF Baker

QgisModelBaker/utils/tools.py
```diff
@@ -0,0 +1,271 @@
...
+                ),
+                False,
+                Qgis.MessageLevel.Warning,
+                15,
+            )
+            return suc_files, failed_files
+
+        if len(model_files):
+            self.log(
+                f"Dropped model files are ignored {model_files} (but if not available in repo, they are used because they are in the same directory as the tranfer files)."
```
gacarrillor, line 125: +transfer

QgisModelBaker/utils/tools.py
```diff
@@ -0,0 +1,271 @@
...
+            status, db_file = self.import_file(file, self.parent.ili2db_configuration)
+            status_map[file] = {}
+            status_map[file]["status"] = status
+            status_map[file]["db_file"] = db_file
+            if status:
+                suc_files.add(file)
+            else:
+                failed_files.add(file)
+
+        # this is a fallback: in case we have a custom model directory defined, it could be that we didn't found the model in the tranferfile's directory
```
gacarrillor, line 143: + transfer file's

QgisModelBaker/utils/tools.py
```diff
@@ -0,0 +1,271 @@
...
+        if (
+            len(failed_files) > 0
+            and self.parent.ili2db_configuration.custom_model_directories_enabled
+        ):
```
gacarrillor, line 147: Would it make sense for you to check if `XTF_DIR` is not yet in `custom_model_directories` before the fallback, so that perhaps we skip re-attempting if the `XTF_DIR` was already taken into account. That is, something like:

        if (
            len(failed_files) > 0
            and self.parent.ili2db_configuration.custom_model_directories_enabled
            and "%XTF_DIR" not in base_config.custom_model_directories
        ):

gacarrillor (approved):
Great addition! Only minor comments.

(Note: I didn't test it via GUI)

QgisModelBaker/utils/tools.py
signedav: got rid of this name anyway. it's "Quick Visualizer"

QgisModelBaker/utils/tools.py
signedav: good idea

QgisModelBaker/gui/drop_message.py
signedav: The mistake was, that I had not "visualize these {}, or would", because the first parameter is "files" or "file" depending on if there are more than one. Then the list. And then the "and {} more" if there are more than 10. Thanks for pointing on that :-)

Issue #1045, opened 2025-07-17
Validator: Mark warnings different than errors


signedav:
Beacuse otherwise it's hard to make a difference:

```
2025-07-17 11:08:11,572 ERROR   Warning: Table abwasserknoten(id 25282) more than one value for role AbwasserbauwerkRef; value of abwasserbauwerkref_normschacht ignored
2025-07-17 11:08:26,491 ERROR   Warning: Table abwasserknoten(id 22167) more than one value for role AbwasserbauwerkRef; value of abwasserbauwerkref_normschacht ignored
2025-07-17 11:08:26,541 ERROR   Warning: Table abwasserknoten(id 22173) more than one value for role AbwasserbauwerkRef; value of abwasserbauwerkref_normschacht ignored
2025-07-17 11:17:08,275 ERROR   Error: line 0: LKSOB_Abwasser_V1_0.Abwasser.VirtuellerAbwasserknoten: tid 0851bde6-558f-42fc-bcdd-ae0342132e4f: AbwasserbauwerkRef should associate 1 to 1 target objects (instead of 0)
2025-07-17 11:17:08,325 ERROR   Error: line 0: LKSOB_Abwasser_V1_0.Abwasser.Haltung: tid 1403eb3a-57d4-432b-a86e-cef1ad38fbca: AbwasserbauwerkRef should associate 1 to 1 target objects (instead of 0)
```

<img width="919" height="266" alt="Image" src="https://github.com/user-attachments/assets/e95b5fc7-9cb2-4bc9-b087-1a0cfeb82a57" />

signedav:
What does the warning mean btw?

And does it have impacts on the validity of the data?

signedav:
<img width="620" height="165" alt="Image" src="https://github.com/user-attachments/assets/b36f82b0-3285-4a42-b67f-bf0932e7c3c1" />

Actually messages from the type "warning" are not something defining the data as invalid. They might be interesting (like info) the ilivalidator prints, but I think I would would avoid having them in the result panel. In the end the user wants to know whether the data is valid or not (and only if not, there should be further info). In case it would be interesting to print warnings / info, then it should be a debug mode (or a possibility to open the log/xtf in the end).

PR #1053, opened 2025-08-26
Validator crossed and grayed out when error is marked as fixed

signedav:
While evaluating this #1045 I noted again that the "fixed" entries in the list with this green button are hard to see. Maybe it's because of my color-blindness but even then, it's better to have it clearer.

Screenshot from 2025-08-26 09-42-46

And name it "mark as fixed" and not "set to fixed" to avoid that the user believes they really *fix* the error.


domi4484 (approved):
Looks good to me

PR #1067, opened 2025-09-02
Do not expose baskets without t_ili_tid (BID)

### LINKED ISSUE #957: Data validation of baskets with empty t_ili_tid fails
* a database schema with three datasets, each of the dataset has one basket
* since the import was done _without_ `--importBid` the t_ili_tid column is empty, see the database table:

grafik

* the baskets in the Model Baker Data Validator widget:

grafik

the basket validation fails with:

```
  File "C:\Users/bjsvwweb/AppData/Roaming/QGIS/QGIS3\profiles\INTEGRATION/python/plugins\QgisModelBaker\gui\validate.py", line 391, in _run
    self._validator_stdout(f"Run: {validator.command(True)}")
  File "C:\Users/bjsvwweb/AppData/Roaming/QGIS/QGIS3\profiles\INTEGRATION/python/plugins\QgisModelBaker\libs\modelbaker\iliwrapper\iliexecutable.py", line 99, in command
    args = self._args(hide_password)
  File "C:\Users/bjsvwweb/AppData/Roaming/QGIS/QGIS3\profiles\INTEGRATION/python/plugins\QgisModelBaker\libs\modelbaker\iliwrapper\ilivalidator.py", line 42, in _args
    args = super()._args(hide_password)
  File "C:\Users/bjsvwweb/AppData/Roaming/QGIS/QGIS3\profiles\INTEGRATION/python/plugins\QgisModelBaker\libs\modelbaker\iliwrapper\iliexecutable.py", line 80, in _args
    return get_ili2db_args(self.configuration, hide_password)
  File "C:\Users/bjsvwweb/AppData/Roaming/QGIS/QGIS3\profiles\INTEGRATION/python/plugins\QgisModelBaker\libs\modelbaker\iliwrapper\ili2dbargs.py", line 40, in get_ili2db_args
    return configuration.to_ili2db_args(db_args)
  File "C:\Users/bjsvwweb/AppData/Roaming/QGIS/QGIS3\profiles\INTEGRATION/python/plugins\QgisModelBaker\libs\modelbaker\iliwrapper\ili2dbconfig.py", line 451, in to_ili2db_args
    self.append_args(args, ["--baskets", ";".join(self.baskets)])
TypeError: sequence item 0: expected str instance, NoneType found
```

* filling an id into the t_ili_id column:

grafik

the validation works!
It is the same issue in the export Wizard as well.

signedav:
if baskets do not have a t_ili_tid, they cannot be validated or exported.

so, we do not expose baskets without t_ili_tid to the gui. and if there is no basket, they are neither listed in the filter-combo. This resolves #957

### DIFF
--- a/QgisModelBaker/gui/panel/filter_data_panel.py
+++ b/QgisModelBaker/gui/panel/filter_data_panel.py
@@ -65,9 +65,10 @@ class FilterDataPanel(QWidget, WIDGET_UI):
             self.filter_combobox.addItem(
                 self.tr("Datasets"), gui_utils.SchemaDataFilterMode.DATASET
             )
-            self.filter_combobox.addItem(
-                self.tr("Baskets"), gui_utils.SchemaDataFilterMode.BASKET
-            )
+            if self.parent.current_baskets_model.rowCount() > 0:
+                self.filter_combobox.addItem(
+                    self.tr("Baskets"), gui_utils.SchemaDataFilterMode.BASKET
+                )
         if self.filter_combobox.itemData(stored_index):
             self.filter_combobox.setCurrentIndex(stored_index)
             if (
--- a/QgisModelBaker/utils/gui_utils.py
+++ b/QgisModelBaker/utils/gui_utils.py
@@ -1214,6 +1214,8 @@ class SchemaBasketsModel(CheckEntriesModel):
         ):
             baskets_info = db_connector.get_baskets_info()
             for record in baskets_info:
+                if not record["basket_t_ili_tid"]:
+                    continue
                 basketname = f"{record['datasetname']}-{record['topic']} ({record['basket_t_ili_tid']}) {record['attachmentkey']}"
                 basketnames.append(basketname)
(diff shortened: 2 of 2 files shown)


sjib:
> Disclaimer: Maybe I don't get the whole situation. But, generally I would be careful when hiding informations. Wouldn't it be better to gray out those baskets? maybe with a tool tip telling what needs to be fixed to be able to validate them?

Good thought. Or is there another way to see which baskets exist?

signedav:
@sjib you can see them in the Dataset / Basket Manager...

... the thought about hinding them is, that I belief that when there are no `t_ili_tid` the user are not interested in baskets.

When someone has no `t_ili_tid` in the baskets it's *only* when the data is imported with *manually removing* the parameter `--importBid` from the command. You still can work then with the data, but I doubt that you want to work with baskets then...

signedav:
I'll merge. Gray baskets are something that we can technically and conceptionally seen easily bring in later on, if required.

Issue #973, opened 2024-10-31
Clarify how to deal with available languages for translated models


gacarrillor:
Since https://github.com/opengisch/QgisModelBaker/pull/971 (and ili2db 5.2.0), MB has support for translated models.

![image](https://github.com/user-attachments/assets/7be34631-9fd0-48de-81b0-33f694b2f665)

For an imported translated model, we'll have the original model and the translated model languages available.

We need to know whether for an already imported translated model, we should have language selection or not when building the QGIS project. If that's the case, we could select the translated model language by default.

For that, we need to get both the model language and whether the model is a translation of another model (possible after https://github.com/claeis/ili2db/commit/e4379495895771db8113dd56a4fe23f3ceae5419).
 
We also need to have somewhere available the list of models that were imported in the schema import operation. This one I'm still unsure that we have it available.

Before implementing something in this respect, let's agree on what would be needed/desirable.

signedav:
With this https://github.com/opengisch/QgisModelBakerLibrary/pull/117 we already have a solution, but it would be preferable to have a more proper solution like suggested, because it's much more future proof.

PR #1087, opened 2025-10-27
Smart default of language combo

signedav:
if a translation of model is found, it should select its language for the combobox to create the project per default

Requires https://github.com/opengisch/QgisModelBakerLibrary/pull/156


signedav:
> @signedav, is this fully addressing #973?

I think so

PR #1092, opened 2025-11-04
Pass jar-location + 'plugins' as plugin dir on validation

signedav:
if there is a plugins folder alongside the jar file, we pass it as --plugins

this folder is the default folder in ilivaidator

there are still discussions about the general concept of the plugins like if ili2db should download the plugins itself and so on. Maybe we might need a setting for a custom path or whatever, but until then, we keep it static.

depends on https://github.com/opengisch/QgisModelBakerLibrary/pull/159


gacarrillor (approved):

signedav:
Pity that the zip upload does not work anymore - i need to check this out, but not now...

PR #1099, opened 2025-11-25
Fix stdout missmatch on validator

signedav:
Fix that signal is passed instead of slot.


signedav:
> Should be easier to spot when we finish this [opengisch/QgisModelBakerLibrary#99](https://github.com/opengisch/QgisModelBakerLibrary/issues/99)

Yes.

PR #1122, opened 2026-02-24
New option on creating the schema: Run without not null

signedav:
Avoiding to create mandatory not null constraints on the field. This sets disable_mandatory on the configuration and will lead to the parameter sqlEnableNull on ili2db command.

Screenshot From 2026-02-25 20-15-51

See in Doku:
> When "Run without constraints" is selected, the parameters `--sqlEnableNull` and `--sqlColsAsText` are additionally passed to ili2db, while the parameters `--createNumChecks`, `--createUnique` and `--createFk` are removed. This results in a super-tolerant database that allows for lots of data inputs when fixing invalid data or performing a migration. When "Run without not null" is selected, only the additional parameter `--sqlEnableNull` is passed, and all other constraints and column types are still considered.

Requires https://github.com/opengisch/QgisModelBakerLibrary/pull/165

### DIFF
--- a/QgisModelBaker/gui/panel/session_panel.py
+++ b/QgisModelBaker/gui/panel/session_panel.py
@@ -89,6 +89,14 @@ class SessionPanel(QWidget, WIDGET_UI):
         self.set_button_to_create_without_constraints_action.triggered.connect(
             self.set_button_to_create_without_constraints
         )
+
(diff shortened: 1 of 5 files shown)


gacarrillor (approved):
Please add a screenshot showing the new option in the GUI.

signedav:
> Please add a screenshot showing the new option in the GUI.

Done

Issue #1100, opened 2025-11-25
Automate model import settings  (Setting Prophet with ili2py)


signedav:
Users are often overwhelmed with these settings. And the worst part: they often make no sense at all for the specific model.

<img width="648" height="667" alt="Image" src="https://github.com/user-attachments/assets/422979da-24d3-470b-a9ff-80dcd60b4117" />

- When BASKET OID is defined, this setting is mandatory (no setting should be exposed)
- When the model does not have multiple geometries per table, the next setting does not make sense either
- The same applies to arcs. When there are no arcs in the model, why have a setting?
- And the extra meta attribute files (toml-files) are always the same for INTERLIS 2.3 MultiSurface or Multilingual. Why not just generate it automatically?
- If there is no TRANSLATION OF model, the language dropdown does not make sense either

So why not emit these settings by parsing the model in advance? This would be easy for a single model with no inheritance, but if it extends topics from other models it gets very, very complicated...

... but luckily we now have a tool doing all this work: ili2py 🐍

It would be possible to have this:

<img width="638" height="145" alt="Image" src="https://github.com/user-attachments/assets/0c0c8356-b481-4da4-acbd-da4f0f59c83f" />

And parts (basket, multiple geometry columns, arcs) are already done in this PoC https://github.com/opengisch/QgisModelBakerLibrary/pull/160

PR #1137, opened 2026-06-18
GUI part to control enumeration handling

signedav:
Allows us to set the enumeration handling of ili2db via GUI. Find the backend implementations here and here https://github.com/opengisch/QgisModelBakerLibrary/pull/162 and here https://github.com/opengisch/QgisModelBakerLibrary/pull/168

image

(QGIS 4 - dark theme)

### Financed 
by the QGIS Model Baker Group

### DIFF
--- a/QgisModelBaker/gui/ili2db_options.py
+++ b/QgisModelBaker/gui/ili2db_options.py
@@ -132,7 +132,15 @@ class Ili2dbOptionsDialog(QDialog, DIALOG_UI):
         self.create_basket_col_checkbox.toggled.connect(
             lambda: self._restyle_concerning_metaconfig()
         )
-
+        self.enum_tabswithid_radio_button.toggled.connect(
+            lambda: self._restyle_concerning_metaconfig()
+        )
+        self.enum_tabs_radio_button.toggled.connect(
+            lambda: self._restyle_concerning_metaconfig()
+        )
+        self.enum_singletab_radio_button.toggled.connect(
+            lambda: self._restyle_concerning_metaconfig()
+        )
         self.toml_file_line_edit.textChanged.connect(
             lambda: self._restyle_concerning_metaconfig()
         )
@@ -214,6 +222,14 @@ class Ili2dbOptionsDialog(QDialog, DIALOG_UI):
         else:
             return "smart2"
 
+    def enum_type(self):
+        if self.enum_tabs_radio_button.isChecked():
+            return "tabs"
+        elif self.enum_singletab_radio_button.isChecked():
+            return "singletab"
+        else:
+            return "tabsid"
+
     def set_toml_file_key(self, key_postfix):
         if key_postfix:
             self.toml_file_key = "QgisModelBaker/ili2db/tomlfile/" + key_postfix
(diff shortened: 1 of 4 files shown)


QgisModelBaker/gui/ili2db_options.py
```diff
@@ -265,6 +282,13 @@ def restore_configuration(self):
         create_basket_col = settings.value(
             "QgisModelBaker/ili2db/create_basket_col", defaultValue=True, type=bool
         )
+        enum_type = settings.value("QgisModelBaker/ili2db/enum_type", "smart2")
```
gacarrillor: Forgot to update the default value here?

QgisModelBaker/ui/ili2db_options.ui
```diff
@@ -315,6 +279,84 @@ strategy. Concrete classes are mapped using a NewAndSubClass strategy.</string>
...
+        </property>
+        <property name="text">
+         <string>createEnumSingleTab (stores the enumeration value in the field)</string>
+        </property>
+       </widget>
+      </item>
+      <item row="1" column="0">
+       <widget class="QRadioButton" name="enum_tabs_radio_button">
+        <property name="toolTip">
+         <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;&lt;span style=&quot; font-family:'monospace';&quot;&gt;createEnumTabs&lt;/span&gt; creates a table for each enumeration definition, containing the individual enumeration values.&lt;br/&gt;In the tables the enumeration &lt;span style=&quot; font-weight:600;&quot;&gt;values&lt;/span&gt; (&lt;span style=&quot; font-family:'monospace';&quot;&gt;iliCode&lt;/span&gt;) are stored.&lt;/p&gt;&lt;p&gt;&lt;span style=&quot; font-style:italic;&quot;&gt;Note: Inherited enumerations cannot be implemented with smart1 using this option.&lt;/span&gt;&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
```
gacarrillor, line 314: Do users get a warning somewhere else if they select `smart1` and `createEnumTabs` having extended enumerations?

I'm just wondering if we could show a warning in the QGIS message bar when configuring the project, so that in the aforementioned case, it is clear that there is something that will not work as expected.

But as I told you in the latest review, I haven't used `smart1` too much, so I'm not sure if it actually makes sense.

gacarrillor (approved):
Nice!

QgisModelBaker/gui/ili2db_options.py
signedav: Oopsy. Thanks.

QgisModelBaker/ui/ili2db_options.ui
signedav: I like the idea but because this https://github.com/opengisch/QgisModelBaker/issues/1100 is on the roadmap and where we parse the model previously and won't offer this option in the case of smart1 and inherited enums, I prefer not to make the effort (although it might be small).

gacarrillor: Cool!

Issue #993, opened 2024-11-27
UX Wizard-Roadmap / Breadcrumbs


signedav:
User mentioned, it would be nice to know "where" they are currently. Maybe kind of a road-map or breadcrumbs makes sense. 

***These here are not suggestions - rather some notes...***

The Model Baker Wizard is quite dynamic. According to the users choice it offers several possibilities 

```mermaid
stateDiagram-v2
    intro --> source
    intro --> exportdb
    intro --> generatedb
    drag --> source
    source --> db
    exportdb --> exportpage
    exportpage --> exportsession
    exportsession --> xtffile
    generatedb --> generate
    generate --> tidmanager
    tidmanager --> qgisproject
    db --> importschema
    importschema --> importschemasession
    importschemasession --> baskets
    baskets --> importdata
    baskets --> generate
    baskets --> end
    db --> importdata
    importdata --> importdatasession
    importdatasession --> end
    importdatasession --> generate

state "Source Selection" as source
state "DB Settings for Data or Schema Import" as db
state "DB Settings for Export" as exportdb
state "DB Settings for Project Creation" as generatedb
state "OID Manager" as tidmanager
state "Import INTERLIS Model Settings" as importschema
state "***Importing INTERLIS Model***" as importschemasession
state "***Creating Baskets***" as baskets
state "Import Data Settings" as importdata
state "***Importing Data***" as importdatasession
state "Data Export Settings" as exportpage
state "***Exporting Data***" as exportsession
state "***Generating QGIS Project***" as generate
state "Intro Page" as intro
state "*File drop*" as drag
state "*End without QGIS Project*" as end
state "*QGIS Project*" as qgisproject
state "*XTF Export File*" as xtffile
```

So I guess it should change depending on what one chooses. Having the whole monster displayed is more confusing that it helps.

But luckily mostly after choosing what to do in the **intro page** or depending on what files has been chosen in the **source selection** it's more linear. E.g. 

```mermaid
stateDiagram-v2
    drag --> source
    source --> db
    generate --> tidmanager
    tidmanager --> qgisproject
    db --> importschema
    importschema --> importschemasession
    importschemasession --> baskets
    baskets --> generate

state "Source Selection" as source
state "DB Settings for Data or Schema Import" as db
state "OID Manager" as tidmanager
state "Import INTERLIS Model Settings" as importschema
state "Importing INTERLIS Model" as importschemasession
state "Generating QGIS Project" as generate
state "File drop" as drag
[...]

PR #1148, opened 2026-07-03
SettingsProphet integration

signedav:
Requires https://github.com/opengisch/QgisModelBakerLibrary/pull/175

## Add libs for ili2py
Because ili2py is using absolute import paths, we need to add the lib path of the modelbaker library to the sys path.

## Settings Prophet

In the `ImportSchemaConfigurationPage` the models model is reloaded (means by entering the page or activating/deactivating a model or receiving an additional model (by metaconfig or by dependency)), we do the following:
1. Get the checked models with their source (local or repository)
2. For each model, get the local file path (download if necessary)
3. For each model, compile the model to get the imd file (if not already cached)
4. For each model, create a BakerPyIndex from the imd file
5. Create a SettingsProphet with the model indexes and the MODELS_BLACKLIST
6. Load the prophety settings into the ili2db_options

### Actions in the ili2db pptions:
- Hides/unhides the Basket Column Setting depending on whether a basket column is required.
- Sets the Basket Column Setting depending on whether a basket column is required. If it is not required, it does not touch the setting (meaning it does not unset it).
- Hides/unhides the StrokeArcs Setting depending on whether arcs are found in the model.
- Hides/unhides the Multiple Geometry Columns Setting depending on whether multiple columns are found in a class of the model.
- Hides/unhides the Multiple Enumeration Setting depending on whether enums are found in the model.
- Hides/unhides EnumTabs option if extended enums are found in the model and smart1 is chosen -> this is interactive, see below. 
- Hides/unhides the Translation Setting if it detects the model as a translated model.

### EnumTabs with Smart1 and extended enumerations

Extended enumerations cannot be handled with smart1 inheritance strategy and enumTabs option. This means we have to avoid this combination.

https://github.com/user-attachments/assets/6acafbc0-b3d1-42ad-b19c-ab5dc666577a

When changing to smart1, it hides enumtabs, and if enumtabs was currently active, it jumps to singleenumtab. When changing back to smart2 it remains on singleenumtab (meaning it does not set it back - this is because otherwise it would be too confusing).

### Examples

|  `KbS_V1_5`  | Settings |
| :--- | :--- |
| <ul><li>has no BASKET OID definition and no extended topics</li><li>has no arc geometries</li><li>has multiple geometries in one class</li><li>has enumerations (but not extended ones)</li><li>it's not a translated model</li></ul> | image |

|  `Nutzungsplanung_V1_2`  | Settings |
| :--- | :--- |
| <ul><li>has BASKET OID definition and no extended topics</li><li>has arc geometries</li><li>has no multiple geometries in one class</li><li>has enumerations (but not extended ones)</li><li>it's not a translated model</li></ul> | image |

|  `PlansDAffectation_V1_2`  | Settings |
| :--- | :--- |
| <ul><li>has BASKET OID definition and no extended topics</li><li>has arc geometries</li><li>has no multiple geometries in one class</li><li>has enumerations (but not extended ones)</li><li>it's a translated model</li></ul> | image|

|  `KT_ArcInfrastruktur_V1`  \* | Settings |
| :--- | :--- |
| <ul><li>has no BASKET OID definition but extended topics</li><li>has arc geometries</li><li>has multiple geometries in one class</li><li>has no enumerations</li><li>it's not a translated model</li></ul> | image |

|  `Colors_V2`  \* | Settings |
| :--- | :--- |
| <ul><li>has no BASKET OID definition and no extended topics</li><li>has no arc geometries</li><li>has no multiple geometries in one class</li><li>has extended enumerations</li><li>it's not a translated model</li></ul> | image |

-> see above (EnumTabs with Smart1 and exteded enumations) what would happen on `Colors_V2` on changing to smart1.

\* test models from the [modelbaker library ](https://github.com/opengisch/QgisModelBakerLibrary)

## Fix inconsequential busy and inactive state

On receiving the model list, it does:
- check for referenced data (e.g. catalogues and find the models there: if so, it triggeres the whole thing again)
- check for available metaconfig files
- prophesy the settings

And on selecting a metaconfig file, it might find additional models, what would trigger the whole thing again as well.

See more otes in [comment here](https://github.com/opengisch/QgisModelBaker/pull/1148#issuecomment-5021168280) might be refactored once anyway, but for now I just improved the behavior like this: Avoided multiple connected slots, updating even when no models found and improved the busy state to have one busy phase instead of multiple small ones here 1785de5f707a417994cddcda3983b589886432cb

## Sponsors

QGIS Model Baker Group :v:

### DIFF
--- a/QgisModelBaker/gui/workflow_wizard/import_schema_configuration_page.py
+++ b/QgisModelBaker/gui/workflow_wizard/import_schema_configuration_page.py
@@ -27,13 +27,19 @@ from qgis.PyQt.QtCore import QSettings, Qt
 from qgis.PyQt.QtWidgets import QCompleter, QWizardPage
 
 from QgisModelBaker.gui.ili2db_options import Ili2dbOptionsDialog
+from QgisModelBaker.libs.modelbaker.ili2pytools.prophets import SettingsProphet
+from QgisModelBaker.libs.modelbaker.ili2pytools.pythonizer import BakerPyIndex
 from QgisModelBaker.libs.modelbaker.iliwrapper.globals import DbIliMode
 from QgisModelBaker.libs.modelbaker.iliwrapper.ilicache import (
     IliDataCache,
     IliDataFileCompleterDelegate,
     IliDataItemModel,
 )
-from QgisModelBaker.libs.modelbaker.utils.globals import LogLevel
+from QgisModelBaker.libs.modelbaker.utils.globals import (
+    MODELS_BLACKLIST,
+    LogLevel,
+)
+from QgisModelBaker.libs.modelbaker.utils.ili2db_utils import Ili2DbUtils
 from QgisModelBaker.utils import gui_utils
 from QgisModelBaker.utils.globals import CRS_PATTERNS
 from QgisModelBaker.utils.gui_utils import (
@@ -186,6 +192,7 @@ class ImportSchemaConfigurationPage(QWizardPage, PAGE_UI):
                 break
         self._update_ilimetaconfigcache()
         self._update_ilireferencedatacache()
+        self._update_prophety_settings()
 
     def _update_ilimetaconfigcache(self):
         self.ilimetaconfigcache = IliDataCache(
@@ -436,6 +443,8 @@ class ImportSchemaConfigurationPage(QWizardPage, PAGE_UI):
         self.setComplete(True)
 
     def _update_linked_models(self):
+        self.workflow_wizard.busy(self, True, "Update linked models...")
+
         linked_models = []
 
         for r in range(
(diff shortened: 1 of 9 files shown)


signedav:
### Some notes about the busy state of the Import Schema Configuration Page

As mentioned above, this might be improved once, but this is not main scope of this PR. This PR improved the busy handling at least and fixed some parts (see above)...

#### Some notes
On a page change (pressing Next) we do `workflow_wizard.next_id()` and depending on the current page and some parameters it decides what new page should appear #993 The parameters are individual like "are there data files chosen at the beginning?" etc.

Our use case is after the ImportDatabaseSelection-Page. It _update_configuration, but this is not costy. Then it calls `refresh_import_model` (this is what leads to all the loading). We have a source_model (having all the sources: modelfile, datafile, models from repo etc.) - this is not updated (doesn't need to. there we add and remove always manually, no loading from somewhere, that's why other models like the `import_data_file_model` doesn't need a refresh because it's just the filtered source_model). We take this source_model in refresh_import_model and call import_models_model.refresh_model, it gets the models from the source_model ili models, but as well the models from the transfer files (not sure if this is already costy - TO CHECK) and then the dependcies are triggered (on modelReset): `ImportSchemaConfigurationPage._update_models_dependent_info`:
- _update_crs_info (not costy)
- update of ilimetaconfig cache to provide metaconfigurations
  - bau neues ilimetaconfigcache (IliDataCache) und setze connections: file_download_succeeded (_on_metaconfig_received), file_download_failed (_on_metaconfig_failed), model_refreshed (_update_metaconfig_completer)
  - dann via _refresh_ili_metaconfig_cache ein refresh des IliDataCache
- update of ilireferencedata cache to load referenced
  - init__ connection zwischen ilireferencedata und model_refreshed schon viel vorher gemacht zu _update_linked_models
  - workflow_wizard.refresh_referencedata_cache and there self.ilireferencedatacache.refresh()
    - und folglich _update_linked_models, was whs nicht costy ist, dort aber: _refresh_model_data, und dort triggered es natürlich wieder _update_model_dependent_info
- update of prophety settings to provide a good default for the ili2db options

#### And a sketch

image

gacarrillor:
Adjustment In the Examples section:
`PlansDAffectation_V1_2` is a translated model and, looking at the screenshot, I guess `KT_ArcInfrastruktur_V1` is not.

QgisModelBaker/gui/workflow_wizard/import_schema_configuration_page.py
```diff
@@ -463,7 +509,233 @@ def _update_linked_models(self):
...
+                LogLevel.INFO,
+            )
+            return
+
+        settings_prophet = SettingsProphet(model_indexes, MODELS_BLACKLIST)
+        if settings_prophet is not None:
+            self.ili2db_options.load_prophecy_settings(settings_prophet)
+        self.workflow_wizard.log_panel.print_info(
+            self.tr(
+                "Enable/disable the ili2db options according the models finished."
```
gacarrillor: I needed to read that line a couple of times to understand. Can you double check?

Perhaps something like the following?

"Enabling/disabling the ili2db options according to the models has finished."

gacarrillor (approved):
Looks good to me.
I didn't test this in the GUI, though.

QgisModelBaker/gui/workflow_wizard/import_schema_configuration_page.py
signedav: Maybe "Successfully enabled/disabled the ili2db options according the models"

signedav: according *to* the models

